PR #3, opened 2019-08-29
Agency excel export

DadaDaMotha:
- Adds Excel Export Button to people view and Download link to /people
- Adds cli command to create PDF (has to be fixed later)
- Fixes bug for static assigned page break level in views

### DIFF
--- a/src/onegov/agency/cli.py
+++ b/src/onegov/agency/cli.py
@@ -8,6 +8,7 @@ from html5lib.filters.whitespace import Filter as whitespace_filter
 from io import BytesIO
 from onegov.agency.collections import ExtendedAgencyCollection
 from onegov.agency.collections import ExtendedPersonCollection
+from onegov.agency.excel_export import export_person_xlsx
 from onegov.agency.models import ExtendedAgencyMembership
 from onegov.core.cli import command_group
 from onegov.core.cli import pass_group_context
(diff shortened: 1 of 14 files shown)


src/onegov/agency/cli.py
```diff
@@ -293,6 +294,19 @@ def _create_pdf(request, app):
     return _create_pdf
 
 
+@cli.command('export-xlsx')
+@pass_group_context
+@click.option('--people', default=True, is_flag=True)
+def export_xlsx(group_context, people):
+    def _export_xlsx(request, app):
```
href: The underscore prefix is not really necessary here - nobody can reach this function anyway, so you don't need to signal that it is internal.

src/onegov/agency/excel_export.py
```diff
@@ -0,0 +1,73 @@
...
+    collection = ExtendedPersonCollection(session)
+    collection.exclude_hidden = False
+    query = collection.query().outerjoin(ExtendedPerson.memberships)
+
+    write_out = []
+
+    for person in query:
+        out_dict = OrderedDict()
+        memberships = "\n".join(
+            [f"{m.agency.title} - {m.title}" for m in person.memberships]
```
href: You can remove the brackets here (`[]`), which would turn this into a generator expression - a more efficient way of solving this.

src/onegov/agency/views/agencies.py
```diff
@@ -262,14 +263,17 @@ def cache_headers(response):
     form=Form
 )
 def create_root_pdf(self, request, form):
+    orga = request.session.query(Organisation).one()
```
href: I'd say the canonical variable name for an organisation is `org`.

href: Also, you don't need to query the organisation anywhere. The organisation is always cached on request.app.org (guaranteed to be up to date, but cached).

src/onegov/agency/cli.py
DadaDaMotha: I did this mainly because all other functions look the same and thus would like to leave it that way.

src/onegov/agency/views/agencies.py
DadaDaMotha: changed...

PR #6, opened 2019-09-02
Election-Day: Archive Search

DadaDaMotha:


src/onegov/election_day/collections/archived_results.py
```diff
@@ -259,3 +262,210 @@ def delete(self, item, request):
...
+        if not self.to_date or not self.from_date:
+            return
+
+        if self.to_date > date.today():
+            self.to_date = date.today()
+        if self.from_date > self.to_date:
+            self.from_date = self.to_date
+
+    def query(self, no_filter=False, sort=True):
+        if no_filter:
```
href: This block of code could use some spacing. You seem to have some guard clauses, checks and then default values. I would say at least the guard clauses/checks should be visibly separated from the defaults with an extra empty line.

src/onegov/election_day/collections/archived_results.py
```diff
@@ -259,3 +262,210 @@ def delete(self, item, request):
...
+        allowed_results = tuple(a[0] for a in ArchivedResult.types_of_results)
+        assert self.to_date, 'to_date must have a datetime.date value'
+        order = ('federation', 'canton', 'region', 'municipality')
+        if self.app_principal_domain == 'municipality':
+            order = ('municipality', 'federation', 'canton', 'region')
+
+        def generate_cases():
+            return tuple(
+                (ArchivedResult.domain == opt, ind + 1) for
+                ind, opt in enumerate(order)
```
href: Bet you didn't know that you can pass a start value to `enumerate` ;) - Like `enumerate(order, start=1)`, which would free you from adding the `+ 1` above.

src/onegov/election_day/collections/archived_results.py
```diff
@@ -259,3 +262,210 @@ def delete(self, item, request):
...
+            item_type=self.item_type,
+            domains=self.domains,
+            term=self.term,
+            answers=self.answers,
+            locale=self.locale,
+            page=index
+        )
+
+    def group_items(self, items, request):
+        compounded = (
```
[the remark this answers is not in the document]
DadaDaMotha: thanks for this hint!

src/onegov/election_day/collections/archived_results.py
DadaDaMotha: true..

PR #10, opened 2019-09-04
Election day incomplete proporz

DadaDaMotha:
Will hide election results if election is not completed yet.
Also changed titles to make it clear if those are the results or intermediate results.


href:
I don't know if you want me to do those reviews, so just add me as reviewer if you need me 😅

PR #12, opened 2019-09-06
Election day lists api endpoint

DadaDaMotha:
Added endpoint for aggregated lists data. Is avaialable on /election/example/data-aggregated-lists.
api_ui_vergleich_sortierung

### DIFF
--- a/tests/onegov/election_day/views/test_views_api.py
+++ b/tests/onegov/election_day/views/test_views_api.py
@@ -1,0 +1,41 @@
+import json
+from datetime import date
+
+import pytest
+from onegov.ballot import Election, ProporzElection
+from onegov.core.utils import module_path
+from onegov.election_day.utils.election.lists import \
+    get_aggregated_list_results
+from tests.onegov.election_day.common import import_wabstic_data
+
+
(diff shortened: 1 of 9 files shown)


tests/onegov/election_day/views/test_views_api.py
```diff
@@ -0,0 +1,41 @@
...
+
+import pytest
+from onegov.ballot import Election, ProporzElection
+from onegov.core.utils import module_path
+from onegov.election_day.utils.election.lists import \
+    get_aggregated_list_results
+from tests.onegov.election_day.common import import_wabstic_data
+
+
+# @pytest.mark.parametrize("tar_file", [
```
href: Any reason this is commented out? I'd say either remove it or mark it as pytest.mark.xfail if it is meant to fail.

DadaDaMotha: I lack test data at the moment and have to fix the round trip csv import/export that was not working with the data from the test server. It's meant to be un-commented later and tested with a exported test dataset originally fed in by the wabstic cleint of the customer.

PR #13, opened 2019-09-09
Election day widgets

DadaDaMotha:
Adds several widgets in the form of tables.
Tests covered the link in the Layouts that is used.
Tests for the views check every views for status 200.


src/onegov/election_day/assets/js/embedded_widgets.js
```diff
@@ -0,0 +1,15 @@
+var createEmbedLink = function(el)  {
+    var embedLink = $(el).data('embed-link');
+    var embedSource = $(el).data('embed-source');
+    if (embedLink && embedSource) {
```
href: I would think that if you run `eslint` on this, you would find some errors. We haven't set this up on our CI yet, but I would strongly encourage you run this file (and all your javascript changes) against eslint. There's an `.eslintrc` in the repository root that should get picked up to tell you what rules you should follow.

src/onegov/election_day/theme/styles/election_day.scss
```diff
@@ -1003,3 +1010,26 @@ ul.pagination li.last {
...
+}
+
+.justify-between {
+    justify-content: space-between;
+}
+
+.align-center {
+    align-content: center;
+}
+
```
href: What's with all the white-space? :)

PR #15, opened 2019-09-18
Election day roundtrip fix

DadaDaMotha:
Main changes in code:
- list_id can be alphanumeric, change that for all apis to evade roundtrip problems
- complete error handling with helper functions to be consistent (validate_x functions in common.py)
- Adding additional test for panachage results since in database of gr, target of a panachage result was the list.name where it should be list.id in order to make the backref work properly. However, the origin of this panachage result data of sg (using wabsti) is unclear, since the wabsti api format does not parse any panachage results at all. It suspect that to be a remainder from the test servers database.

Main changes in tests:
- Reorganizes sample data into folder strukture like `/domain/principal/{api_format}_{type}.tar.gz` instead of using a flat hierarchy.
- Adds an import_test_dataset fixture to instantiate model object and then load result data from the fixtures folder. 
- Tests in formats folder have become slower due to the segration of some tests resulting in more tar archive loading. Test duration increased from 0.86 s/ (86 tests) test to 1.06 s/ test (94 tests) compared to master. This can be compensated in another branch where csv files are uploaded via webtest and then used as data fixtures.


src/onegov/ballot/models/election/election.py
```diff
@@ -44,6 +44,8 @@ class Election(Base, ContentMixin, TimestampMixin,
         'polymorphic_identity': 'majorz'
     }
 
+    i18n_used_locales = ('de_CH', 'fr_CH', 'it_CH', 'rm_CH')
```
href: Wouldn't it make sense to store this at a single location? Maybe in something like `onegov/ballot/constants.py`. Bonus points for maybe fusing it with what's already in the `get_i18n_used_locales` function in `onegov/election_day/app.py`.

Especially because this seems to be duplicated many times over in other parts of this pull request.

src/onegov/ballot/models/election/election.py
```diff
@@ -345,8 +347,9 @@ def export(self):
         rows = []
         for result in results:
             row = OrderedDict()
-            for locale, title in (result[1] or {}).items():
-                row['election_title_{}'.format(locale)] = (title or '').strip()
+            for locale in Election.i18n_used_locales:
+                title = result[1] and result[1].get(locale, '') or ''
```
href: Slightly better: `title = result[1] and result[1].get(locale) or ''`

src/onegov/election_day/formats/common.py
```diff
@@ -97,6 +99,7 @@ def load_csv(
             encoding=encoding,
             rename_duplicate_column_names=rename_duplicate_column_names
         )
+        # FIXME: Remove if not necessary to speed up
```
href: Why would this be needed? This just consumes the generator (and it would be consumed again the next time you loop over the lines). Are you looping over all records to see if there's an error? If not, then you should remove this.

src/onegov/election_day/formats/election/wabstic_proporz.py
```diff
@@ -283,9 +242,10 @@ def import_election_wabstic_proporz(
 
         # Check if the entity is counted
         try:
-            entity['counted'] = False if int(line.sperrung or 0) == 0 else True
-        except ValueError:
-            line_errors.append(_("Invalid entity values"))
+            counted = validate_integer(line, 'sperrung')
+            entity['counted'] = False if counted == 0 else True
```
href: I find this somewhat puzzling: Basically you say `entity['counted']` is `False` if `counted` is `0`.

This could be rewritten as follows: `entity['counted'] = not counted`. Which I think clarifies why this seems weird. Could this be a big? If not, can you add a comment to explain this?

src/onegov/election_day/formats/vote/wabstic.py
```diff
@@ -149,7 +128,8 @@ def import_vote_wabstic(vote, principal, number, district,
 
         # Check if the entity is counted
         try:
-            counted = False if int(line.sperrung or 0) == 0 else True
+            counted_num = validate_integer(line, 'sperrung')
+            counted = False if counted_num == 0 else True
```
href: Same as above. Is `counted_num` maybe not a good name? It seems like you want to say that something is counted, if the number of uncounted values is 0.

src/onegov/election_day/import_export/mappings.py
```diff
@@ -0,0 +1,279 @@
...
+
+For complex queries, it is much better to write them in pure sql. Return
+the table as named tuples and work with that and this mapping.
+
+"""
+
+
+#  --- INTERNAL
+
+INTERNAL_COMMON_ELECTION_HEADERS = [
```
href: You are inconsistent in this file with lists/tuples. I think you should always use tuples. You can still combine tuples like this:
```python
foo = (1, 2)
bar = (*foo, 3, 4)
```

tests/onegov/election_day/common.py
```diff
@@ -17,6 +19,57 @@ def print_errors(errors):
...
+
+def create_principal(principal=None, municipality=None):
+    if principal in Canton.CANTONS:
+        pr = Canton(canton=principal)
+    else:
+        pr = Municipality(municipality=municipality)
+    return pr
+
+
+def compare_elections(left, right):
```
href: Since this doesn't actually do anything, you should either remove it or raise a `NotImplementedError`

src/onegov/election_day/formats/election/wabstic_proporz.py
DadaDaMotha: So `int(line.sperrung or 0)` corresponds 1:1 to `validate_integer(line, 'sperrung')` because `validate_integer` returns 0 if in the csv there was '' since `treat_emtpy_as_default=True` and `default` is 0 =) are the defaults of the function.
`False if counted == 0 else True` will thus also be the equivalent of the previous expression `False if int(line.sperrung or 0) == 0 else True`.
And you are right, `not counted` also is equivalent, so I will use this.

src/onegov/ballot/models/election/election.py
DadaDaMotha: I tried to get this from the app since I saw it there, but accessing the app there inside the model was not possible. I'll move this to one file.

src/onegov/election_day/formats/election/wabstic_proporz.py
href: `not counted` was just meant to illustrate how weird this is going to be for someone reading this in the future. I think you should add a good comment for this, otherwise someone might be tempted to just invert this, thinking it's an error.

src/onegov/ballot/models/election/election.py
DadaDaMotha: yep =)

src/onegov/election_day/formats/common.py
DadaDaMotha: The solution was to call `next(csv.lines)` to raise the `EmptyLineInFileError`

src/onegov/election_day/formats/vote/wabstic.py
DadaDaMotha: counted here refers hour of locking of the result, so it would be better to make that clear using meaningful variables.

src/onegov/election_day/formats/election/wabstic_proporz.py
DadaDaMotha: I added a comment what this is from the wabstic doc and remaned the variables.
`locking_time = validate_integer(line, 'sperrung', default=False)`
 `entity['counted'] = False if not locking_time else True`

src/onegov/election_day/import_export/mappings.py
DadaDaMotha: thanks! i'll do that...

tests/onegov/election_day/common.py
DadaDaMotha: removed...

PR #34, opened 2019-12-10
Fsi ui improvements

DadaDaMotha:
- auto-resizing of iframe
- Hiding cols in reservation table
- lazy-loading switch for accordions /fsi/courses (not yet working as it should)
- use fa icons when printing "course attended" column

Todo: replace the slider element/checked attribute of input DOM element in /fsi/reservations with the actual response from the view Reservation - toggle-confirm.

To check lazy loading of accordion: uncomment `src/onegov/fsi/layouts/course.py:75`
Is the encoding wrong, or the json response has to be in a different format? Any hint would be appreciated

### DIFF
--- a/src/onegov/fsi/assets/js/ifs.js
+++ b/src/onegov/fsi/assets/js/ifs.js
@@ -1,0 +1,26 @@
+function autoResizebyID(id) {
+    var newheight;
+    var newwidth;
+
+    if (document.getElementById) {
+        newheight = document.getElementById(id).contentWindow.document.body.scrollHeight;
+        newwidth = document.getElementById(id).contentWindow.document.body.scrollWidth;
+    }
+
+    document.getElementById(id).height = (newheight) + "px";
+    document.getElementById(id).width = (newwidth) + "px";
+}
+
+function autoResize() {
+    var newheight;
+    var newwidth;
+    var frame;
+    if (document.getElementsByClassName) {
+        for (frame of document.getElementsByClassName('resizeable')) {
+            newheight = frame.contentWindow.document.body.scrollHeight;
+            newwidth = frame.contentWindow.document.body.scrollWidth;
+            frame.height = (newheight) + "px";
+            frame.width = (newwidth) + "px";
+        }
+    }
+}
--- a/src/onegov/fsi/assets/js/ifs.js
+++ b/src/onegov/fsi/assets/js/ifs.js
@@ -1,26 +1,13 @@
-function autoResizebyID(id) {
-    var newheight;
-    var newwidth;
-
-    if (document.getElementById) {
-        newheight = document.getElementById(id).contentWindow.document.body.scrollHeight;
-        newwidth = document.getElementById(id).contentWindow.document.body.scrollWidth;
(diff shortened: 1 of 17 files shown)


src/onegov/fsi/assets/js/ifs.js
```diff
@@ -0,0 +1,26 @@
+function autoResizebyID(id) {
+    var newheight;
+    var newwidth;
+
+    if (document.getElementById) {
```
href: Not really necessary, since this function has been supported by browsers going back as far as IE 5.5. Also, if for some reason this condition is false, an error will occur further down anyway.

src/onegov/fsi/assets/js/ifs.js
```diff
@@ -0,0 +1,26 @@
...
+
+    document.getElementById(id).height = (newheight) + "px";
+    document.getElementById(id).width = (newwidth) + "px";
+}
+
+function autoResize() {
+    var newheight;
+    var newwidth;
+    var frame;
+    if (document.getElementsByClassName) {
```
href: Same here, this one has been supported since IE9 and we require IE11.

src/onegov/fsi/collections/reservation.py
```diff
@@ -44,12 +44,13 @@ def attendee_collection(self):
 
     def query(self):
         query = super().query()
-        query = query.join(CourseAttendee).order_by(
-            CourseAttendee.last_name,
-            CourseAttendee.first_name,
-        )
+        # query = query.join(CourseAttendee).order_by(
```
href: Try not to include commented-out code.

src/onegov/fsi/templates/course_event.pt
```diff
@@ -12,26 +12,26 @@
...
-                                <p tal:case="True" i18n:translate>No entries found.</p>
-                                <ul tal:case="False">
-                                <tal:b tal:repeat="reservation model.reservations">
-                                    <li tal:attributes="class python: 'grayed-out' if reservation.is_placeholder else ''">${str(reservation)}</li>
-                                </tal:b>
-                            </ul>
 
+                            <tal:b tal:condition="request.is_admin">
+                                <h3 i18n:translate="">Attendees</h3>
+                                    <tal:b tal:switch="model.reservations.count() == 0">
```
href: `not model.reservations` might work the same, without the cost of `count`, which is slow in Postgres.

DadaDaMotha: The problem here with this relationsship is that bool(ResultProxy) is always true, also if there are not results. Is there a less costly way to check the result count?

href: Maybe `.first()` might be a bit faster. But probably not by much. Technically an `exists` query is the quickest way to check this, but those are a bit hard to create in SQLAlchemy. So I'd say leave it as is.

src/onegov/fsi/assets/js/ifs.js
DadaDaMotha: oki

PR #45, opened 2020-01-16
Fsi locked events

DadaDaMotha:
Only admins should be able add subscriptions for locked course events.
Adds property `locked_for_subscriptions` to `CourseEvent` and adapts form.
Introduces checks for reservation adding/editing in views.

### DIFF
--- a/tests/onegov/fsi/test_views_reservation.py
+++ b/tests/onegov/fsi/test_views_reservation.py
@@ -1,6 +1,33 @@
+import pytest
 from onegov.fsi.collections.course_event import CourseEventCollection
-from onegov.fsi.models import CourseReservation, CourseAttendee, CourseEvent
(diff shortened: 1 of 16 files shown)


tests/onegov/fsi/test_views_reservation.py
```diff
@@ -1,8 +1,35 @@
+import pytest
 from onegov.fsi.collections.course_event import CourseEventCollection
-from onegov.fsi.models import CourseReservation, CourseAttendee, CourseEvent
+from onegov.fsi.models import CourseReservation, CourseAttendee, CourseEvent, \
+    Course
 from onegov.user import User
 
 
+@pytest.mark.skip('Webstest ES configuration issue')
```
DadaDaMotha: When you run this test, you get an issue with the ES and the app configuration. Can you reproduce it and see a hint in the StackTracke?

PR #172, opened 2021-11-16
Ogc 124 logos footer

BreathingFlesh:


msom (changes requested):
Please make sure that the tests don't fail.

PR #210, opened 2022-02-02
Show function in membership instead of person function

BreathingFlesh:


msom (changes requested):
Fix tests

msom (changes requested):
should be membership.title not membership.person.title?

PR #228, opened 2022-02-14
Hide newsletter by default

BreathingFlesh:


msom (changes requested):
Please add DB upgrade

PR #242, opened 2022-02-22
make archive header h2

steinbock27:


src/onegov/election_day/templates/archive.pt
```diff
@@ -103,7 +103,7 @@
 
         <div class="row archive" tal:define="years layout.archive.get_years()" tal:condition="years">
             <div class="columns small-12 small-centered">
-                <h1 i18n:translate="" tal:condition="not: h2|nothing">Archive</h1>
+                <h2 i18n:translate="" tal:condition="not: h2|nothing">Archive</h2>
```
msom, line 106: Both condition do the same, we can just use one `<h2 i18n:translate="">Archive</h2>`. The flag is never set anywhere anyway.

msom (changes requested):

PR #244, opened 2022-02-23
Ogc 268

steinbock27:


src/onegov/election_day/templates/archive.pt
```diff
@@ -68,30 +68,30 @@
                             </thead>
                             <tbody>
                                 <tr tal:repeat="vote archive_items[date][domain][type]">
-                                    <td tal:define="prefix vote.title_prefix(request)">
+                                    <th tal:define="prefix vote.title_prefix(request)">
                                         <a href="${vote.url}"><span tal:condition="prefix">${prefix}: </span>${vote.title}</a>
-                                    </td>
-                                    <td>
+                                    </th>
+                                    <th>
```
msom, line 74: Only change the title cells to th, not all.

msom (changes requested):
Make sure that the archive still looks the same.

PR #259, opened 2022-03-04
Ogc 178

steinbock27:


.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
msom, line 1: Please use a separate PR for things unrelated to this ticket.

msom (changes requested):

PR #260, opened 2022-03-04
Ogc 177

steinbock27:
Bugs fixed. It passes the `tests/onegov/feriennet/test_browser.py::test_browse_billing`on local. But it still throws an error in Buildkeit.


steinbock27:
Problem solved.

msom (changes requested):

PR #261, opened 2022-03-04
Ogc 169

steinbock27:
Same issue with ogc-177. Although the bugs fixed (`test_screen_form_populate, test_browse_billing, linting update pip`), it throws error in Buildkeit.


steinbock27:
Mentioned problems are solved.

src/onegov/org/cli.py
```diff
@@ -28,7 +28,6 @@
 from onegov.file import File
 from onegov.form import as_internal_id, parse_form
 from onegov.form import FormCollection
-from onegov.newsletter import Newsletter
```
msom, line 31: Please only included changes regarding this ticket.

msom (changes requested):

PR #274, opened 2022-03-11
Ogc 277

steinbock27:


src/onegov/election_day/templates/macros.pt
```diff
@@ -968,7 +968,7 @@
                         <li tal:condition="not submenu" tal:attributes="class 'active' if active else ''">
                             <a href="${link}">${title}</a>
                         </li>
-                        <li tal:condition="submenu" tal:attributes="class 'has-dropdown active' if active else 'has-dropdown'">
+                        <li tal:condition="submenu" tal:attributes="class 'has-dropdown active' if active else 'has-dropdown';aria-expanded 'true' if active else 'false'">
```
msom, line 971: `active` refers to the active menu item, not the open drop down (this would be the class `hover`, probably set using JavaScript)

src/onegov/election_day/theme/styles/election_day.scss
```diff
@@ -615,8 +615,22 @@ footer {
         }
 
         .left {
-            li.active {
-                border-left: 2px solid $primary-color;
+
+            li {
```
msom, line 619: What is the reason for this restyling?

msom (changes requested):

PR #312, opened 2022-04-13
Ogc 368 categories

BreathingFlesh:


Daverball (changes requested):
Looks fine for the most part, apart from some performance related nitpicks. The import is probably not working however, we will need to add test coverage for that.

src/onegov/org/forms/event.py
```diff
@@ -246,6 +247,9 @@ def populate_submitter(self):
             self.email.data = self.request.current_username
 
     def on_request(self):
+        if self.custom_tags():
+            self.tags.choices = self.custom_tags()
```
Daverball, line 251: This still needs to be a list of tuples.

src/onegov/org/forms/event.py
```diff
@@ -505,7 +515,13 @@ def run_import(self):
         headers = self.headers
         session = self.request.session
         events = EventCollection(session)
-        tags = {self.request.translate(tag[0]): tag[0] for tag in TAGS}
+        all_tags = chain(
+            TAGS,
+            self.custom_tags() if self.custom_tags() else []
```
Daverball, line 520: `self.custom_tags() or []` would be even neater, you can use `or` short-circuiting to provide defaults for values that would evaluate to `False`.

PR #344, opened 2022-05-11
1022 banner

BreathingFlesh:


src/onegov/feriennet/app.py
```diff
@@ -95,9 +95,14 @@ def banner(self, request, id):
...
 
         """
+        language = request.locale[:2]
         candidates = [
             sponsor for sponsor in self.sponsors
-            if getattr(sponsor, 'banners', None) and id in sponsor.banners
+            if (
+                getattr(sponsor, 'banners', None)
+                and id in sponsor.banners
+                and sponsor.banners['bookings']['src'][language]
```
msom, line 104: Does ['bookings']['src''] exist in every case or could it raise an AttributeError? A more error-prone possibility would be to use sponsor.banners.get('bookings', {}).get('src', {}).get('language', None).

src/onegov/feriennet/app.py
```diff
@@ -106,12 +111,17 @@ def banner(self, request, id):
         winner = random.choice(candidates)
         winner = winner.compiled(request)
 
+        if winner.name == "ProJuventute":
```
msom, line 114: A cleaner solution which provides additional functionality would be to make the info also configurable in the yaml (for every language). This would also allow us to use clearer names for the banners (e.g. Projuventute Handy, Projuventute Resilienz, ...)

msom (changes requested):
Please cover your changes in choosing a banner with a test (see test_sponsors).

msom (approved):

PR #360, opened 2022-05-27
Drop the dot in py.test

jugmac00:
pytest dropped the dot in the name already a couple of years ago.


msom:
Thanks for the hint!

PR #429, opened 2022-07-08
Ogc 354 reservation link

BreathingFlesh:


msom (changes requested):
Do the existing tests really cover all the changed and probably new coder?

PR #477, opened 2022-08-25
Ogc 569 find your spot

BreathingFlesh:


src/onegov/org/views/resource.py
```diff
@@ -93,13 +93,31 @@ def get_resource_form(self, request, type=None):
...
+
+    for group, entries in resources.items():
+        # don't include find-your-spot link for categories with
+        # no rooms
+        if not contains_at_least_one_room(entries):
+            continue
+        entries.insert(0, FindYourSpotCollection(
+            request.app.libres_context,
+            group=None if group == default_group else group
+        ))
```
Daverball, line 119: This should happen after `combine_grouped` otherwise "Find Your Spot" will be sorted alphabetically into the group, it should always stick to the top.

Also `combine_grouped` could be slightly optimized by using `itertools.chain` instead of the `+` operator so we don't create a temporary list before sorting.

Daverball:
After you resolve the above issue and fix the linting error, it should be fine to merge.

Bonus points for adding a regression test, that makes sure the resources view contains the find-your-spot url for all groups that contain at least one room.

PR #538, opened 2022-09-28
Election Day: Export all votes and elections of all time

cyrillkuettel:

## Commit message

Election Day: Export all votes and elections of all time

TYPE: Feature
LINK: OGC-483

## Checklist

# Archive
The archive.zip is structured like this:

archive.zip


tests/onegov/election_day/utils/test_archive_generator.py
```diff
@@ -0,0 +1,221 @@
...
+from onegov.election_day.utils import add_local_results
+from onegov.election_day.utils.archive_generator import ArchiveGenerator
+from datetime import date
+from fs.walk import Walker
+from collections import Counter
+from fs.zipfs import ReadZipFS
+from fs.tempfs import TempFS
+from fs.copy import copy_dir
+from fs.osfs import OSFS
+
```
msom, line 14: General hint: The election day test suite is very slow, mostly because importing elections and votes take long, espcially when doing over and over again in different tests. That's why I recently started testing more functionality in one test (see e.g. test for election utils) so that import is done only once.

Think about merging some tests into one, e.g. testing query votes and elections, creating zip, checking zip etc. If you use an app without pre-added data you can also test how it behaves in case no given data is given.

msom (changes requested):
Looks good overall but needs a bit work to remove superfluous functionality and so that it fits to typical approaches and structures used in election day.

src/onegov/election_day/utils/archive_generator.py
```diff
@@ -0,0 +1,190 @@
...
+        return self.generate_csv(subset=["votes"])
+
+    def generate_elections_csv(self):
+        return self.generate_csv(subset=["elections"])
+
+    def all_votes(self):
+        return self.session.query(Vote).order_by(desc(Vote.date)).all()
+
+    def all_elections(self):
+        return self.session.query(Election).order_by(desc(Election.date)).all()
```
msom, line 151: Please exclude elections that are part of election compounds

PR #548, opened 2022-10-10
Ogc 547 accessibility

BreathingFlesh:

## Commit message

Election Day: Add roles for Accessibility

TYPE: Bugfix
LINK: OGC-547

## Checklist

### DIFF
--- a/src/onegov/election_day/assets/js/jquery.tablesorter.js
+++ b/src/onegov/election_day/assets/js/jquery.tablesorter.js
@@ -639,7 +639,7 @@
 			}
 			c.$table.find( c.selectorHeaders ).attr({
 				scope: 'col',
-				role : 'columnheader'
+				role : 'switch columnheader'
 			});
 			// enable/disable sorting
 			ts.updateHeader( c );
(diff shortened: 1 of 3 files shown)


src/onegov/election_day/assets/js/jquery.tablesorter.js
```diff
@@ -639,7 +639,7 @@
 			}
 			c.$table.find( c.selectorHeaders ).attr({
 				scope: 'col',
-				role : 'columnheader'
```
msom, line 642: This is actually an asset from https://github.com/Mottie/tablesorter, if we update this in the future, the change will be lost. If this change is reasonable in general, you should open a pull request there (and keep this change). If not, we should try to introduce this change in an asset fully controlled by us or alternatively at least document this somewhere, in a comment in `get_common_asset`.

msom (changes requested):
Read to merge when PR created or comment added, see comment.

src/onegov/election_day/assets/js/jquery.tablesorter.js
BreathingFlesh: I created a PR for tablesorter.

BreathingFlesh: There hasn't been an update of tablesorter since 2020, I will try the second option of making the change in an asset.

PR #624, opened 2022-12-21
Town6: Fix html rendering in newsletters/new view

cyrillkuettel:

## Commit message

Town6: Fix html rendering in newsletters/new view

WTForms 2.3 started escaping label text with markupsafe [wtforms/wtforms#315](https://github.com/wtforms/wtforms/issues/315)
We have several cases where we actually do want the label to contain
html. For this we can declare the __html__ method, which will be used
by wtforms internally.

TYPE: Bugfix
LINK: OGC-734

## Checklist


msom:
Why not just use Markup from markupsafe (as we do in other cases, such as widgets)?

PR #649, opened 2023-01-17
Town6:  Display context-specific functions on person in the directory

cyrillkuettel:
## Commit message

Town6:  Display context-specific functions on person in the directory

TYPE: Feature
LINK: OGC-731

## Checklist

    - [X] I have tested styling changes/features on different browsers

### DIFF
--- a/src/onegov/org/views/people.py
+++ b/src/onegov/org/views/people.py
@@ -6,7 +6,7 @@ from onegov.org import _, OrgApp
 from onegov.org.elements import Link
 from onegov.org.forms import PersonForm
 from onegov.org.layout import PersonLayout, PersonCollectionLayout
-from onegov.org.models import AtoZ
+from onegov.org.models import AtoZ, Topic
 from onegov.people import Person, PersonCollection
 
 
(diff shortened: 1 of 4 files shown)


tests/onegov/org/test_browser.py
```diff
@@ -439,3 +440,32 @@ def test_external_map_link(browser, client):
     browser.visit('/topics/themen')
     browser.find_by_css(".button-state").click()
     assert browser.is_text_present("Karte Geo-BS")
+
+
+@pytest.mark.flaky(reruns=3)
+def test_context_specific_function_are_displayed_in_person_directory(
```
cyrillkuettel: The exact same test with `client` somehow did not work. So that's the reason I used `browser`.

src/onegov/org/views/people.py
```diff
@@ -33,13 +33,55 @@ def get_items(self):
 @OrgApp.html(model=Person, template='person.pt', permission=Public)
 def view_person(self, request, layout=None):
 
+    pages = request.session.query(Topic)
+    pages = pages.filter(Topic.people != None).all()
```
msom, line 37: Note that you should not use `==` and `!=` with `None` in python, with sqlalchemy you can use `is_(None)` and `isnot(None)`: `Topic.people.isnot(None)`

src/onegov/town6/templates/macros.pt
```diff
@@ -834,6 +834,16 @@
                     <li tal:condition="person.website and 'website' not in exclude" class="person-card-website">
                         <a href="${person.website}">${person.website}</a>
                     </li>
+                    <div class="person-card-context-specific-functions" tal:condition="python: len(functions) != 0 and len(organizations) != 0 and organizations is not None and functions is not None">
```
msom, line 837: just use `python: functions and organizations`

src/onegov/town6/templates/macros.pt
```diff
@@ -834,6 +834,16 @@
...
+                        <div tal:define="map python: zip(organizations, functions)">
```
msom, line 838: If you intend to use it like this, you should probably pass it at a list of lists  in the first place (which also less error prone). Why not pass only only one string with "organization: function" already anyway?

src/onegov/town6/templates/macros.pt
```diff
@@ -834,6 +834,16 @@
...
+                            <span i18n:translate>Functions:</span>
+                            <ul>
+                                <li tal:repeat="(organization, function) map">
+                                    <span>${organization}: ${function}</span>
```
msom, line 842: you check if function is empty above, I guess this means that there will be cases rendered as organization without function, e.g. "Organization:"?

tests/onegov/org/test_browser.py
```diff
@@ -439,3 +440,32 @@ def test_external_map_link(browser, client):
     browser.visit('/topics/themen')
     browser.find_by_css(".button-state").click()
     assert browser.is_text_present("Karte Geo-BS")
+
+
+@pytest.mark.flaky(reruns=3)
```
msom, line 445: Any idea why this is flaky? Maybe better just use a view test instead of a browser test (if possible) instead of introducing more flakiness...

msom (changes requested):

src/onegov/town6/templates/macros.pt
cyrillkuettel: I can do this, but why is it error prone? When using  `map` wouldn't the iterator stop when the shortest input iterable is exhausted, e.g. the if it somehow happens the lists are not equal in length.

src/onegov/town6/templates/macros.pt
cyrillkuettel: Yes, sometimes there exists an organization, but no mapping to function. (The function field is optional)

PR #688, opened 2023-02-07
Tries to speed up CI using pytest-xdist.

Daverball:


Daverball:
Looks like xdist produces some crashes/deadlocks, I guess that's why we're not using it with OneGov.

Maybe there is a clean way to get parallelism in a test script without screwing up the output on Buildkite. That way we could spin up multiple test groups on the same CI worker and utilize all the worker's cores. We should also look at whether there is something better than test-groups that lets you split the workload more evenly based on metrics from past runs (although we would have to be very careful about synchronization issues between the four workers in that case).

PR #696, opened 2023-02-13
Ogc 947 show events in the past from link

Tschuppi81:
## Commit message

Events: Events from the past can be shown.

In the case of 'No events found' we will display two link buttons. One will point to 'all events' where as the other will show 'past events'. In addition a new time range filter 'past events' was added to achieve the same from filters.

TYPE: Feature
LINK: OGC-947

### DIFF
--- a/src/onegov/event/collections/occurrences.py
+++ b/src/onegov/event/collections/occurrences.py
@@ -218,7 +218,7 @@ class OccurrenceCollection(Pagination):
         if self.start is not None or self.outdated is False:
             if self.start is None:
                 start = date.today()
-            elif self.range:
+            elif self.range == 'past':
                 start = self.start
             else:
                 start = self.start
--- a/src/onegov/event/collections/occurrences.py
+++ b/src/onegov/event/collections/occurrences.py
@@ -118,9 +118,9 @@ class OccurrenceCollection(Pagination):
             start = today.replace(day=1)
             return start, start + relativedelta(months=1, days=-1)
         if range == 'past':
-            milenium = date(2000, 1, 1)
+            millennium = date(2000, 1, 1)
             yesterday = today - timedelta(days=1)
-            return milenium, yesterday
+            return millennium, yesterday
         pass
 
     def for_filter(self, **kwargs):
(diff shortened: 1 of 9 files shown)


msom (commented):
Looks good, a few small things:
- Use EN on github
- Update macro for org
- Update JS for date range filter
- Optionally hide "All Events" link if no filter is set

src/onegov/town6/templates/macros.pt
```diff
@@ -1163,6 +1163,13 @@
     </tal:b>
         </div>
     <div tal:condition="not occurrences" i18n:translate>No events found.</div>
+    <div tal:condition="not occurrences">
+        <!-- "${layout.events_url}" -->
```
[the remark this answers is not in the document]
Tschuppi81: done

src/onegov/event/collections/occurrences.py
```diff
@@ -116,6 +117,10 @@ def range_to_dates(self, range, start=None, end=None):
         if range == 'month':
             start = today.replace(day=1)
             return start, start + relativedelta(months=1, days=-1)
+        if range == 'past':
+            milenium = date(2000, 1, 1)
```
BreathingFlesh, line 121: "millenium" should have two L's, but it won't appear anywhere, so it doesn't really matter.

src/onegov/town6/templates/macros.pt
```diff
@@ -1163,6 +1163,15 @@
     </tal:b>
         </div>
     <div tal:condition="not occurrences" i18n:translate>No events found.</div>
+    <div tal:condition="not occurrences">
+        <br>
+        <div tal:condition="'range' in layout.og_url">
+            <a href="${layout.events_url}" class="button hollow">Alle Veranstaltungen</a>
+        </div>
+        <div>
```
BreathingFlesh, line 1171: I don't think that div does anything.

BreathingFlesh (commented):
Only minor changes. Comments on macros.pt in town6 also apply for macros in org.

src/onegov/town6/templates/macros.pt
Tschuppi81: done

src/onegov/town6/templates/macros.pt
```diff
@@ -1163,6 +1163,15 @@
     </tal:b>
         </div>
     <div tal:condition="not occurrences" i18n:translate>No events found.</div>
+    <div tal:condition="not occurrences">
```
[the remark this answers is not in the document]
Tschuppi81: good input

PR #711, opened 2023-02-20
Core: Add compatibility with pglast 5.

msom:
## Commit message

Core: Add compatibility with pglast 5.

TYPE: Feature
LINK: OGC-992

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -113,7 +113,7 @@ install_requires =
     pdfdocument
     pdfrw
     pdftotext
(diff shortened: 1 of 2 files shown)


msom:
@Daverball is there something besides the tests that I need to check before merging?

Daverball (approved):
@msom No, if the tests all pass it should be fine. There is a lot of core functionality that depends on this function so most tests would already fail in the setup phase if this isn't working properly.

PR #718, opened 2023-02-23
Prevent pitfall on theme changes using auto compile setting

Tschuppi81:


Tschuppi81:
not needed anymore, already done

PR #730, opened 2023-02-28
Org: Fix Forms without group not being displayed.

cyrillkuettel:
## Commit message

Org: Fix Forms without group not being displayed.

To summarize the issue, because it's very subtle:
1. Forms can be defined at `/forms`.
2. The user can create a new form, and the group field is optional. If no group is set by the user, the system will fall back using `default_group` which called the `General` group. However, if the user creates a new form and *explicitly sets the string `General` as the group name*, the previous `General` group will be overwritten! This overwriting is fixed in this commit. 

TYPE: Bugfix
LINK:  OGC-857

## Checklist


src/onegov/org/utils.py
```diff
@@ -884,6 +884,8 @@ def sort_key(record):
 
     transform = transform or (lambda v: v)
 
+    # groupby expects the input iterable (records) to already be sorted
+    records = sorted(records, key=group_key)
```
cyrillkuettel, line 888: Who doesn't like a oneline fix :relaxed:

PR #735, opened 2023-03-03
Adding help creating a collapsable list

Tschuppi81:
Forms: Adds user snippet for subfields to formcode

Feature
LINK: ogc-977


Tschuppi81:
@Daverball, do you think this is useful for the clients?

Daverball:
@Tschuppi81 Difficult to predict. But I think "Collapsable list" as a name for the concept is misleading, it's more like "Subfields dependent on selected option". It might also be a little bit more helpful to have one example each for checkboxes and radio buttons, otherwise people may get the idea, that only of the two work.

Also just as a side note: It's called formcode and not YAML forms. Formcode is closer to Markdown than YAML. We just happen to transform the Formcode to YAML, so we can use the YAML parser to do some of the heavy lifting for us, rather than having to specify the Formcode grammar completely in pyparsing and manually having to process the ParseResult into the proper structure of fieldsets, fields and subfields.

src/onegov/form/parser/snippets.py
```diff
@@ -39,6 +39,13 @@ class Snippets:
             '    [ ] B\n'
             '    [ ] C'
         )),
+        (_("Subfields depending on selected option"), (
```
Daverball, line 42:
```suggestion
        (_("Subfields depending on choice"), (
```

This wording is a little bit more consistent

Daverball (commented):
The wording is currently a little bit inconsistent, so I'd slightly adjust it to avoid potential confusion. Otherwise this looks good now.

src/onegov/form/parser/snippets.py
Tschuppi81: I like it, will adapt

src/onegov/form/locale/de_ch/LC_MESSAGES/onegov.form.po
```diff
@@ -80,6 +80,9 @@ msgstr "Auswahl"
 msgid "Multiple Choice"
 msgstr "Mehrfachauswahl"
 
+msgid "Subfields depending on choice"
+msgstr "Unterfelder je nach Wahl"
```
Daverball, line 84:
```suggestion
msgstr "Unterfelder je nach Auswahl"
```

Daverball (approved):
Minor nit

PR #742, opened 2023-03-06
Event: Make Test less flaky

cyrillkuettel:
## Commit message

Event: Make Test less flaky

Test has repeatedly failed depending on the time, so we freeze the time. It appears that the assert statements have failed in the past because the comparison is off by just about 0.12 seconds: 

microseconds

TYPE: Bugfix

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,7 +65,7 @@ install_requires =
     bleach
     blinker
     cached_property
-    chameleon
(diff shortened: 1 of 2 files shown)


msom:
I don't think freezing time will work because `latest_occurrence` uses the SQL function `now` which is not affected by `freezegun`.

PR #746, opened 2023-03-09
Feature/ogc 263 allow adding root pages

Tschuppi81:
Topics: Now any admin can add topics (pages) on the top (root) level. 

TYPE: Feature
LINK: ogc-263

## Checklist


src/onegov/org/views/editor.py
```diff
@@ -74,6 +80,36 @@ def handle_new_page(self, request, form, src=None, layout=None):
     }
 
 
+def handle_new_root_page(self, request, form, src=None, layout=None):
```
cyrillkuettel, line 83: I'm guessing the `src` parameter was created as a result of copy-pasting from `handle_new_page`. 
It's not needed and  always == `None`, it can be removed.

cyrillkuettel (changes requested):

tests/onegov/town6/test_views_homepage.py
```diff
@@ -88,3 +88,23 @@ def test_homepage(client):
     client.app.org.disable_chat = True
     transaction.commit()
     assert not client.get('/').pyquery('.chat')
+
+
```
cyrillkuettel, line 92: Do we need two duplicate tests? I think one can be removed, since `town6` uses the `org` implementation.

src/onegov/org/views/editor.py
msom: Or do wee need to be able to paste to the root as well?

msom (changes requested):
Add town6 icon and decide whether paste into root should be supported (not necessary in my opinion).

src/onegov/org/views/editor.py
Tschuppi81: I would not support 'paste' on root level. So let me remove the 'src' param from handle_new_root_page

src/onegov/town6/layout.py
```diff
@@ -2410,5 +2410,13 @@ def editbar_links(self):
                     _("Sort"),
                     self.request.link(self.model, 'sort'),
                     attrs={'class': ('sort-link')}
-                )
+                ),
+                Link(
+                    _("Add"),
+                    self.request.link(Editor('new-root', self.model, 'page')),
+                    classes=(
```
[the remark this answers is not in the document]
Tschuppi81: done

src/onegov/org/views/editor.py
Tschuppi81: done

PR #760, opened 2023-03-17
Feriennet: Make sure importing transaction files works again.

cyrillkuettel:
## Commit message

Feriennet: Make sure importing transaction files works again

TYPE: Bugfix
LINK: PRO-1156

## Checklist


src/onegov/activity/iso20022.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, **kwargs):
         for key, value in kwargs.items():
             setattr(self, key, value)
 
-        self.username = None
+        self.username = ''  # Has to be str for comparison used in sort
```
cyrillkuettel, line 31: This was the root cause of the infamous error we were getting a lot: 
`TypeError: '<' not supported between instances of 'NoneType' and 'str'`

src/onegov/activity/iso20022.py
```diff
@@ -38,11 +38,11 @@ def __repr__(self):
...
+        if self.reference:
+            return {self.reference}
         return set(self.extract_references())
 
     def extract_references(self):
-        if self.reference:
-            yield self.reference
 
         # currently only one schema supports text extraction, the others are
         # stored in the designated reference field
```
cyrillkuettel, line 48: It was creating unique references based on the `booking_text`, which was always the same. 
This was the reason for transactions incorrectly being labelled as duplicates.

cyrillkuettel:
Also I'm pretty sure [this comment](https://github.com/OneGov/onegov-cloud/blob/96a6f05885a74fcfe4024778d1b55aad62a3fb4a/src/onegov/activity/iso20022.py#L229) is a typo.
 I think he meant to write "*debit* transactions are completely irrelevant for us" That matches the commit message in (bf7561c7d15f392bbcba86596eacbc52d37c2160)

src/onegov/activity/iso20022.py
msom: Does this really work with other schemas than feriennet-v1? ESR/QR-References are obviously stored in other fields...

msom: Failing test supports this suspicion...

msom: Maybe the logic "reference field present -> must be a reference schema" is flawed. Maybe it should be more something like references contains all possible references (ESR and things from the note).

src/onegov/activity/iso20022.py
```diff
@@ -38,11 +38,11 @@ def __repr__(self):
 
     @cached_property
     def references(self):
+        if self.reference:
+            return {self.reference}
         return set(self.extract_references())
```
cyrillkuettel, line 43: I was thinking that once `self.reference` is present, there's no need to go any further. We can just return it. (assuming we have a reference schema).

src/onegov/activity/iso20022.py
cyrillkuettel: I don't have a test file for `feriennet-v1` so it' hard to say.

PR #762, opened 2023-03-21
agency: Adding many new fields to tables 'people' and 'agencies' 

Tschuppi81:
agency: Adding many more fields for agencies and people

I also added a db upgrade job to add the columns to the two tables. In addition it will parse the people.address column and separate the information in the newly added fields.

TYPE: Feature
LINK: ogc-966
HINT: Run `onegov-people --select /onegov_agency/* migrate-people-address-field --dry-run` after upgrade

## Open questions
- is it ok that the api changes and will no longer support the former 'address' field?
http://localhost:8080/onegov_agency/bs/api/people
http://localhost:8080/onegov_agency/bs/api/agencies

## Checklist


src/onegov/fsi/upgrade.py
msom, line 1: Place upgrades in the module where the model is defined (poeple, not fsi).

tests/onegov/fsi/test_upgrade.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from onegov.fsi.upgrade import parse_and_split_address_field
+
+
+@pytest.mark.parametrize('address,expected', [
```
msom, line 6: Does parametrizing increase the test count? Because this will imbalance the CI by adding many new and fast tests...

tests/onegov/fsi/test_upgrade.py
msom, line 1: This is in the wrong module (should be people, not fsi). Note that we generally don't test upgrades (because we can't, there might be a ticket for this.)

src/onegov/agency/forms/agency.py
```diff
@@ -33,6 +36,28 @@ class ExtendedAgencyForm(Form):
         render_kw={'rows': 10}
     )
 
+    location_address = TextAreaField(
+        label=_("Location address"),
+        render_kw={'rows': 1},
+    )
+    location_code_city = StringField(_("Location Code and City"))
```
msom, line 43: Please keep format within files consistent (using keyword label everywhere).

src/onegov/agency/models/agency.py
```diff
@@ -86,6 +86,24 @@ def portrait_html(self):
 
         return self.portrait
 
+    @staticmethod
+    def _get_html_paragraph_with_line_breaks(prop):
```
msom, line 90: Use `Markup`

src/onegov/agency/models/person.py
```diff
@@ -53,17 +53,27 @@ def phone_es(self):
                 result.append('0' + number[-9:])
         return [r for r in result if r]
 
-    @property
-    def address_html(self):
+    @staticmethod
+    def _get_html_paragraph_with_line_breaks(prop):
         return '<p>{}</p>'.format(
```
msom, line 58: Use `Markup`. Since you use this function at different files, maybe better add a function in `utils`

src/onegov/agency/templates/macros.pt
```diff
@@ -7,18 +7,70 @@
 <body>
 
     <metal:agency_card define-macro="agency_card" i18n:domain="onegov.agency">
-        <div class="agency-card agency-single-item" tal:define="memberships list(agency.memberships); children list(agency.children)">
+        <div class="agency-card agency-single-item" tal:define="memberships list(agency.memberships); children list(agency.children)" tal:attributes="class 'columns medium-12'">
```
msom, line 10: No need for `tal:attributes` since the content is not dynamic. Also, this conflicts with the already existing `class` attribute...

src/onegov/people/models/person.py
```diff
@@ -167,9 +185,24 @@ def vcard_object(self, exclude=None, include_memberships=True):
             line = result.add('url')
             line.value = self.website
 
-        if 'address' not in exclude and self.address:
+        if 'location_address' not in exclude and self.location_address:
```
msom, line 188: Have you tested this with the various vCard consumers such as outlook etc.? I think there were problems here previosuly, right @BreathingFlesh?

src/onegov/fsi/upgrade.py
```diff
@@ -172,3 +174,125 @@ def add_active_property_to_attendees(context):
...
+
+    # default
+    print(f'*** parse_and_split_address_field: no match found for address:'
+          f' {address}')
+    return location_addr, location_pcc, postal_addr, postal_pcc
+
+
+@upgrade_task('ogc-966 extend agency and person tables with more fields')
+def extend_agency_and_person_with_more_fields(context):
+    ################################################################
```
msom, line 245: We don't use such marking comments anywhere...

src/onegov/fsi/upgrade.py
```diff
@@ -172,3 +174,125 @@ def add_active_property_to_attendees(context):
...
+    # add columns to table 'agencies'
+    agencies_columns = ['email', 'phone', 'phone_direct', 'website',
+                        'location_address', 'location_code_city',
+                        'postal_address', 'postal_code_city',
+                        'opening_hours']
+    table = 'agencies'
+
+    for column in agencies_columns:
+        print(f'Checking table {table} for column {column} .. ')
```
msom, line 254: Don't use print statements

src/onegov/fsi/upgrade.py
```diff
@@ -172,3 +174,125 @@ def add_active_property_to_attendees(context):
...
+                default=lambda x: ''
+            )
+        else:
+            print(' -> column exists')
+
+    print(f"Migrate data from table {table} column 'address' field to "
+          f"'location_address', 'location_code_city', 'postal_address' and "
+          f"'postal_code_city ..")
+    for person in context.session.query(Person):
+        if not person.address:
```
msom, line 291: This has the potential to blow up during upgrade, which is the worst moment because then you have to manually fix and complete the upgrade (or somebody else). In general, you should avoid using models in upgrade steps, since they might have been changed in the meantime (some of our servers have updates only every half year). In the future, we should add some functionality for reliable data migration (e.g. using historical models), I think there is a ticket somewhere. In the meantime, I use this (error prone) pattern for data migration:

1. Add schema migration in an upgrade step + a CLI command for data migration.
2. Upgrade the server.
3. Run the CLI command (on the servers they need them)
4. Add a post-migration schema migration (e.g. dropping now obsolete columns)

msom (changes requested):

src/onegov/fsi/upgrade.py
Tschuppi81: @msom Can you please point me to such an example to get an idea how it can be done (the error prone way)?

src/onegov/fsi/upgrade.py
Tschuppi81: done

src/onegov/fsi/upgrade.py
Tschuppi81: removed

src/onegov/fsi/upgrade.py
Tschuppi81: removed

src/onegov/agency/templates/macros.pt
Tschuppi81: got it

src/onegov/fsi/upgrade.py
msom: Have a look here: https://github.com/OneGov/onegov-cloud/commit/cda513c9e713d36d894037a76cb667ea35dd0025 (only CLI relevant; there is no schema migration because newly added tables are created automatically) + Post-Step: https://github.com/OneGov/onegov-cloud/commit/73573b2e0fbeab7647ffbb96902d4397f4bb9421

src/onegov/people/models/person.py
Tschuppi81: I will validate the vcard. I found a tool https://l0b0.wordpress.com/2009/12/25/vcard-parser-and-validator/

tests/onegov/fsi/test_upgrade.py
Tschuppi81: I made this to ensure all addresses currently in /onegov_agency/bs, zg and ar will properly migrated to the new address scheme

msom: What I mean is: Why parametrizing the test and not just use the list of (address, expected) tuples within a single test? I suspect that parametetrized tests are counted as individual tests and therefore influence to test splitting in the CI (I might be wrong). This will lead to adding 20 new and very fast tests, which will probably been executed in one of the four processes (since test splitting is done by dividing to the number of processes). This one CI process will then have a shorter runtime compared to the other ones (and probably increasing the runtime at least one of the others and therefore the whole pipeline).

src/onegov/people/models/person.py
Tschuppi81: I brought street address, code and city to one address entry to make it readable on windows.

src/onegov/agency/forms/agency.py
Tschuppi81: Not sure what you exactly mean...

tests/onegov/fsi/test_upgrade.py
Tschuppi81: Will adjust it

src/onegov/fsi/upgrade.py
Tschuppi81: Understood that I should separate things. One open item is how to make sure the CLI command will be executed during next roll-out (and only once)?
And if I understood right, I would need a separate branch to do the 'post' (cleanup) steps for a further release.

msom: By running it after the upgrade (on those server that need that) by hand. Yes, the post step would be added to the next release.

We could think about to adding a similar framework for data migration. In comparison, this should be (1) much more robust and never break upgrades and (2) allow migrations for models, even when they have changed.

src/onegov/agency/forms/agency.py
msom: This is just a cosmetic (and more a general hint to try to keep things consistent within files/modules): Use e.g. `label=_("Location Code and City")`, if a few lines above labels are added using the `label` Argument.

Tschuppi81: Clear!

src/onegov/fsi/upgrade.py
Tschuppi81: Hope, my last changes are a big step for the separation of concern :smile:

Tschuppi81:
Btw: I did the following test run 
                                                                                
# prepare db locally (all agency schemas)                                       
onegov-core transfer --transfer-schema /onegov_agency/* --no-filestorage nike.seantis.ch
                                                                                
# upgrade and migrate people address field                                      
onegov-core --select /onegov_agency/* upgrade                                   
onegov-people --select /onegov_agency/* migrate-people-address-field           -> option --dry-run available

# => it workes for bs, zg, ar :tada:

msom (approved):

BreathingFlesh (changes requested):

src/onegov/agency/views/settings.py
```diff
@@ -83,8 +83,7 @@ class AgencySettingsForm(Form):
         label=_('Field used for internal phone numbers'),
         fieldset=_('Customize search results'),
         choices=[
-            ('phone_direct', _('Direct Phone Number')),
-            ('phone', _('Phone')),
+            ('phone_direct', _('Direct Phone Number or Mobile')),
```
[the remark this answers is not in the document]
Tschuppi81: Good finding, I removed 'phone' by mistake. Fixed it.

src/onegov/org/forms/person.py
```diff
@@ -20,17 +20,25 @@ class PersonForm(Form):
...
-        render_kw={'rows': 5}
+    location_address = TextAreaField(
+        label=_("Location address"),
+        render_kw={'rows': 1}
+    )
+    location_code_city = StringField(label=_("Location Code and City"))
+
+    postal_address = TextAreaField(
+        label=_("Postal address"),
+        render_kw={'rows': 2}
```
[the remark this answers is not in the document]
Tschuppi81: number of rows to be 2 for all location and postal address fields -> done

src/onegov/agency/forms/membership.py
```diff
@@ -75,7 +75,7 @@ def title(person):
                     person.phone_direct
                     or person.phone
                     or person.email
-                    or person.address
+                    or person.postal_address
```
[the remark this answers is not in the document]
Tschuppi81: I didn't want to make it to complex so I skipped the location address as it might be often not used...
Do you think I should add it?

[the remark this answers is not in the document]
Tschuppi81: @msom What do you think about @BreathingFlesh's input? Should I add the location_address for the person info?

PR #791, opened 2023-04-18
Skip migration for complex addresses, just take over

Tschuppi81:
agency: Skip migration for complex addresses, just take over

TYPE: Feature
LINK: ogc-966
HINT: for Basel Stadt (BS): sudo ogc people --select /agency_with_msal/bs migrate-people-address-field --dry-run


Tschuppi81:
Not needed anymore

PR #798, opened 2023-04-25
Ogc 740 directory entries

BreathingFlesh:

## Commit message

Town6: Display previous and next entries

TYPE: Feature
LINK: OGC-740

## Checklist


src/onegov/org/views/directory.py
```diff
@@ -565,10 +565,27 @@ def handle_change_request(self, request, form, layout=None):
...
+        directory,
+        published_only=not request.is_manager
+    ).query())
+
+    entry_index = siblings.index(self)
+
+    prev_entry = siblings[entry_index - 1] if entry_index != 0 else False
+    next_entry = siblings[
+        entry_index + 1] if entry_index != len(siblings) - 1 else False
+    more_entries = bool(prev_entry or next_entry)
```
msom, line 580: Should be `prev_entry is not False or next_entry is not False`, because prev_entry can be 0 (which converts to False)

src/onegov/town6/templates/macros.pt
```diff
@@ -651,12 +651,39 @@
                 </li>
             </ul>
         </div>
+
+        <div class="more-entries side-panel" tal:condition="more_entries|False">
+            <ul class="more-list">
+                <li tal:condition="prev_entry">
```
msom, line 657: prev_entry = 0?

msom (changes requested):

src/onegov/org/views/directory.py
BreathingFlesh: But prev_entry will contain the entry itself (if there is one), not the index. How can it be 0?

PR #802, opened 2023-05-01
Remove param for split

Tschuppi81:
Org: Minor code improvement

TYPE: Improvement
LINK: OGC-1076


cyrillkuettel (approved):
I even made a ticket for this, see OGC-1076 

(Was indeed quite overkill)

PR #803, opened 2023-05-01
Activate indent check only for form validation but not for viewing th…

Tschuppi81:
Org: Indent check only activated for new and edit forms but not for displaying.

Activate indent check while parsing form code only if originating from core validators which means there is either a new or an edited form code to be parsed. 

TYPE: Bugfix
LINK: ogc-1027

## Checklist

- Tests:
    - only affects new or edited form code
    - town6: forms
    - org: tickets for course registration
- Result:
    - Even wrongly indented form code will be parsed and displayed correctly. Once editing user needs to adjust the wrong indents for the old parts of the form code

### DIFF
--- a/src/onegov/form/parser/core.py
+++ b/src/onegov/form/parser/core.py
@@ -923,12 +923,16 @@ class CheckboxField(OptionsField, Field):
 
 
 @lru_cache(maxsize=1)
-def parse_formcode(formcode):
+def parse_formcode(formcode, enable_indent_check=False):
     """ Takes the given formcode and returns an intermediate representation
     that can be used to generate forms or do other things.
 
+    :param formcode: string representing formcode to be parsed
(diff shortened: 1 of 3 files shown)


msom (approved):

msom:
Optionally add a test.

Daverball:
Für mich ist etwas fragwürdig ob die Formcode-Validierung in dem Kontext überhaupt laufen sollte. Also das nicht nur der neue Check übersprungen werden sollte, sondern die komplette Validierung.

Tschuppi81:
Formcode validation is executed for viewing, editing or new forms, hence we cannot skip. 
As the solution of with the formcode parser is not optimal we opened a refactor ticket https://linear.app/onegovcloud/issue/OGC-1084/

PR #837, opened 2023-05-22
Org: Adds import/export for newsletter subscribers.

cyrillkuettel:
## Commit message

Org: Adds import/export for newsletter subscribers.

TYPE: Feature
LINK: OGC-1085

## Checklist

- [X] I have updated the PO files
- [X] I made changes/features for both org and town6


src/onegov/org/forms/newsletter.py
```diff
@@ -212,3 +220,102 @@ def recipient(self):
...
+        return {
+            'address': self.request.translate(_("Address")),
+        }
+
+    def run_export(self):
+        recipients = RecipientCollection(self.request.session)
+        headers = self.headers
+
+        def get(recipient, attribute):
+            result = (getattr(recipient, attribute, None))
```
msom, line 266: One pair of brackets to much, also use '' instead of None or else the next line might fail...

src/onegov/org/forms/newsletter.py
```diff
@@ -212,3 +220,102 @@ def recipient(self):
...
+                'application/zip',
+                'text/csv',
+                'text/plain',
+            }),
+            FileSizeLimit(10 * 1024 * 1024)
+        ],
+        render_kw=dict(force_simple=True)
+    )
+
+    @property
```
msom, line 255: Maybe use a cached_property? Then you can skip the headers = self.headers below

src/onegov/org/forms/newsletter.py
```diff
@@ -212,3 +220,102 @@ def recipient(self):
...
+                v: get(recipient, k)
+                for k, v in headers.items()
+            })
+        return result
+
+    def run_import(self):
+        headers = self.headers
+        session = self.request.session
+        recipients = RecipientCollection(session)
+        csvfile = convert_excel_to_csv(self.file.file)
```
msom, line 282: As far as I remember this function might throw a few exceptions which you might want to catch.... (https://github.com/OneGov/onegov-cloud/blob/7c06b111f30e73e8ed075bcb151ca7cb8b86b584/src/onegov/election_day/formats/common.py#L61)

src/onegov/org/forms/newsletter.py
```diff
@@ -212,3 +220,102 @@ def recipient(self):
                     .one()
 
         return NewsletterSendFormWithRecipients
+
+
+class NewsletterSubscriberImportForm(Form):
```
msom, line 225: Maybe a class name that reflect both import/export usage?

msom (approved):
Just a few suggestions you might want to consider. Otherwise merge if you like.

PR #862, opened 2023-06-05
Feature/ogc 1048 event export xml view

Tschuppi81:
event: Adds xml view for events

TYPE: Feature
LINK: ogc-1048

## Checklist

### DIFF
--- a/src/onegov/org/path.py
+++ b/src/onegov/org/path.py
@@ -1,8 +1,9 @@
 """ Contains the paths to the different models served by onegov.org. """
-
 import sedate
 
 from datetime import date
+from lxml import etree, objectify
+
 from onegov.chat import MessageCollection
 from onegov.chat import TextModule
 from onegov.chat import TextModuleCollection
@@ -456,6 +457,51 @@ def get_occurrences(app, request, page=0, range=None, start=None, end=None,
         locations=locations,
         only_public=(not request.is_manager)
     )
+
+
+class tags(etree.ElementBase):
+    """
+    Custom class as 'tag' is a member of class Element and cannot be
+    used as tag name.
+    """
+    def __init__(self, tags=()):
+        super().__init__()
+        self.tag = 'tags'
+
+        for t in tags:
+            tag = etree.Element('tag')
+            tag.text = t
+            self.append(tag)
+
+
+@OrgApp.view(model=OccurrenceCollection, name='xml')
+def events_export_xml(app, request):
+    xml = '<events></events>'
+
+    root = objectify.fromstring(xml)
+
+    for ev in EventCollection(app.session).query().all():
+        event = objectify.Element('event')
+        event.id = ev.id
+        event.title = ev.title
+        txs = tags(ev.tags)
+        event.append(txs)
+        event.description = ev.description
+        event.start = ev.start
+        event.end = ev.end
+        event.location = ev.location
+        event.price = ev.price
+        event.organizer = ev.organizer
+        event.event_url = ev.external_event_url
+        event.organizer_email = ev.organizer_email
+        event.modified = ev.last_change
+        root.append(event)
+
+    # remove lxml annotation
+    objectify.deannotate(root)
+    etree.cleanup_namespaces(root)
+
+    return etree.tostring(root, pretty_print=True, encoding='utf-8')
 
 
 @OrgApp.path(model=Occurrence, path='/event/{name}')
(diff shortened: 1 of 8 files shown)


src/onegov/org/path.py
msom, line 1: Please move this to views/occurence.py, path.py is only for paths, not views.

src/onegov/org/path.py
```diff
@@ -458,6 +459,51 @@ def get_occurrences(app, request, page=0, range=None, start=None, end=None,
...
+        self.tag = 'tags'
+
+        for t in tags:
+            tag = etree.Element('tag')
+            tag.text = t
+            self.append(tag)
+
+
+@OrgApp.view(model=OccurrenceCollection, name='xml')
+def events_export_xml(app, request):
```
msom, line 478: Move this function to the collection, e.g. `as_xml` (similar to `as_ical`).

src/onegov/org/path.py
```diff
@@ -458,6 +459,51 @@ def get_occurrences(app, request, page=0, range=None, start=None, end=None,
...
+        event.event_url = ev.external_event_url
+        event.organizer_email = ev.organizer_email
+        event.modified = ev.last_change
+        root.append(event)
+
+    # remove lxml annotations
+    objectify.deannotate(root, pytype=True, xsi=True, xsi_nil=True)
+    etree.cleanup_namespaces(root)
+
+    return etree.tostring(root, encoding='utf-8', xml_declaration=True,
```
msom, line 503: Is the content type of your response correct? I think you need to return a `morepath.Response` with the right `content_type`.

src/onegov/org/path.py
```diff
@@ -458,6 +459,51 @@ def get_occurrences(app, request, page=0, range=None, start=None, end=None,
...
+            self.append(tag)
+
+
+@OrgApp.view(model=OccurrenceCollection, name='xml')
+def events_export_xml(app, request):
+    xml = '<events></events>'
+    root = objectify.fromstring(xml)
+
+    for ev in EventCollection(app.session).query().all():
+        event = objectify.Element('event')
```
msom, line 483: Please also comment in the docstring off the new collection function the format or link to the definition (I guess anthrazit? Mabybe even name the function `as_antrazit_xml`?), so that nobody comes up with the bad idea to change things here without checking the format first...

msom (changes requested):

src/onegov/org/path.py
Tschuppi81: done

src/onegov/org/path.py
Tschuppi81: I did that and the browser recognizes the xml

msom:
Shouldn't this be in the `OccurrenceCollection` rather than the `EventCollection`? An event may contain a recurrence rule (e.g. "every week for the next 10 weeks"), in which case (10) occurrences will be created. I would expect the XML export either to contain the recurrence rule or all occurrences...

src/onegov/org/views/occurrence.py
```diff
@@ -190,6 +190,21 @@ def cors(response):
     ]
 
 
+@OrgApp.view(model=OccurrenceCollection, name='xml', permission=Public)
+def xml_export_events(self, request):
```
msom, line 194: How will this export been used? This public view is not linked everywhere...

msom:
What is the expectation regarding filtering? Currently you have the export in the `EventCollection` which probably export all events (even unpublished?). BTW: Using `self.query` in a collection will use all filters (e.g. given by the query parameters) but requires to use `self` in a view with the right model/collection. I guess what you really want is the export to be public; in the occurrence collection; honoring filters; being linked in the UI to the collection without enabled filters (`request.link(self, `xml`)?

src/onegov/org/views/occurrence.py
Tschuppi81: The export is used from an mobile app. 
To link it, I can add it to the existing 'export' list http://localhost:8080/onegov_winterthur/winterthur/events/export

Tschuppi81:
> Shouldn't this be in the `OccurrenceCollection` rather than the `EventCollection`? An event may contain a recurrence rule (e.g. "every week for the next 10 weeks"), in which case (10) occurrences will be created. I would expect the XML export either to contain the recurrence rule or all occurrences...

Good point, let me check on this...

Tschuppi81:
I will merge for today's release. Let's discuss on Monday what else I should improve as I am not a 100% sure about the comments...

PR #866, opened 2023-06-06
mypy: Adds type annotations for onegov.core.collection

Daverball:
## Commit message

mypy: Adds type annotations for onegov.core.collection

## Checklist


src/onegov/core/collection.py
```diff
@@ -7,89 +7,133 @@
...
+    from collections.abc import Collection, Iterable, Iterator, Sequence
+    from sqlalchemy import Column
+    from sqlalchemy.orm import Query, Session
+    from sqlalchemy.sql.elements import ClauseElement
+    from typing_extensions import Self
+    from uuid import UUID
 
-    def __init__(self, session):
+    from onegov.core.orm import Base
+    from onegov.form.core import Form
```
Daverball, line 21: For some reason this introduces a runtime dependency, even though it's only imported when type checking...

Maybe this requires using sys.import?!

Daverball: Unfortunately if we use a dynamic way to import the module, mypy will not understand what type the imported symbol is, so for now we just depend on wtforms.Form, even though we need our own Form class which implements `get_useful_data`

Daverball:
The failure on the latest commit appears to be a flaky test, no runtime code was changed and that failure appeared without being present in master, so I will merge this

PR #892, opened 2023-06-16
Org: Send mail for rejected reservations.

cyrillkuettel:
## Commit message

Org: Send mail for rejected reservations.

TYPE: Feature
LINK: OGC-946

## Screenshot
Neu auch für abgelehnte Reservationen:

out

## Checklist


src/onegov/org/views/reservation.py
```diff
@@ -790,6 +790,50 @@ def reject_reservation(self, request, text=None, notify=False):
...
+    q = q.order_by(None).order_by(ResourceRecipient.address)
+    q = q.with_entities(ResourceRecipient.address, ResourceRecipient.content)
+    recipients = [
+        r.address
+        for r in q
+        if (
+            self.resource.hex in r.content['resources']
+            and r.content.get('rejected_reservations', {})
+        )
+    ]
```
Daverball, line 805: Consider changing this to a generator expression, since you don't use the list for anything.

You could also further improve things by using `send_transactional_email_batch`, which is more efficient than calling `send_transactional_email` multiple times, since each send will generate a file on the disk and a datamanager in the current transaction. Look for other places where we use `send_transactional_email_batch` to see how this works.

src/onegov/org/views/ticket.py
```diff
@@ -346,15 +348,81 @@ def thread():
...
+        'ticket_reference': ticket.reference(request)
+    }
+    content = render_template(template, request, args)
+
+    for r in internal_notes_recipients:
+        request.app.send_transactional_email(
+            subject=args['title'],
+            receivers=(r),
+            content=content,
+        )
```
Daverball, line 404: Same thing here

Daverball (changes requested):

src/onegov/org/views/reservation.py
```diff
@@ -651,48 +652,53 @@ def accept_reservation(self, request, text=None, notify=False):
...
+                        'title': title,
+                        'form': form,
+                        'model': self,
+                        'resource': resource,
+                        'reservations': reservations,
+                        'show_submission': show_submission,
+                        'message': message,
+                    },
+                )
+                plaintext = html_to_text(content)
```
Daverball, line 690: You should move this outside the loop, the title,  content and plaintext are the same for all the emails.

cyrillkuettel: Yes (!) I can't believe I didn't see that.

cyrillkuettel: The two methods in `ticket.py` and `reservation.py` share a lot of characteristics (in fact, differ only in detail) it might be another improvement to pull the logic upwards and extract this into a shared method. I'm not sure if it's a good idea though, maybe a bit overkill?

Daverball: You could probably pull `mail_iter` out into a utility function for mail batches that have static content.

It would accept an `Iterable[str]` of recipients, as well as `subject`, `html`, `plaintext` where you only need to provide either `html` or `plaintext`, if only `html` is given, `plaintext` is created using `html2text`.

src/onegov/org/forms/resource_recipient.py
```diff
@@ -57,6 +57,14 @@ class ResourceRecipientForm(Form):
                       "above."),
     )
 
+    rejected_reservations = BooleanField(
+        label=_("Rejected Reservations"),
+        fieldset=_("Notifications *"),
+        description=("Bei der Stornierung einer Reservation wird eine "
+                     "Benachrichtigung an den obenstehenden Empfänger "
+                     "gesendet."),
```
Daverball, line 65: Please use a translation string.

src/onegov/org/locale/de_CH/LC_MESSAGES/onegov.org.po
```diff
@@ -3628,7 +3640,7 @@ msgstr ""
 "Suche im Navigationsbereich."
 
 msgid "Back"
-msgstr "Zurück"
+msgstr "zurück"
```
Daverball, line 3643: Why did this change to lowercase?

src/onegov/org/locale/de_CH/LC_MESSAGES/onegov.org.po
```diff
@@ -3418,6 +3421,15 @@ msgstr "Anmeldung"
...
 msgstr "Ihre Ticket Referenz"
 
+msgid "The following reservations have been rejected:"
+msgstr ""
+
+msgid ""
+"This is a notification for the rejected reservations for ${organisation}. If "
+"you no longer wish to receive these notifications, please contact an "
+"administrator so they can remove you from the recipients list."
+msgstr ""
```
Daverball, line 3431: It's possible I accidentally removed these when merging master. These translations seem to exist for the other languages.

src/onegov/org/locale/fr_CH/LC_MESSAGES/onegov.org.po
```diff
@@ -3808,6 +3824,9 @@ msgstr ""
 msgid "Receives notifications for internal notes on reservations."
 msgstr "Reçoit des notifications pour les notes internes sur les réservations."
 
+msgid "Receives notifications for rejected reservations."
+msgstr ""
```
Daverball, line 3828: This was missing from French and Italian translations

src/onegov/org/views/reservation.py
Daverball: I think merging this with master somehow brought this back to an older version.

src/onegov/town6/locale/de_CH/LC_MESSAGES/onegov.town6.po
```diff
@@ -1407,6 +1407,15 @@ msgstr "Anmeldezeitraum"
...
 msgstr "Ihre Ticket Referenz"
 
+msgid "The following reservations have been rejected:"
+msgstr "Folgende Reservationen sind abgelehnt worden:"
+
+msgid ""
+"This is a notification for the rejected reservations for ${organisation}. If "
+"you no longer wish to receive these notifications, please contact an "
+"administrator so they can remove you from the recipients list."
+msgstr ""
```
Daverball, line 1417: This is missing a translation in town6.

Daverball (changes requested):

src/onegov/org/locale/fr_CH/LC_MESSAGES/onegov.org.po
cyrillkuettel: It's because the branch is based on `ogc-1068-interne-notizen`, not master. I did this preemptive merge because I expected this would avoid conflicts due to overlapping modifications. 
In hindsight that was clearly not the best idea. I will cherry-pick the changes and merge manually

PR #918, opened 2023-07-13
Feature/ogc 508 replace elastic search by postgres text search

Tschuppi81:
INITIAL FEEDBACK TO TEXT SEARCH IN POSTGRES - WORK IN PROGRESS

Feature/ogc 508 replace elastic search by postgres text search                  
                                                                                                                                                                
core: Replace elastic search by postgres full text search                       
                                                                                
With this changes elastic search (es) is still in place and productive while we introduce the postgres full text search accessible using 'search_postgres' keyword in the url.
es url: http://localhost:8080/onegov_agency/bs/search?q=Arnold                  
psql url: http://localhost:8080/onegov_agency/bs/search_postgres?q=Arnold        

hint: onegov-core upgrade
                                          
Main changes:                                                                   
- Re-indexing                                                                   
../onegov-cloud/src/onegov/search/cli.py                                        
- Upgrade script (uses the same code as re-indexing)                            
../onegov-cloud/src/onegov/search/upgrade.py                                    
- Extension of mixin class Searchable                                           
.. /onegov-cloud/src/onegov/search/mixins.py                                    
- Adjusting all searchable models adding the fts index in postgres                                                                         
                                                                                
Open Items:                                                                     
- unit tests are not reworked yet                                               
- language configuration                                                        
- improve ranking                              
                                                                                
                                                                                                                                         

## Checklist


src/onegov/activity/models/attendee.py
```diff
@@ -43,6 +45,20 @@ class Attendee(Base, TimestampMixin, ORMSearchable):
...
+    def search_score(self):
+        return 3
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on columns username, name and notes
+        """
+        return Searchable.create_tsvector_string('username', 'name', 'notes')
+
```
Daverball, line 61: See comment on `Activity`.

src/onegov/core/orm/session_manager.py
```diff
@@ -683,6 +683,8 @@ def ensure_schema_exists(self, schema: str) -> None:
                 for base in self.bases:
                     base.metadata.schema = None
 
+                conn.execute('COMMIT')
+
```
Daverball, line 687: Explain why you are doing this with a comment. This looks fragile. Are you working around exceptions here to ensure the commit always happens?

Maybe we should switch to using conn as a context manager i.e. `with conn:` and get rid of the manual commit.

src/onegov/form/models/definition.py
```diff
@@ -115,6 +119,23 @@ class FormDefinition(Base, ContentMixin, TimestampMixin, Extendable):
...
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on column ticket number
+        """
+        s = Searchable.create_tsvector_string('title')
+        s += " || ' ' || coalesce(((meta ->> 'lead')), '')"
+        s += " || ' ' || coalesce(((content ->> 'text')), '')"
+        return s
+
```
Daverball, line 138: See comment on `Acitivity`.

src/onegov/fsi/models/course_attendee.py
```diff
@@ -107,6 +109,21 @@ def __str__(self):
...
+    )
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on the following columns
+        """
+        return Searchable.create_tsvector_string('first_name', 'last_name',
+                                                 'organisation', '_email')
+
```
Daverball, line 126: See comment on `Acitivity`.

src/onegov/fsi/models/course_event.py
```diff
@@ -20,15 +21,29 @@
...
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from .course import Course  # noqa: F811
+    from .course_notification_template import (  # noqa: F811
+        CancellationTemplate,   # noqa: F811
+        CourseNotificationTemplate,  # noqa: F811
+        InfoTemplate,  # noqa: F811
+        ReminderTemplate,   # noqa: F811
+        SubscriptionTemplate  # noqa: F811
```
Daverball, line 46: This looks like a merge conflict resolution error, you don't need this if statement twice...

src/onegov/fsi/models/course_event.py
```diff
@@ -176,6 +191,19 @@ def localized_end(self):
...
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on the following columns
+        """
+        return Searchable.create_tsvector_string('location',
+                                                 'presenter_name',
+                                                 'presenter_company',
+                                                 'presenter_email')
+
```
Daverball, line 206: See comment on `Acitivity`.

src/onegov/landsgemeinde/models/agenda.py
```diff
@@ -103,6 +104,21 @@ class AgendaItem(
...
+        """
+        index is built on column title as well as  on the json
+        fields overview, text and resolution in content column
+        """
+        s = Searchable.create_tsvector_string('title')
+        s += " || ' ' || coalesce(((content ->> 'overview')), '')"
+        s += " || ' ' || coalesce(((content ->> 'text')), '')"
+        s += " || ' ' || coalesce(((content ->> 'resolution')), '')"
+        return s
+
```
Daverball, line 121: See comment on `Acitivity`.

src/onegov/landsgemeinde/models/assembly.py
```diff
@@ -94,5 +95,15 @@ def es_suggestion(self):
...
+    # column for full text search index
+    fts_idx = Column(TSVECTOR)
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on the json field overview in content column
+        """
+        return "coalesce(((content ->> 'overview')), '')"
+
```
Daverball, line 107: See comment on `Activity`.

src/onegov/landsgemeinde/models/votum.py
```diff
@@ -97,6 +98,24 @@ def es_suggestion(self):
...
+        """
+        s = Searchable.create_tsvector_string('person_name',
+                                              'person_function',
+                                              'person_place',
+                                              'person_political_affiliation')
+        s += " || ' ' || coalesce(((content ->> 'text')), '')"
+        s += " || ' ' || coalesce(((content ->> 'motion')), '')"
+        s += " || ' ' || coalesce(((content ->> 'statement_of_reasons')), '')"
+        return s
+
```
Daverball, line 118: See comment on `Activity`.

src/onegov/newsletter/models.py
```diff
@@ -86,6 +88,20 @@ def validate_name(self, key, name):
...
+    def search_score(self):
+        return 6
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on columns title, lead, html
+        """
+        return Searchable.create_tsvector_string('title', 'lead', 'html')
+
```
Daverball, line 104: See comment on `Activity`.

src/onegov/org/models/external_link.py
```diff
@@ -48,6 +50,23 @@ class ExternalLink(Base, ContentMixin, TimestampMixin, AccessExtension,
...
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        builds the index on column title and lead within meta column
+        email.
+        """
+        s = Searchable.create_tsvector_string('title')
+        s += " || ' ' || coalesce(((meta ->> 'lead')), '')"
+        return s
+
```
Daverball, line 69: See comment on `Activity`.

src/onegov/page/model.py
```diff
@@ -5,19 +5,40 @@
...
+    def psql_tsvector_string():
+        """
+        index is built on column title as well as on the json
+        fields lead and text in content column if not NULL
+        """
+        s = Searchable.create_tsvector_string('title')
+        s += " || ' ' || coalesce(((content ->> 'lead')), '')"
+        s += " || ' ' || coalesce(((content ->> 'text')), '')"
+        return s
+
```
Daverball, line 41: See comment on `Activity`.

src/onegov/people/models/agency.py
```diff
@@ -81,6 +83,17 @@ class Agency(AdjacencyList, ContentMixin, TimestampMixin, ORMSearchable,
...
+    fts_idx = Column(TSVECTOR)
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        builds the index on columns title, description and portrait.
+        """
+        return Searchable.create_tsvector_string('title', 'description',
+                                                 'portrait')
+
```
Daverball, line 96: See comment on `Activity`.

src/onegov/people/models/membership.py
```diff
@@ -88,6 +90,20 @@ class AgencyMembership(Base, ContentMixin, TimestampMixin, ORMSearchable,
...
+    def search_score(self):
+        return 3
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        builds the index on column title.
+        """
+        return Searchable.create_tsvector_string('title')
+
```
Daverball, line 106: See comment on `Activity`.

src/onegov/people/models/person.py
```diff
@@ -129,6 +130,21 @@ def spoken_title(self):
...
+        """
+        builds the index on columns first name, last name, function and email.
+        """
+        return Searchable.create_tsvector_string('first_name', 'last_name',
+                                                 'function', 'email')
+
+    @property
+    def search_score(self):
+        return 3
+
```
Daverball, line 147: See comment on `Activity`.

src/onegov/reservation/models/custom_allocation.py
```diff
@@ -5,7 +5,7 @@
 
 
 class CustomAllocation(Allocation, ModelBase):
-    __mapper_args__ = {'polymorphic_identity': 'custom'}  # type:ignore
+    __mapper_args__ = {'polymorphic_identity': 'custom'}
```
Daverball, line 8: Another merge conflict resolution error, this ignore is necessary on latest libres version.

src/onegov/reservation/models/custom_reservation.py
```diff
@@ -7,7 +7,7 @@
 
 
 class CustomReservation(Reservation, ModelBase, Payable):
-    __mapper_args__ = {'polymorphic_identity': 'custom'}  # type:ignore
+    __mapper_args__ = {'polymorphic_identity': 'custom'}
```
Daverball, line 10: Another merge conflict resolution error, this ignore is necessary on latest libres version.

src/onegov/reservation/models/resource.py
```diff
@@ -138,6 +138,10 @@ class Resource(ORMBase, ModelBase, ContentMixin, TimestampMixin):
     #: the view to open in the calendar (fullCalendar view name)
     view: str | None = 'month'
 
+    @property
+    def search_score(self):
+        return 4
+
```
Daverball, line 144: See comment on `Allocation`.

src/onegov/search/upgrade.py
```diff
@@ -2,3 +2,30 @@
...
+    # onegov-core --select /onegov_org/risch upgrade
+    # onegov-core --select /onegov_org/* upgrade
+    # onegov-core --select /onegov_town6/meggen upgrade
+    # onegov-core --select /onegov_town6/* upgrade
+    # onegov-core upgrade
+
+    session = context.session
+    schema = context.schema
+
+    for model in searchable_sqlalchemy_models(Base):
```
Daverball, line 29: Shouldn't this be `Searchable` since `add_fts_column` is defined on that, and not on `Base`?

src/onegov/ticket/model.py
```diff
@@ -73,6 +75,20 @@ class Ticket(Base, TimestampMixin, ORMSearchable):
...
+    def search_score(self):
+        return 6
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on column ticket number
+        """
+        return Searchable.create_tsvector_string('number')
+
```
Daverball, line 91: See comment on `Allocation`.

src/onegov/town6/templates/directory.pt
Daverball, line 1: This looks like another merge conflict resolution error, checkout file from master.

src/onegov/translator_directory/models/translator.py
```diff
@@ -188,6 +189,17 @@ def expertise_professional_guilds_all(self):
...
+    fts_idx = Column(TSVECTOR)
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on the following columns
+        """
+        return Searchable.create_tsvector_string('first_name', 'last_name',
+                                                 'email')
+
```
Daverball, line 202: See comment on `Allocation`.

src/onegov/user/models/user.py
```diff
@@ -144,10 +145,15 @@ def userprofile(self) -> list[str]:
...
-    )
+    fts_idx = Column(TSVECTOR)
+
+    @property
+    def search_score(self) -> int:
+        return 5
+
+    @staticmethod
+    def psql_tsvector_string() -> str:
+        return Searchable.create_tsvector_string('username', 'realname')
```
Daverball, line 156: See comment on `Allocation`.

src/onegov/search/mixins.py
```diff
@@ -134,6 +134,122 @@ def es_tags(self):
...
+        raise NotImplementedError
+
+    @staticmethod
+    def reindex(session, schema, model):
+        """
+        Re-indexes the table by dropping and adding the full text search
+        column.
+        """
+        Searchable.drop_fts_column(session, schema, model)
+        Searchable.add_fts_column(session, schema, model)
```
Daverball, line 170: Since we are doing that anyways, do we even need the upgrade step? It seems simpler to just ignore missing columns in `drop_fts_column`.

src/onegov/search/mixins.py
```diff
@@ -134,6 +134,122 @@ def es_tags(self):
...
+        json fields to be used for full text search index.
+
+        Example:
+        s = create_tsvector_string('title', 'location')
+        s += " || ' ' || coalesce(((content ->> 'description')), '')"
+        s += " || ' ' || coalesce(((content ->> 'organizer')), '')"
+
+        :return: tsvector string
+        """
+        raise NotImplementedError
```
Daverball, line 161: See my comment on `Activity`. It seems to me like we should be able to generate the `tsvector_string` entirely based on what was previously available in terms of `es_` properties.

I technically don't mind if we move this complexity to the models, but then we need to be able to merge conflicting tsvectors into one that adds cases based on the polymorphic identity.

Daverball (changes requested):

src/onegov/search/upgrade.py
Daverball: Actually never mind. But you may need to also search for `libres.db.ORMBase`

src/onegov/search/mixins.py
Tschuppi81: Good point, we can skip the upgrade step. I will test it!

Daverball: You also might need to be more careful about polymorphism here, so you don't reindex the same table multiple times, since more than one Model can share the same table. So you might want to factor this a little differently.

setup.cfg
```diff
@@ -136,7 +136,7 @@ install_requires =
     pyyaml
     rcssmin
     redis
-    reportlab
```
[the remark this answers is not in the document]
Tschuppi81: pinning removed

src/onegov/activity/models/activity.py
```diff
@@ -101,6 +102,24 @@ class Activity(Base, ContentMixin, TimestampMixin):
...
+    def psql_tsvector_string():
+        """
+        index is built on columns title and location as well as the json
+        fields description and organizer in content column
+        """
+        s = Searchable.create_tsvector_string('title', 'location')
+        s += " || ' ' || coalesce(((meta ->> 'lead')), '')"
+        s += " || ' ' || coalesce(((content ->> 'text')), '')"
+        return s
+
```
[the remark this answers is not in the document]
Tschuppi81: Correct, I accidentally added it to the base class of `VacationActivity.`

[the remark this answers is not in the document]
Tschuppi81: Moved fts index from Activity to VactionActivity and verified that activites for feriennet have the fts index column where as activities for communities e.g. meggen have no fts index column

src/onegov/core/upgrades.py
```diff
@@ -63,7 +63,7 @@ def json_columns(cls: type[Any]) -> 'Iterator[Column[Any]]':
             # TODO: we should change find_models to operate on
             #       sqlalchemy.orm.DeclarativeBase once we upgrade
             #       to SQLAlchemy 2.0
-            ORMBase,  # type: ignore[arg-type]
+            ORMBase,
```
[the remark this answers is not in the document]
Tschuppi81: added again

src/onegov/core/orm/session_manager.py
Tschuppi81: This line came in while fixing tests. The line is not needed.

src/onegov/core/upgrades.py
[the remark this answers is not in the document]
Daverball: @Tschuppi81 not quite right still. Generally I'd just recommend fixing these by checking the file out from master, i.e. reverting to the original state, then nothing can go wrong.
```shell
git checkout master -- src/onegov/core/upgrades.py
```

Daverball:
@Tschuppi81 As a side note you probably should merge master again (or at least cherry pick 25d7d0e8436c537c8a314c0b742ebe0385fafed3 to skip the currently broken browser tests.

src/onegov/core/upgrades.py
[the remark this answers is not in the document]
Tschuppi81: I did, but mypy complains about it

reto@reto-pc:~/workspace/onegov-cloud$ git checkout master -- src/onegov/core/upgrades.py
...
src/onegov/core/upgrades.py:66: error: Unused "type: ignore" comment  [unused-ignore]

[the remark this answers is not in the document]
Daverball: Like I said, it requires the newest version of libres. So make sure your venv is up to date.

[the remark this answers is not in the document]
Tschuppi81: with the latest master and updated venv it works

src/onegov/search/upgrade.py
Tschuppi81: But both base classes Àllocation`and `Reservation`in libres do not define es_properties` as well as the two customization classes `CustomAllocation`and `CustomAllocation`. Shall I still add it to 'searchable' models?

src/onegov/search/mixins.py
Tschuppi81: I tried using `hasattr` and `getattr` and `isinstance` but the `content_property()`s are special not real attributes. Also the type information cannot be used to distinguish beetween meta and content. Any idea how to identify if the property is in meta or content column?
```
(Pdb) self
<class 'onegov.org.models.page.Topic'>
(Pdb) getattr(self, 'title')
<sqlalchemy.orm.attributes.InstrumentedAttribute object at 0x7f485f7e58f0>
(Pdb) getattr(self, 'lead', None) == None
True
``

Below the snipped of the content_property lead
```class Topic(Page, TraitInfo, SearchableContent, AccessExtension,
            PublicationExtension, VisibleOnHomepageExtension,
            ContactExtension, ContactHiddenOnPageExtension,
            PersonLinkExtension, CoordinatesExtension, ImageExtension):
    __mapper_args__ = {'polymorphic_identity': 'topic'}

    es_type_name = 'topics'

    lead = content_property()
``

Comment (Daverball) [reply src/onegov/search/mixins.py:161]: `content_property` and `meta_property` are just factory functions which create a `onegov.core.orm.mixins.content.dict_property` so in this example you would be able to look at `cls.lead.attribute` and see that it is set to `content`, which would tell you that the value is stored inside `content`.

Comment (Daverball) [reply src/onegov/search/mixins.py:161]: The other complication is that there's some magic in the Declarative ORM metaclass, you probably have to look at the mapper to get the real property, i.e. 
```python
from sqlalchemy import inspect
getattr(inspect(cls), 'lead', None)
```

Tschuppi81: I managed to bring all tsvector code to mixins except one

src/onegov/town6/templates/directory.pt
Tschuppi81: done

src/onegov/feriennet/models/activity.py
```diff
@@ -21,9 +24,28 @@ class VacationActivity(Activity, CoordinatesExtension, SearchableContent):
...
+    def psql_tsvector_string(model):
+        """
+        index is built on column title as well as the json
+        fields lead and text in meta resp. content column
+        """
+        s = Searchable.create_tsvector_string('title')
+        s += " || ' ' || coalesce(((meta ->> 'lead')), '')"
+        s += " || ' ' || coalesce(((content ->> 'text')), '')"
+        return s
+
```
Daverball, line 48: This should already be handled by the Mixin, no?

To make `organiser` work turn it into a `hybrid_property`. Writing the equivalent in SQL isn't trivial, but it should be possible with some combination of arrays, json operators and coalesce.

You could probably also get away with implementing something that isn't fully equivalent (i.e. contains some elements with empty strings or returns a concatenated string where all the keywords are separated by at least one space)

src/onegov/search/mixins.py
```diff
@@ -134,6 +136,139 @@ def es_tags(self):
...
+            Searchable.multi_language_tsvector_expression(
+                model.psql_tsvector_string(model))
+
+        query = f"""
+            ALTER TABLE "{schema}".{model.__tablename__}
+            ADD COLUMN IF NOT EXISTS {col_name} tsvector GENERATED ALWAYS AS
+            ({tsvector_expression}) STORED;
+        """
+        session.execute(query)
+        session.execute("COMMIT")
```
Daverball, line 243: We probably need to do something more sophisticated in here to properly handle polymoprhic models.
1.  When you determine all the searchable models get their polymorphic base class and use that to deduplicate the polymorphic variants of the models and only pass in non-polymorphic or polypmorphic base classes into this function.
2. Inside this function retrieve all the polymorphic subclasses which derive from Searchable.
3. Create a CASE statement which maps each polymorphic model's ts_vector_expression to the polymorphic identity

Steps 2/3 are only necessary if the model is polymorphic, otherwise you can use your simple version.

You can probably simplify 3 a little bit by merging all the cases which produce an identical tsvector_statement. The Postgres query compiler will probably do that for you, but we probably should still try to generate the most simple expression possible.

src/onegov/search/upgrade.py
Daverball, line 1: Let's revert this to master, no changes necessary.

src/onegov/ticket/model.py
```diff
@@ -107,9 +116,10 @@ def created(cls):
         'title': {'type': 'text'},
         'subtitle': {'type': 'text'},
         'group': {'type': 'text'},
-        'ticket_email': {'type': 'keyword'},
-        'ticket_data': {'type': 'localized_html'},
-        'extra_localized_text': {'type': 'localized'}
+        # properties cannot be added to search index
+        # 'ticket_email': {'type': 'keyword'},
+        # 'ticket_data': {'type': 'localized_html'},
+        # 'extra_localized_text': {'type': 'localized'}
```
Daverball, line 122: Don't modify the search properties. You can make them work with tsvector by changing them to hybrid_properties. The function that generates the tsvector should probably emit an warning if it encounters a property name which it cannot translate to a tsvector statement. In the future it should be a hard failure, but we definitely don't want to change the search results for the existing elastic search.

src/onegov/translator_directory/models/translator.py
```diff
@@ -188,6 +189,17 @@ def expertise_professional_guilds_all(self):
...
+    # column for full text search index
+    fts_idx = Column(TSVECTOR)
+
+    @staticmethod
+    def psql_tsvector_string():
+        """
+        index is built on the following columns
+        """
+        return Searchable.create_tsvector_string('first_name', 'last_name',
+                                                 'email')
```
Daverball, line 201: This should be handled by the Mixin.

Daverball (changes requested):

src/onegov/search/mixins.py
```diff
@@ -134,6 +136,139 @@ def es_tags(self):
...
+                   and isinstance(model.__dict__[p], InstrumentedAttribute)]
+        s = Searchable.create_tsvector_string(*columns) if columns else ''
+
+        # identify content and meta properties
+        for p in model.es_properties:
+            if p in model.__dict__ and isinstance(model.__dict__[p],
+                                                  content.dict_property):
+                s += join if s else ''
+                s += f"coalesce((({model.__dict__[p].attribute} " \
+                     f"->> '{p}')), '')"
```
Daverball, line 177: Maybe we can simplify this by making `dict_property` a subclass of `hybrid_property` which implicitly generates the JSON access syntax.

Daverball: Also this could go wrong if the name of the attribute doesn't match the name of the column. So you should retrieve the column from the model using attribute, and use the column to generate the first part of the query, rather than trusting that the names match.

src/onegov/search/mixins.py
```diff
@@ -134,6 +136,139 @@ def es_tags(self):
...
+        return s
+
+    @staticmethod
+    def reindex(session, schema, model):
+        """
+        Re-indexes the table by dropping and adding the full text search
+        column.
+        """
+        Searchable.drop_fts_column(session, schema, model)
+        Searchable.add_fts_column(session, schema, model)
```
Daverball, line 188: Since `add_fts_column` needs to be more sophisticated to be correct it would probably be a good idea to assert that the model that was passed in is either not polymorphic, or a polymorphic base class.

You could also just do the same assertion in `drop_fts_column` and `add_fts_column`. Although `drop_fts_column` can afford to be a little bit more forgiving.

src/onegov/search/mixins.py
Daverball: I think I will try my hand at this. Rather than deriving from hybrid_property I can just extend dict_property to behave the same as a hybrid_property (i.e. returning a SQL expression when accessed from the class, rather than an instance), we won't be able to handle custom getters, without also defining a custom expression, but that's fine, in that case we can just fall back to what it did before.

Tschuppi81: No problem, I am happy if you support here.

src/onegov/search/mixins.py
```diff
@@ -134,6 +136,139 @@ def es_tags(self):
...
+    @property
+    def search_score(self):
+        """
+        the lower the score they higher the class type will be shown in search
+        results. Default is 10 (lowest)
+        """
+        return 10
+
+    def __lt__(self, other):
+        return self.search_score < other.search_score
```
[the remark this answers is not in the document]
Tschuppi81: Yes, I do use it for sorting search results. I will go for your suggestion.

src/onegov/search/mixins.py
Tschuppi81: I don't get your idea fully. 

Note that I check the __tablename__ to ensure only re-indexing each table once...  -> see ../onegov-cloud/src/onegov/search/integration.py +L401 in psql_perform_reindex()

src/onegov/search/upgrade.py
Tschuppi81: done

src/onegov/search/mixins.py
Daverball: Consider the following example:
```
class User(Base):
    __tablename__ = 'users'
    __mapper__args = {'polymorphic_on': 'type', 'polymorphic_identity': 'generic'}
    type = Column(String, nullable=False, default=lambda: 'generic')
    username = Column(String, nullable=False)
    email = Column(String, nullable=False)

class Person(User, Searchable):
    es_properties = {'username': {'type': 'text'}}
    es_public = True

class Participant(User, Searchable):
    es_properties = {'username': {'type': 'text'}, 'email': {'type': 'text'}}
    es_public = False
```

Now you have three polymorphic variations that each have their own requirements. The base class isn't supposed to show up in results at all, the two derived classes are searchable, but the first only allows searching for the username, while the other also allows searching for email but as a trade-off the results are only visible to logged in users.

With your naive approach of generating one ts_vector_expression per table you don't know which expression it will pick up, and the results will be wrong for all the polymorphic variants it did not pick.

You can use `sqlalchemy.inspect` to get the `Mapper` associated with the model, which will give you access to properties such as `polymorphic_identity` and `self_and_descendants` to get all the polymorphic variations of the class, starting from the current variant.

Daverball (changes requested):
I don't think this is necessary, if you properly escape the column name (i.e. enclose it in double quotes) inside your SQL statement then you won't have to rename it.

src/onegov/search/mixins.py
Daverball: the improved `dict_property` has been merged into master, so now you shouldn't need to do this second loop anymore, the first loop should be sufficient.

Tschuppi81:
Escaping in postgres is done using quotes around column name \"<column>\"

src/onegov/search/mixins.py
```diff
@@ -258,10 +258,12 @@ def create_tsvector_string(*cols):
 
         :param cols: column names to be indexed
         :return: tsvector string for multiple columns i.e. for columns title
-        and body: coalesce(title, '') || ' ' || coalesce(body, '')
+        and body: coalesce([title], '') || ' ' || coalesce([body], '')
+        Note that the '\"{}\"' escapes the column name in case it is a
+        keyword, see http://www.postgresql.org/docs/current/static/sql-keywords-appendix.html
         """
-        base = "coalesce({}, '')"
-        ext = " || ' ' || coalesce({}, '')"
```
Daverball: I don't think this will work, since you won't get strings as inputs, it happens to work for columns, but in general you are probably better off using `sqlalchemy.func.coalesce` and `sqlalchemy.func.concat` on the input. That will ensure you will generate valid SQL statement, regardless of whether the column is an actual column, or some arbitrarily complex expression (which is what you will get with `hybrid_property`).

Daverball: This also should do the necessary escaping on column names for you.

src/onegov/search/mixins.py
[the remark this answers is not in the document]
Tschuppi81: Done

src/onegov/feriennet/models/activity.py
Tschuppi81: Correct, removed

src/onegov/activity/models/attendee.py
```diff
@@ -43,6 +45,13 @@ class Attendee(Base, TimestampMixin, ORMSearchable):
     }
     es_public = False
 
+    # column for full text search index
+    fts_idx = Column(TSVECTOR)
```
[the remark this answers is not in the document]
Tschuppi81: Finally managed, thanks @Daverball

src/onegov/translator_directory/models/translator.py
Tschuppi81: done

src/onegov/org/models/resource.py
Daverball, line 1: Looks like that stayed in there by accident

src/onegov/agency/models/person.py
```diff
@@ -18,9 +18,15 @@ def es_public(self):
...
     es_properties = {
+        'first_name': {'type': 'text'},
+        'last_name': {'type': 'text'},
         'title': {'type': 'text'},
         'function': {'type': 'localized'},
         'email': {'type': 'text'},
+        'phone': {'type': 'text'},
+        'phone_direct': {'type': 'text'},
+        # FIXME: properties cannot be used for ts tsvector, using phone and
+        # phone_direct instead
```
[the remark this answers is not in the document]
Tschuppi81: Reverted to original es_properties

src/onegov/ticket/model.py
Tschuppi81: hybrid_property support added while generating tsvector string and re-enabled all properties in Ticket

src/onegov/search/mixins.py
Tschuppi81: Adjusted tsvector string generation to support hybrid_properties

src/onegov/org/models/resource.py
Tschuppi81: removed

src/onegov/search/mixins.py
Tschuppi81: second loop completely removed

src/onegov/search/mixins.py
Daverball: This is still incorrect, you shouldn't use strings at all, you should build the expression using the sqlalchemy functions, as long as you use string interpolation, it won't work correctly in every case.

So `base` would look something like `base = func.coalesce(expr, '')` and `ext` something like `func.concat(left_expr, func.concat(' ', func.coalesce(right_expr, ''))`

src/onegov/search/mixins.py
Tschuppi81: Weird, I did exactly this otherwise re-indexing would not work!

src/onegov/search/mixins.py
Tschuppi81: Weird, I did exactly that, re-indexing would not work!

src/onegov/search/mixins.py
```diff
@@ -134,7 +152,129 @@ def es_tags(self):
...
+
+        :param cols: column names to be indexed
+        :return: tsvector string for multiple columns i.e. for columns title
+        and body: coalesce(\"title\", '') || ' ' || coalesce(\"body\", '')
+        Note that the '\"{}\"' escapes the column name in case it is a
+        keyword, see
+        http://www.postgresql.org/docs/current/static/sql-keywords-appendix.html
+        """
+
+        base = "'func.coalesce({}, '')'"
```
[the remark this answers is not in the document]
Daverball: Yes, that's still using string interpolation, the only difference is, that sqlalchemy will resolve coalesce to the generic function which can be replaced by driver specific version but the result is the same. You are still using the `||` operator instead of the generic concat function.

sqlalchemy will do its best to accept a string as a SQL expression and will try to substitute some symbols by name, but it adds a lot of overhead for one, and is very error prone and potentially dangerous in terms of properly escaping inputs (SQL Injection). Your API incorrectly assumes that it should it should build a tsvector string, but what it actually should build is a tsvector expression. I know there are some other places in OneGov where we use raw string expressions (particularly for JSONB access), but everything can be done with just function calls.

src/onegov/search/mixins.py
```diff
@@ -134,7 +152,129 @@ def es_tags(self):
...
+        ext = " || ' ' || 'func.coalesce({}, '')'"
+
+        exp = base
+        for _ in range(len(cols) - 1):
+            exp += ext
+
+        return exp.format(*cols)
```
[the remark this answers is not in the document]
Tschuppi81: Ach so, jetzt verstehe ich was du meintest!! Danke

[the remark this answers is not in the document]
Tschuppi81: After switching everything to statements using func.concat and func.coalesce I ran into 'generation expression is not immutable' error. Searching for this issue brought this page https://stackoverflow.com/questions/60249404/why-am-i-getting-a-an-error-when-creating-a-generated-column-in-postgresql saying concat is not immutable. On the following page
https://stackoverflow.com/questions/12310986/combine-two-columns-and-add-into-one-new-column/12320369#12320369
I found I should use the concat operator ' || ' or writing an own immuatable concat function. Would you also prefer the ' || ' operator?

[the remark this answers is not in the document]
Daverball: Alright I think you can force SQLAlchemy to use `||` (which is only valid for strings) by using the `concat` method on the expr instead, since the expression should know its value is a string, otherwise you may need to insert a cast to string.
```python
expr = expr.concat(' ').concat(func.coalesce(col, ''))
```
I would recommend searching in the SQLAlchemy docs rather than on stackoverflow, they're quite good.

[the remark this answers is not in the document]
Daverball: Otherwise there should also be a `op.concat` somewhere

src/onegov/search/mixins.py
```diff
@@ -161,52 +163,46 @@ def search_score(self):
...
-        :return: tsvector string
+        :return: tsvector expression
         """
-
-        columns = [p for p in model.es_properties if p in model.__dict__
-                   and not p.startswith('es_')]
-        return Searchable.create_tsvector_string(*columns) if columns else ''
+        columns = [p for p in model.es_properties if
+                   p in model.__dict__ and not p.startswith('es_')]
+        objects = [getattr(model, col) for col in columns]
```
Daverball, line 176:
```suggestion
        objects = [getattr(model, p) for p in model.es_properties]
```
if `es_properties` contains attributes that don't exist on the model that should be an error.

Also as a side note, you are currently ignoring the type of the `es_properties`, even though some of them can be represented in tsvector expressions (e.g. `translated_text` vs `text`). There may also be a way to treat `keywords` correctly. On that note `keywords` might be a JSON array, so you would need to special case it anyways.

src/onegov/search/integration.py
```diff
@@ -394,8 +398,24 @@ def reindex_model(model):
...
+        done = []
+
+        for model in searchable_sqlalchemy_models(Base):
+            if model.__tablename__ not in done:
+                model.reindex(request, model)
+                done.append(model.__tablename__)
+
+    def psql_search(self, query=''):
+        from onegov.org.models import SearchPostgres
+        return SearchPostgres(self.request(), query).generic_search()
```
Daverball, line 415: Please get rid of this dependency on `onegov.org` from `onegov.search`.

`onegov.search` is a more generic module that should work with or without `onegov.org`. If you want this method to be accessible through `psql_search` then you should either move `SearchPostgres` to `onegov.search` or extract the functionality you want into `onegov.search` and use it in both places.

`SearchPostgres` does not appear to contain any dependency on `onegov.org` itself, so there's no good reason to put it there anyways, besides the original `Search` pagination also being there.

Tschuppi81: removed

src/onegov/search/mixins.py
Tschuppi81: if `es_property` from type `keyword` a property it works well, see directory_entry

Tschuppi81: done

Daverball (commented):
This will probably break a lot of things (anything that uses `directory_id` in a query) and also does not do what you want it to do. The reason this previously failed for search indexing is simply because `to_tsvector` is not implemented for `UUID` so it needs to first be converted to `TEXT`, however, I'm not sure if the format is the same between Postgres and Python (i.e. if there will be dashes or not in the resulting string if you do a cast).

Type coercion is going to be tricky in general and doing it automatically might not be possible, although you may be able to do something with `CASE` and `pg_typeof`, instead of the the `coalesce` you've been doing to handle `NULL`.

`to_tsvector` definitely works for strings and JSON, although in the case of JSON i'm not sure what will end up in the search index, if it's just the values or also the keys. I don't remember if it works for array and hash, but you will need to figure out a way to make all the various cases work correctly.

Also removing all those `hybrid_property` is also not what you want to do to make this work, but you will need to figure out a way to generate an equivalent SQL expression for what the hybrid_property returns, so it can end up in the search index. In some cases this might not be possible and we may need a computed column to calculate the value in-app and store this value in the database.

Tschuppi81:
@Daverball In the above [commit](https://github.com/OneGov/onegov-cloud/pull/918/commits/e6fcdb16bf373b01038fe613ec355568649af58c) I am fighting with more complex hybrid expressions. Most of them gather data from other tables and I could not figure out how to properly join the corresponding tables. Currently the 'reindexing' fails in `CourseEvent`.
Note: This commit to be reverted - just shows the join problem...

Daverball:
@Tschuppi81 This is going to be tricky in general and probably a bad idea in terms of performance anyways, you are going to be better off adding caches for the joined attributes and using `@observes` methods to keep the caches up to date.

Or we could give up on calculating the search index online, since we're still quite a ways off and just generate a dictionary offline of all the search values which then can be turned into a `TSVECTOR` directly if interpreted as `JSONB`. While somewhat inefficient, his would still be quicker than what we had before elasticsearch, since we save ourselves from having to talk with the elasticsearch server. It would also get rid of the issues related to polymorphism, since SQLAlchemy is aware of what polymorphic identity we're talking about, while in postgres we would have to write CASE statements to distinguish the various versions in the same table.

PR #922, opened 2023-07-17
Pro 1092 invoice items

BreathingFlesh:

## Commit message

Feriennet: Fix invoice item bug on attendee name change

Check for attendee id instead of name in the invoice items after a new booking. Also rename the invoice item group in the current period if the attendee name changes.

TYPE: Bugfix
LINK: PRO-1092

## Checklist


src/onegov/activity/upgrade.py
```diff
@@ -888,3 +888,13 @@ def add_invoice_item_organizer(context):
...
+
+
+@upgrade_task('Add attendee id to invoice item and rm')
+def add_attendee_id_to_invoice_item_and_rm(context):
+    if not context.has_column('invoice_items', 'attendee_id'):
+        context.operations.add_column('invoice_items', Column(
+            'attendee_id', UUID, ForeignKey("attendees.id"), nullable=True
+        ))
+
+    context.operations.drop_column('invoice_items', 'period_id')
```
Daverball, line 900: You should probably go one step further and initialize `attendee_id` based on the current value of `group`, i.e. do the old annoying query where you have to join `Attendee` and `InvoiceItem` based on the `Attendee.username`, `Invoice.user_id`, `User.id` and the `Attendee.name`.

If you get more than one attendee for the same InvoiceItem somewhere, you can just default to whichever one is first.

src/onegov/feriennet/collections/billing.py
```diff
@@ -291,11 +292,11 @@ def __init__(self, session, period):
...
             WHERE "group" != 'manual'
         """)
-        self.billed_attendees = {
-            (r.username, r.group) for r in session.execute(
+        self.billed_attendees = [
+            r.attendee_id for r in session.execute(
                 select(stmt.c).where(stmt.c.period_id == period.id)
             )
-        }
+        ]
```
Daverball, line 299: Why did you switch from a set comprehension to a list comprehension? Did you misread it as a dict comprehension?

A set lookup should be faster, plus it gets rid of duplicate entries.

src/onegov/activity/collections/invoice.py
```diff
@@ -57,6 +57,13 @@ def for_schema(self, schema, schema_config=None):
         return self.__class__(self.session, self.period_id, self.user_id,
                               schema, schema_config)
 
+    def update_attendee_name(self, attendee_id, attendee_name):
+        invoice_items = self.query_items()
+
+        for item in invoice_items:
+            if item.attendee_id == attendee_id:
+                item.group = attendee_name
+
```
Daverball, line 66: This is a little fragile, since we have to remember to use this whenever we change an `attendee_name`. I'd personally prefer adding a trigger/rule on `attendees.name`. But I don't think SQLAlchemy has good support yet for triggers that don't only directly affect a foreign key, so I guess it's fine for now, but maybe add a TODO comment to `Attendee.name` that we would like to trigger automatic updates to any linked `InvoiceItem`s `group` column.

Daverball (changes requested):

src/onegov/activity/upgrade.py
Daverball: The upgrade step should still populate `InvoiceItem.attendee_id` based on the old way of joining the tables. Otherwise none of the existing InvoiceItems will be linked properly and a new InvoiceItem will be created every time a new course is booked.

PR #929, opened 2023-07-24
Api: Authentication to bypass rate limits.

cyrillkuettel:
## Commit message

Api: Authentication to bypass rate limits.

Adds token-based authentication to the api. For now,
 this is only used for rate limits. 

TYPE: Feature
LINK: OGC-1102

## Testing it

1. Go to `http://127.0.0.1:8080/onegov_agency/zg/api-keys` generate an API access key. 
2.  To request a token, make a GET to `authenticate` and provide the API key via HTTP basic authentication  in the username part (the password part is not used):
3. The token must be provided with all requests using the the username part of the HTTP basic authentication (the password part is not used). 

### curl example:

```bash
#/bin/bash

JSON=$(curl -u <your api access key from UI>: \
    --silent http://127.0.0.1:8080/onegov_agency/zg/api/authenticate)

TOKEN=$(echo $JSON | sed "s/{.*\"token\":\"\([^\"]*\).*}/\1/g")

# make a api request with the token to any endpoint, using `agencies` as an example. 
curl -X GET \
   -u $(echo -n "$TOKEN:") \
   http://127.0.0.1:8080/onegov_agency/zg/api/agencies
```

## Checklist

- [X] I have performed a self-review of my code
- [x] Write docs


src/onegov/org/views/settings.py
```diff
@@ -1,28 +1,38 @@
 """ The settings view, defining things like the logo or color of the org. """
 
 from copy import copy
+
+from aiohttp.web_exceptions import HTTPForbidden
```
Daverball, line 5: I think you probably wanted the WebOb version of the exception.

src/onegov/org/views/settings.py
```diff
@@ -302,3 +312,78 @@ def truncate(text):
...
+        request.session.add(key)
+        request.session.flush()
+        request.success(_("Your changes were saved"))
+
+    layout = layout or SettingsLayout(self, request)
+
+    def current_api_keys_by_user(
+        request: CoreRequest, self: Organisation, user: User, layout
+    ):
+        # TODO: how to make mypy happy for `api_keys` backref?
```
Daverball, line 347: You can forward declare the backref on the user Model inside a `if TYPE_CHECKING` block i.e.
```python
if TYPE_CHECKING:
    api_keys: relationship[list[ApiKey]]
```
It's possible that `test_hierarchy` will complain if you do that, since an import statement will create a soft dependency even while inside an `if TYPE_CHECKING:` block. You might be able to get around it by using `importlib` instead. Alternatively you can punt and annotate it as `relationship[list[Any]]`.

src/onegov/api/models.py
```diff
@@ -206,3 +221,47 @@ def collection(self):
...
+
+    id = Column(
+        UUID,
+        nullable=False,
+        primary_key=True,
+        default=uuid4
+    )
+
+    # the user that created the api key
+    user_id = Column(UUID, ForeignKey('users.id'))
```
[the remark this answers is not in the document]
cyrillkuettel: Ah right, it's Nullable by default.

src/onegov/api/models.py
```diff
@@ -206,3 +221,47 @@ def collection(self):
...
+
+    def __init__(self, app):
+        self.app = app
+
+
+class ApiKey(Base):
+
+    __tablename__ = 'api_keys'
+
+    id = Column(
```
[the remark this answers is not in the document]
cyrillkuettel: I removed them because I got a mypy error I was not able to solve / understand: (Maybe you know it)

`src/onegov/api/models.py:245: error: Argument 1 to "Column" has incompatible type "type[onegov.core.orm.types.uuid_type.UUID]"; expected "type[TypeEngine[uuid.UUID]]"  [arg-type]`

this happens when I add

```python
    id: 'Column[uuid.UUID]' = Column(
        UUID,
        nullable=False,
        primary_key=True,
        default=uuid4
    )

    # the user that created the api key
    user_id: 'Column[uuid.UUID]' = Column(
        UUID, ForeignKey('users.id'), nullable=False
    ```

File: src/onegov/api/views.py [hunk 5]
```diff
@@ -1,13 +1,41 @@
...
-from onegov.api.models import ApiEndpoint
+from onegov.api.models import ApiEndpoint, ApiException, AuthEndpoint, ApiKey
 from onegov.api.models import ApiEndpointCollection
 from onegov.api.models import ApiEndpointItem
-from onegov.api.models import ApiException
+from onegov.api.token import get_token, jwt_decode
 from onegov.core.security import Public
 
 
+def _authenticate(request):
```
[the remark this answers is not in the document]
cyrillkuettel: Thanks for the suggestions I will do that. Yes it was just extracted in a function... for no particular semantic reason.

src/onegov/api/models.py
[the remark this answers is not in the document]
Daverball: Yes, it's because the two types don't map cleanly to each other in the sqlalchemy plugin, you can just ignore the type error for now, there's another UUID column further below which has the appropriate `type:ignore` comment.

In SQLAlchemy 2.0 we will be able to clean this up.

cyrillkuettel:
What I wanted to mention: This function `check_rate_limit` is duplicated, for no particular reason. This one has no usages at all. I'm not sure why it's there. 

https://github.com/OneGov/onegov-cloud/blob/b8abad86c33947af7f43b08a33a2f6f628f6a33e/src/onegov/api/utils.py#L6-L50

src/onegov/api/token.py
```diff
@@ -1,5 +1,6 @@
 from base64 import b64decode
-from datetime import timedelta
+from datetime import timedelta, timezone
+import datetime
```
Daverball, line 3: why not just add datetime to the imports above?

src/onegov/api/token.py
```diff
@@ -13,33 +14,34 @@ def jwt_decode(request: CoreRequest, token: str | bytes) -> Any:
...
         'id': str(api_key.id),
     }
     return {'token': jwt_encode(request, payload)}
+
+
+def try_get_encoded_token(request: CoreRequest) -> str:
+    assert request.authorization[0].lower() == 'basic'  # type: ignore[index]
+    auth = b64decode(
+        request.authorization[1].strip()  # type: ignore[index, union-attr]
+    ).decode('utf-8')
```
Daverball, line 45:
```suggestion
    assert request.authorization is not None
    assert request.authorization.authtype == 'Basic'
    assert isinstance(request.authorization.params, str)
    auth = b64decode(
        request.authorization.params.strip()
    ).decode('utf-8')
```
You can use assertions to perform type narrowing, so you don't need to use type ignores here. Also the authorization is a `NamedTuple`, so you can access the two parts by name.

But this helped me notice that the parsing for the authorization header will only create a dict if `authtype` != `Basic`, so I might be able to replace this with a tagged union so `assert request.authorization.authtype == 'Basic'` will cause the the second item in the tuple to narrow to a `str`. But it might not work until there is `Not` type.

src/onegov/api/views.py
```diff
@@ -10,24 +9,24 @@
...
-    except NoResultFound as no_res:
-        raise HTTPClientError() from no_res
     except Exception as e:
         raise ApiException() from e
 
+    try:
+        request.browser_session.get(ApiKey, data['id'])
+    except NoResultFound as no_res:
+        raise HTTPClientError() from no_res
+
```
Daverball, line 29:
```suggestion
    if request.session.query(ApiKey).get(data['id']) is None:
        raise HTTPClientError() from no_res

```
This is my bad `session.get` wasn't added to SQLAlchemy until 1.4, I forgot we were still on 1.3. So you will still need to use the query, but `get` on query is still faster than a filter. Also `get` will not raise an exception, but that's what we want

Daverball (changes requested):

Daverball:
> What I wanted to mention: This function `check_rate_limit` is duplicated, for no particular reason.
> 
> This one is used everywhere:
> 
> https://github.com/OneGov/onegov-cloud/blob/2bc070d236f6cc03f2d239b6b9677561095b4c64/src/onegov/api/views.py#L11-L50
> 
> The other one is here. No usages. I'm not sure why it's there. I'll remove it.
> 
> https://github.com/OneGov/onegov-cloud/blob/b8abad86c33947af7f43b08a33a2f6f628f6a33e/src/onegov/api/utils.py#L6-L50

I assume someone wanted to move it from views to utils and forgot to delete it from views. Should be fine to delete it from views as long as you fix up all the imports.

src/onegov/api/views.py
cyrillkuettel: Ah, I was actually curious what you mean by this. It didn't work, so I had to use`browser_session`, but that makes sense, I'll commit this.

src/onegov/api/token.py
cyrillkuettel: Yes. Done

src/onegov/api/token.py
cyrillkuettel: Learned about type narrowing today!  Thanks.

Daverball: Another useful tool when messing around with types is the `reveal_type` function, it will cause the type checker to emit a warning with the current type of the variable, so you can observe how the type of a variable changes after you use an assertion to knock out some of the possibilities in a type union.

cyrillkuettel: So this is the type that the type checker _believes a variable to be_ at a particular point? That's quite useful.

Daverball: yes

cyrillkuettel:
@Daverball Can you take a quick lock again, I think I've got everything covered.

src/onegov/api/utils.py
```diff
@@ -1,6 +1,22 @@
+from webob.exc import HTTPUnauthorized, HTTPClientError
+from onegov.api.models import ApiException, ApiKey
+from onegov.api.token import try_get_encoded_token, jwt_decode
+import jwt
```
Daverball, line 4: Those imports have gotten a bit messy

src/onegov/org/locale/de_CH/LC_MESSAGES/onegov.org.po
```diff
@@ -5031,6 +5059,18 @@ msgid ""
...
 "Migriert Links von der angegebenen Domain zur aktuellen Domain \"${domain}\"."
 
+msgid "OneGov API"
+msgstr "OneGov API"
+
+msgid "Do you really want to delete this API key?"
+msgstr "Möchten sie diesen API Key wirklich löschen?"
+
+msgid "This action cannot be undone."
+msgstr "Diese Aktion kann nicht rückgängig gemacht werdenl"
```
Daverball, line 5069: Small typo `werdenl`

Daverball (commented):
Looks good, just some minor nitpicks

pyproject.toml
```diff
@@ -41,6 +41,8 @@ module = [
     "onegov.server.*",
     "onegov.ticket.*",
     "onegov.user.*"
+#    "onegov.api.token.*",
+#    "onegov.api.models.*"
```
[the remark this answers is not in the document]
cyrillkuettel: Ok I will make a seperate pull request, then adding all in that one.

Daverball (approved):

PR #937, opened 2023-08-02
Org: Fix linking in contact side panel 

cyrillkuettel:
## Commit message

Org: Fix linking in contact side panel 

<Optional Description>

TYPE: Bugfix
LINK:  OGC-1215

## Checklist

### DIFF
--- a/src/onegov/core/utils.py
+++ b/src/onegov/core/utils.py
@@ -1,5 +1,7 @@
 import base64
 import bleach
+from bleach import linkifier
+from bleach.linkifier import TLDS
 import errno
 import fcntl
 import gzip
@@ -416,17 +418,27 @@ def linkify(text: str, escape: bool = True) -> str:
     if not text:
         return text
 
-    linkified = linkify_phone(bleach.linkify(text, parse_email=True))
+    extended_tlds_list: list[str] = ['agency'] + TLDS
+    extended_tlds_list.remove('ag')
+    assert 'ag' not in extended_tlds_list
+
+    linkifier.URL_RE = linkifier.build_url_re(tlds=extended_tlds_list)
+    linkifier.EMAIL_RE = linkifier.build_email_re(tlds=extended_tlds_list)
+
+    bleached = bleach.linkify(text, parse_email=True)
+
+    linkified = linkify_phone(bleached)
 
     if not escape:
         return linkified
 
-    return bleach.clean(
+    cleaned = bleach.clean(
         linkified,
         tags=['a'],
         attributes={'a': ['href', 'rel']},
         protocols=['http', 'https', 'mailto', 'tel']
     )
+    return cleaned
 
 
 def paragraphify(text: str) -> str:
--- a/src/onegov/core/utils.py
+++ b/src/onegov/core/utils.py
@@ -1,7 +1,5 @@
 import base64
 import bleach
-from bleach import linkifier
-from bleach.linkifier import TLDS
 import errno
 import fcntl
 import gzip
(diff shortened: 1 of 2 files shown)


src/onegov/core/utils.py
```diff
@@ -416,7 +417,16 @@ def linkify(text: str, escape: bool = True) -> str:
...
+    long_top_level_domains = ['.agency']
+
+    # bleach.linkify supports only a fairly limited amount of tlds
+    if any(domain in text for domain in long_top_level_domains):
+        if '@' in text:
+            linkified = str(Markup(f'<a href="mailto:{text}">{text}</a>'))
+        else:
+            linkified = str(Markup(f'<a href="{text}">{text}</a>'))
+    else:
+        linkified = linkify_phone(bleach.linkify(text, parse_email=True))
```
cyrillkuettel, line 430: The current TLD list at [main/bleach/linkifier.py#L11-L23](https://github.com/mozilla/bleach/blob/main/bleach/linkifier.py?rgh-link-date=2022-04-22T08%3A05%3A19Z#L11-L23) does not contain a lot of currently-existing TLDs.

We'll have to replace bleach with something (Markupsafe?) anyway (deprecated since January 2023)

src/onegov/core/utils.py
```diff
@@ -416,7 +417,16 @@ def linkify(text: str, escape: bool = True) -> str:
     if not text:
         return text
 
-    linkified = linkify_phone(bleach.linkify(text, parse_email=True))
+    long_top_level_domains = ['.agency']
+
+    # bleach.linkify supports only a fairly limited amount of tlds
+    if any(domain in text for domain in long_top_level_domains):
+        if '@' in text:
+            linkified = str(Markup(f'<a href="mailto:{text}">{text}</a>'))
```
Daverball, line 425: This is not actually safe, you are bypassing the HTML escaping by using a f-string, you need to use `.format` on the `Markup` i.e. `Markup('<a href="mailto:{text}">{text}</a>').format(text=text)`

The purpose of the `Markup` class is to mark text that already contains markup that you validated and to mix in safely with regular strings. so `Markup` + `str` is the same as `Markup + escape(str)`

Daverball (changes requested):

src/onegov/core/utils.py
cyrillkuettel: I see. Fixed now.

src/onegov/core/utils.py
```diff
@@ -416,7 +417,22 @@ def linkify(text: str, escape: bool = True) -> str:
...
         return text
 
-    linkified = linkify_phone(bleach.linkify(text, parse_email=True))
+    long_top_level_domains = ['.agency']
+
+    # bleach.linkify supports only a fairly limited amount of tlds
+    if any(domain in text for domain in long_top_level_domains):
+        if '@' in text:
+            linkified = str(
+                Markup('<a href="mailto:{text}">{' 'text}</a>').format(
```
Daverball, line 426: Looks like you accidentally inserted a `' '` here, it won't change the outcome, since string literals will implicitly concat with each other, but it definitely doesn't seem intentional.

src/onegov/core/utils.py
Daverball: `bleach` and `Markupsafe` accomplish different goals. `bleach` is still useful for sanitizing user submitted Markup, i.e. in a field where they are allowed to enter HTML, but aren't allowed to insert let's say a script tag with malicious javascript code.

Markupsafe just completely escapes Markup, so it assumes the user input was never meant to be interpreted as markup.

Daverball: It seems easy enough to implement our own linkify, we already did it for phone numbers anyways. So I don't mind stop using the bleach method for urls.

Daverball (approved):

PR #938, opened 2023-08-03
Make event price multiline capable (as shown while editing)

Tschuppi81:
Event: Event price is now multi line capable

TYPE: Bugfix
LINK: ogc-1217

## Checklist


Daverball:
We should probably be consistent and handle all the multiline fields the same (i.e. replace the `\n` with `<br>`)

Just need to be careful not to repeat the mistake from the description field

Tschuppi81:
It is actually not about the TextAreaField, it is about how the content gets displayed in the template

Daverball (approved):

PR #957, opened 2023-08-17
Api: type annotations for api module.

cyrillkuettel:
## Commit message

Api: type annotations for api module.

TYPE: Feature

## Checklist


src/onegov/core/types.py
```diff
@@ -52,3 +55,6 @@ class LaxFileDict(TypedDict):
 
     _T = TypeVar('_T')
     SequenceOrScalar = Union[Sequence[_T], _T]
+
+    class PaginatedGenericCollection(GenericCollection[_M], Pagination[_M]):
+        pass
```
cyrillkuettel, line 60: As seen in [ApisMixin](https://github.com/OneGov/onegov-cloud/blob/56e88649b900ef5eda6a8047b1f1d79ac843189b/src/onegov/agency/api.py#L35-L48) we use the following types:

 `PaginatedAgencyCollection` 
`ExtendedPersonCollection`
`PaginatedMembershipCollection`

They all have the thing in common (and this is expected from the implementation) that extend `GenericCollection` and `Pagination`, so i attempted to create `PaginatedGenericCollection` to imitates this. But I know I'm missing something here.
https://github.com/OneGov/onegov-cloud/blob/175c2b2b91528440709575c3882b04f45b826e49/src/onegov/api/models.py#L237

the last mypy error:
```bash
src/onegov/api/models.py:204: error: Argument 1 to "for_item" of "ApiEndpoint" has incompatible type "_M"; expected "ApiEndpointItem"  [arg-type]
```

src/onegov/api/models.py
```diff
@@ -217,16 +234,30 @@ def item_links(self, item):
...
 
 
+class ApiEndpointCollection:
+    """ A collection of all available API endpoints. """
+
+    def __init__(self, app: 'Framework'):
+        self.app = app
+
+    @cached_property
+    def endpoints(self) -> dict[str, Type['ApiEndpoint']]:
```
Daverball, line 251:
```suggestion
    def endpoints(self) -> dict[str, type[ApiEndpoint]]:
```

src/onegov/api/models.py
```diff
@@ -217,16 +234,30 @@ def item_links(self, item):
         raise NotImplementedError()
 
     @property
-    def collection(self):
+    def collection(self) -> 'PaginatedGenericCollection[_M]':
```
Daverball, line 237: This implies to me that `ApiEndpoint` should be a `Generic` in `_M` as well, that way you can also improve `by_item` `by_id` and `batch` and can replace `Any` with `_M`. In this case `_M` is unbound anyways, so it's only slightly better than `PaginatedGenericCollection[Any]`

src/onegov/api/path.py
```diff
@@ -18,7 +23,12 @@ def get_api_endpoints(app):
     path='/api/{endpoint}',
     converters=dict(page=int)
 )
-def get_api_endpoint(app, endpoint, page=0, extra_parameters=None):
+def get_api_endpoint(
+    app: 'Framework',
+    endpoint: str,
+    page: int = 0,
+    extra_parameters: dict[str, Any] | None = None,
+) -> 'ApiEndpoint | AuthEndpoint':
```
Daverball, line 31:
```suggestion
) -> ApiEndpoint | AuthEndpoint:
```

src/onegov/core/types.py
Daverball: Yeah, you would need `Intersection` for that, so you can't really type this properly yet, unless you define a `Protocol` that matches this `Intersection` manually.

Daverball (changes requested):
Looks good for the most part, but there are some inconsistencies that need to be addressed.

src/onegov/api/models.py
cyrillkuettel: Something like `Intersection[FooClass, BarMixin]` would be pleasant to have, but I think I'll remove this `PaginatedGenericCollection` and use a Protocol.

src/onegov/api/models.py
```diff
@@ -9,6 +9,7 @@
 from sqlalchemy.orm import relationship
 from uuid import uuid4
 from sqlalchemy import Text
+from typing_extensions import Self
```
Daverball, line 12: I would move this import to the `if TYPE_CHECKING:` block. Some of our dependencies will already depend on `typing_extensions`, so it should generally be around at runtime, but we don't directly depend on it ourselves outside of a type checking context, so we should be consistent about that and only use it in type checking context.

src/onegov/core/types.py
```diff
@@ -56,5 +68,99 @@ class LaxFileDict(TypedDict):
...
+
+        def subset(self) -> 'Query[_M]':
+            ...
+
+        @cached_property
+        def cached_subset(self) -> 'Query[_M]':
+            ...
+
+        @property
+        @abstractmethod
```
Daverball, line 121: `abstractmethod` is redundant inside a protocol, mypy will probably get confused about this when checking the protocol, so it's better to get rid of it.

Also not everyone seems to use `page` consistently, I have seen some collections where `page` is optional. So it might be better to annotate it as `page: int | None`, so it matches as many implementations as possible.

src/onegov/api/models.py
```diff
@@ -235,7 +243,7 @@ def item_links(self, item: 'ApiEndpointItem | None') -> dict[str, Any]:
         raise NotImplementedError()
 
     @property
-    def collection(self) -> Any:
+    def collection(self) -> 'PaginatedGenericCollection[Any]':
```
Daverball, line 246: Might as well use `_M` here as well and make `ApiEndpoint` generic i.e. `ApiEndpoint(Generic[_M])`. That way implementations can be specific about what model they return.

Daverball: You will need to move the import of `_M` outside the type checking block though.

src/onegov/api/models.py
```diff
@@ -107,51 +116,59 @@ class ApiEndpoint:
...
+    def for_filter(self, **filters: Any) -> 'ApiEndpoint':
         """ Return a new endpoint instance with the given filters while
         discarding the current filters and page.
 
         """
 
         return self.__class__(self.app, filters)
 
-    def for_item(self, item):
+    def for_item(self, item: 'ApiEndpointItem') -> 'ApiEndpointItem | None':
```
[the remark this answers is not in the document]
cyrillkuettel: I think you are absolutely right `ApiEndpointItem` is not correct here.

src/onegov/api/path.py
```diff
@@ -28,7 +29,8 @@ def get_api_endpoint(
     endpoint: str,
     page: int = 0,
     extra_parameters: dict[str, Any] | None = None,
-) -> ApiEndpoint | AuthEndpoint:
+) -> ApiEndpoint[_M] | AuthEndpoint | Any:
+
```
cyrillkuettel, line 33: Without `Any`, this fails with

`error: Incompatible return value type (got "ApiEndpoint[_M]", expected "ApiEndpoint[_M] | AuthEndpoint")  [return-value]`

I am not quite sure what's the issue

src/onegov/api/models.py
```diff
@@ -83,14 +81,16 @@ def __init__(self, app: 'Framework', endpoint: str, id: str):
         self.id = id
 
     @property
-    def api_endpoint(self) -> 'ApiEndpoint | None':
+    def api_endpoint(self) -> 'ApiEndpoint[_M] | None':
```
Daverball, line 84: Might as well make `ApiEndpointItem` generic in `_M` as well.

src/onegov/api/models.py
```diff
@@ -149,7 +149,7 @@ def for_filter(self, **filters: Any) -> 'Self':
 
         return self.__class__(self.app, filters)
 
-    def for_item(self, item: 'ApiEndpointItem') -> 'ApiEndpointItem | None':
+    def for_item(self, item: Any | None) -> 'ApiEndpointItem | None':
```
Daverball, line 152: This should be `_M | None`.

src/onegov/api/path.py
```diff
@@ -28,7 +29,8 @@ def get_api_endpoint(
     endpoint: str,
     page: int = 0,
     extra_parameters: dict[str, Any] | None = None,
-) -> ApiEndpoint | AuthEndpoint:
+) -> ApiEndpoint[_M] | AuthEndpoint | Any:
```
Daverball, line 32: This should be `ApiEndpoint[Any]` as well. Unbound type parameters sometimes make sense when defining callback protocols, but generally they should be avoided.

Daverball (changes requested):

src/onegov/api/path.py
Daverball: This should be `ApiEndpoint[Any]`, unbound type variables are generally to be avoided, there's some rare edge cases where they can be useful.

src/onegov/api/models.py
```diff
@@ -81,9 +81,8 @@ def __init__(self, app: 'Framework', endpoint: str, id: str):
         self.id = id
 
     @property
-    def api_endpoint(self) -> 'ApiEndpoint[_M] | None':
+    def api_endpoint(self) -> 'ApiEndpoint[Any] | None':
```
Daverball, line 84:
```suggestion
    def api_endpoint(self) -> 'ApiEndpoint[_M] | None':
```

Now that `ApiEndpointItem` is itself generic you can actually return `ApiEndpoint[_M]` here, the reason it was complaining before was that there were two different versions of `_M` involved, an unbound one and a bound one, which are incompatible.

Some other type checkers such as pyright help visualize this issue a bit better by adding a unique numeric suffix to each type variable instance, so it's easier to distinguish different instances of the same type variable in error messages.

src/onegov/api/models.py
```diff
@@ -149,7 +148,7 @@ def for_filter(self, **filters: Any) -> 'Self':
 
         return self.__class__(self.app, filters)
 
-    def for_item(self, item: Any | None) -> 'ApiEndpointItem | None':
+    def for_item(self, item: _M | None) -> 'ApiEndpointItem[Any] | None':
```
Daverball, line 151:
```suggestion
    def for_item(self, item: _M | None) -> 'ApiEndpointItem[_M] | None':
```

src/onegov/api/models.py
```diff
@@ -199,7 +198,7 @@ def links(self) -> dict[str, 'Self | None']:
         return result
 
     @property
-    def batch(self) -> dict['ApiEndpointItem | None', Any]:
+    def batch(self) -> dict['ApiEndpointItem[Any] | None', Any]:
```
Daverball, line 201:
```suggestion
    def batch(self) -> dict['ApiEndpointItem[_M]', _M]:
```

src/onegov/api/models.py
cyrillkuettel: Ah _that_ was the reason. I was getting confused because of this nonsensical error message, something like expected type `_M`, got type `_M`

src/onegov/api/models.py
```diff
@@ -177,18 +198,17 @@ def links(self):
         return result
 
     @property
-    def batch(self):
+    def batch(self) -> dict['ApiEndpointItem[_M] | None', _M]:
```
Daverball, line 201:
```suggestion
    def batch(self) -> dict['ApiEndpointItem[_M]', _M]:
```
I think this is the correct type since `collection.batch` should have no `None` values. mypy probably complains because `self.for_item` has no overloads and thus it cannot infer that the return value should be `ApiEndpointItem[_M]` as long as `item` is not `None`

src/onegov/api/models.py
```diff
@@ -107,67 +117,78 @@ class ApiEndpoint:
...
+    def for_filter(self, **filters: Any) -> 'Self':
         """ Return a new endpoint instance with the given filters while
         discarding the current filters and page.
 
         """
 
         return self.__class__(self.app, filters)
 
-    def for_item(self, item):
+    def for_item(self, item: _M | None) -> 'ApiEndpointItem[_M] | None':
```
Daverball, line 151:
```suggestion
    @overload
    def for_item(self, None) ->  None: ...
    @overload
    def for_item(self, item: _M) -> 'ApiEndpointItem[_M]': ...
   
    def for_item(self, item: _M | None) -> 'ApiEndpointItem[_M] | None':
```
This is how you let mypy infer the return type a bit better from the passed in value

src/onegov/api/models.py
```diff
@@ -12,31 +12,46 @@
 from onegov.core.orm import Base
 from onegov.core.orm.types import UUID, UTCDateTime
 from onegov.user import User
+from onegov.core.collection import _M
 
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Generic
```
Daverball, line 18:
```suggestion
from typing import TYPE_CHECKING, Any, Generic, overload
```
Necessary import to make the other suggestion work

Daverball (approved):

cyrillkuettel:
Thanks for the review!  @Daverball

PR #975, opened 2023-08-29
Org: Increase robustness in `get_links` und `get_summary`.

cyrillkuettel:
## Commit message

Org: Increase robustness in `get_links` und `get_summary`.

<Optional Description>

TYPE: Bufix
LINK: OGC-1236
HINT: <Optional upgrade hint>

## Checklist

### DIFF
--- a/src/onegov/org/models/ticket.py
+++ b/src/onegov/org/models/ticket.py
@@ -218,11 +218,12 @@ class FormSubmissionHandler(Handler, TicketDeletionMixin):
 
     def get_summary(self, request):
         layout = DefaultLayout(self.submission, request)
-
-        return render_macro(layout.macros['display_form'], request, {
-            'form': self.form,
-            'layout': layout
-        })
+        if self.submission:
+            return render_macro(layout.macros['display_form'], request, {
+                'form': self.form,
+                'layout': layout
+            })
+        return None
 
     def get_links(self, request):
         layout = DefaultLayout(self.submission, request)
--- a/src/onegov/org/models/ticket.py
+++ b/src/onegov/org/models/ticket.py
@@ -231,7 +231,8 @@ class FormSubmissionHandler(Handler, TicketDeletionMixin):
         extra = []
 
         # there's a decision to be made about the registration
-        window = self.submission.registration_window
+        window = (self.submission.registration_window
+                  if self.submission else None)
 
         if window:
             if self.submission.spots and self.submission.claimed is None:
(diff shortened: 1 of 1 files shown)


src/onegov/org/models/ticket.py
```diff
@@ -154,19 +154,19 @@ def deleted(self):
 
     @property
     def email(self):
-        return self.submission.email
+        return self.submission.email if self.submission else ''
 
     @property
     def title(self):
-        return self.submission.title
+        return self.submission.title if self.submission else ''
```
Daverball, line 161:
```suggestion
        return self.submission.title if self.submission is not None else ''
```

src/onegov/org/models/ticket.py
```diff
@@ -218,11 +218,12 @@ def ticket_deletable(self):
 
     def get_summary(self, request):
         layout = DefaultLayout(self.submission, request)
-
-        return render_macro(layout.macros['display_form'], request, {
-            'form': self.form,
-            'layout': layout
-        })
+        if self.submission:
```
Daverball, line 221:
```suggestion
        if self.submission is not None:
```

src/onegov/org/models/ticket.py
```diff
@@ -323,16 +325,17 @@ def get_links(self, request):
                 )
             )
 
-        edit_link = URL(request.link(self.submission))
-        edit_link = edit_link.query_param('edit', '').as_string()
+        if self.submission:
```
Daverball, line 328:
```suggestion
        if self.submission is not None:
```

src/onegov/org/models/ticket.py
```diff
@@ -830,7 +834,7 @@ def submitter_address(self):
 
     @property
     def title(self):
-        return self.submission.title
+        return self.submission.title if self.submission else ''
```
Daverball, line 837:
```suggestion
        return self.submission.title if self.submission is not None else ''
```

Daverball (changes requested):
Just a small nitpick

cyrillkuettel:
> Just a small nitpick

Makes sense I didn't actually know that.

Daverball:
> > Just a small nitpick
> 
> Makes sense I didn't actually know that.

FYI on the `Files changed` view you can add individual suggestions to a batch add commit them all at once.

Also looks like one of the suggestions pushed one of the lines just barely past the column limit, so that's the only thing left to clean up.

Daverball (approved):

PR #990, opened 2023-09-05
Feature/ogc 1219 winti filter for event calendar

Tschuppi81:
Events: Configurable event filters for Winterthur (analog to directories)

TYPE: Feature
LINK: ogc-1219
HINT: `onegov-core --select /onegov_winterthur/* upgrade` or `onegov-core upgrade

## Checklist


src/onegov/org/layout.py
```diff
@@ -1719,19 +1720,28 @@ def breadcrumbs(self):
...
-        if self.request.is_manager:
-            return [
-                Link(
+        def links():
+            if self.request.is_admin and isinstance(self.app, WinterthurApp):
+                yield Link(
+                    text=_("Configure"),
+                    url=self.request.link(self.model, '+edit'),
+                    attrs={'class': 'edit-link'}
+                )
```
Daverball, line 1729: Why are you restricting this to `WintherthurApp`? It doesn't make any sense to me.

If you really wanted to do that, then this is the wrong place anyways, you would need to overwrite the layout in `onegov.winthethur` and also override all the views that use it.

src/onegov/org/templates/occurrence.pt
```diff
@@ -14,10 +14,15 @@
                     <div tal:condition="occurrence.event.location">${occurrence.event.location}</div>
                 </div>
 
-                <div class="occurrence-tags">
+                <div class="occurrence-tags" tal:condition="occurrence.tags and not winti">
```
Daverball, line 17: Same here, don't have `onegov.winterthur` specific code in `onegov.org`

src/onegov/org/views/occurrence.py
```diff
@@ -26,13 +89,20 @@ def view_occurrences(self, request, layout=None):
...
-            url=request.link(self.for_filter(tag=tag)),
-            active=tag in self.tags and 'active' or ''
-        ) for tag, translation in translated_tags
-    ]
+    filters = None
+    tags = None
+
+    if isinstance(request.app, WinterthurApp) and self.event_config:
+        keyword_counts = keyword_count(request, self)
+        filters = get_filters(request, self, keyword_counts)
```
Daverball, line 97: same here

src/onegov/org/views/event.py
```diff
@@ -62,6 +66,13 @@ def event_form(model, request, form=None):
         # unlike typical extended models, the property of this is defined
         # on the event model, while we are only using the form extension part
         # here
+        if isinstance(request.app, WinterthurApp):
```
Daverball, line 69: same here

Daverball (changes requested):
There are some architectural issues with this PR, I realize that using `onegov.directory` as the reference was probably a bad move and sent you off in a bad direction, since it's one of the older modules and does some things in ways we shouldn't really do anymore. I probably should've emphasized looking at `onegov.reservation` which is a little bit more modern.

`onegov.event` is even older and has no polymorphism, so it's more difficult to extend in `onegov.org`, nevertheless we should avoid adding new tables just to store three configuration values, when there is already a good place to store them, even if it means moving some of the features out of the `onegov.event` module.

There's also the issue of back referencing `onegov.winterthur` from `onegov.org` which is code smell and also doesn't seem necessary to me, we don't need to restrict this feature to Winterthur, everyone is allowed to benefit.

src/onegov/org/forms/event.py
```diff
@@ -266,12 +270,14 @@ def populate_submitter(self):
             self.email.data = self.request.current_username
 
     def on_request(self):
-        if self.custom_tags():
-            self.tags.choices = [(tags, tags) for tags in self.custom_tags()]
+        if self.tags:
+            if self.custom_tags() and self.tags:
```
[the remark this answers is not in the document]
Tschuppi81: I remove the tags from the form to use filters instead in event.py line +75. Do you see a better way of doing so?
In my opinion we should have tags or filters, not both. Or do you see a reason for it?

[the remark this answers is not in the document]
Tschuppi81: But you are right, I don't need to check it a second time

src/onegov/org/layout.py
Tschuppi81: As mentioned above, I think we should use either tags or filters. How should one decide? By setting the filter config? Or would you introduce a setting?

src/onegov/org/forms/event.py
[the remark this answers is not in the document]
Daverball: Since tags can be implemented using the new keywords.

src/onegov/org/layout.py
Daverball: If you go the settings route, the setting should probably be `tags`, `filters` and `tags_and_filters` with the default being `tags`.

Daverball: The filter specific configuration fields can then only be displayed when either of the former two values has been selected.

src/onegov/event/upgrade.py
```diff
@@ -55,3 +57,12 @@ def add_filter_keywords_column(context):
...
                     HSTORE), nullable=True))
+
+
+@upgrade_task('Adds default value for event filter type')
+def add_default_setting_for_event_filter_type(context):
+    org = context.session.query(Organisation).first()
+
+    if org and 'event_filter_type' not in org.meta:
+        org.meta['event_filter_type'] = 'tags'
+
```
Daverball, line 68: I don't think this is needed since you provided a default value on the `meta_property`, itwill return its default value if it hasn't been set yet.

src/onegov/event/collections/occurrences.py
```diff
@@ -303,6 +389,24 @@ def query(self):
...
             query = query.filter(Occurrence._tags.has_any(array(self.tags)))
 
+        if self.filter_keywords:
+            keywords = self.valid_keywords(self.filter_keywords)
+
+            values = [val for sublist in keywords.values() for val in sublist]
+            values.sort()
+
+            values = [
+                Event.filter_keywords.contains(value)
```
Tschuppi81, line 399: @Daverball I could not figure out how to properly filter for the selected keywords on the UI. Any idea?
(code runs but filtering is not working)

Daverball: @Tschuppi81 That's because `dict_property` by default assume a string value, I will probably get rid of that, I had a good idea how to infer the value type from the annotation on the left hand side.

But in the meantime you can do `content_property(value_type=list)` to force it to not coerce to a string in SQL expressions, then `contains` should work just fine.

Tschuppi81: Thanks. I did set the value type to dict as I stored keywords in this form `{"filter_keywords": {"kalender": "DWS Sport", "sportart": "Fussball"}}`. I guesss I need to get the values first and then need to test...

src/onegov/event/types/event_configuration.py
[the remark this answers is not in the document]
Tschuppi81: removed

src/onegov/org/forms/event.py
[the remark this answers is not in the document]
Tschuppi81: removed second check
supporting tags, filters or both

src/onegov/event/models/mixins.py
```diff
@@ -30,6 +30,11 @@ class OccurrenceMixin:
     _tags = Column(  # type:ignore[call-overload]
         MutableDict.as_mutable(HSTORE), nullable=True, name='tags')
 
+    #: All filter keywords defined for this entry
+    _filter_keywords = Column(  # type:ignore
+        MutableDict.as_mutable(HSTORE), nullable=True, name='filter_keywords'
+    )
```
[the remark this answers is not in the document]
Tschuppi81: Done

src/onegov/org/layout.py
Tschuppi81: Limitation to winterthur removed, filters are generally available and can be configured in organisation settings

src/onegov/org/templates/occurrence.pt
Tschuppi81: done

src/onegov/org/views/occurrence.py
Tschuppi81: done

src/onegov/org/views/event.py
Tschuppi81: done

src/onegov/event/upgrade.py
Tschuppi81: ok, got you

src/onegov/event/collections/occurrences.py
```diff
@@ -211,6 +252,51 @@ def tag_counts(self) -> defaultdict[str, int]:
...
+        """
+        keywords = tuple(
+            as_internal_id(k) for k in
+            self.request.app.org.event_filter_configuration.get('keywords',
+                                                                set)
+        )
+
+        fields = {
+            f.id: f for f in
+            self.request.app.org.event_filter_fields if (f.id in keywords)
```
Tschuppi81, line 283: @Daverball Do you have a better approach to access the organization settings. This way the collection requires the request which gives me some headache in the unit tests

Daverball: I would manually pass them into the collection, rather than trying to retrieve them. This decouples the module properly from `onegov.org` anyways, since otherwise it has a soft dependency that should not be there.

This means that you will need to properly bind those parameters in the views on that collection, but I think that's fine. Alternatively you can override the `OrgApp.path` directive for this collection to automatically bind those parameters for you.

src/onegov/event/collections/occurrences.py
```diff
@@ -272,15 +277,17 @@ def available_filters(self, sort_choices=False, sortfunc=None):
         :rtype tuple(tuples(keyword, title, values as list)
 
         """
+        assert self.event_filter_configuration
+        assert self.event_filter_fields
```
Daverball, line 281: I would just return an empty dict if they're not set. I don't think we need to require them to be set.

Tschuppi81: allright

Tschuppi81:
@Daverball Please review again after my latest changes

Daverball (commented):
This looks a lot better now. There's only a few things left I would still change.

There's a couple of small bugs, which should be addressed, while the rest are small improvement suggestions.

src/onegov/event/collections/occurrences.py
```diff
@@ -223,7 +323,7 @@ def used_tags(self):
         """
         base = self.session.query(Occurrence._tags.keys()).with_entities(
             sqlalchemy.func.skeys(Occurrence._tags).label('keys'),
-            Occurrence.start)
+            Occurrence.end)
```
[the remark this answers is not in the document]
Tschuppi81: this is a change for a different PR, will remove it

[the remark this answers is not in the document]
Tschuppi81: done

src/onegov/event/models/event.py
```diff
@@ -374,6 +377,9 @@ def _update_occurrences(self):
         for start in self.occurrence_dates():
             self.occurrences.append(self.spawn_occurrence(start))
 
+        for occ in self.occurrences:
+            occ.filter_keywords = self.filter_keywords
+
```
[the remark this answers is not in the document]
Tschuppi81: done

src/onegov/org/views/occurrence.py
```diff
@@ -114,6 +199,66 @@ def view_occurrence(self, request, layout=None):
...
+
+
+@OrgApp.form(model=OccurrenceCollection, name='edit',
+             template='directory_form.pt', permission=Secret,
+             form=EventConfigurationForm)
+def handle_edit_event_filters(self, request, form, layout=None):
+    try:
+        if form.submitted(request):
+            keywords = form.keyword_fields.data
+            keywords = keywords.split('\r\n') if '\r\n' in keywords \
```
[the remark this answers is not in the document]
Tschuppi81: Nice one!!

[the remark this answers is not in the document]
Tschuppi81: done

src/onegov/org/models/organisation.py
```diff
@@ -237,3 +242,12 @@ def title_lines(self):
 
     def excluded_person_fields(self, request):
         return [] if request.is_logged_in else self.hidden_people_fields
+
+    @property
+    def event_filter_fields(self):
+        @staticmethod
+        @lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
Tschuppi81: Did a quick test between the suggested `cached_property` and the current implementation using `lru_cache`.
Using the `cached_property` is better as it only gets evaluated once per call on `/events`. I will switch it!

src/onegov/org/forms/event.py
```diff
@@ -381,6 +385,20 @@ def populate_obj(self, model):
         else:
             raise NotImplementedError
 
+        if self.request.app.org.event_filter_type in ['filters',
+                                                      'tags_and_filters']:
+            from onegov.form.parser.core import CheckboxField, RadioField
+            filter_keywords = dict()
+            for field in self.request.app.org.event_filter_fields:
+                if isinstance(field, (CheckboxField, RadioField)):
```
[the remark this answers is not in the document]
Tschuppi81: Background: User can specify whatever as event filter configuration e.g. email, text and more. For filtering we only process `Checkbox`and `Radio` no others.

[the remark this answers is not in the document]
Tschuppi81: done

src/onegov/org/models/organisation.py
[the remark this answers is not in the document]
Daverball: You will need to be a little bit careful, since the `cached_property` will not be cleared automatically when `definition` changes, so a `lru_cache` might still make sense, it just needs to be defined in an outer scope.

You could also go a slightly different route and move this property to `OrgApp` and use `orm_cached`. That way the property will remain cached on the application until the organization table gets updated. There's already a couple of properties like that on `OrgApp`, such as `root_pages`.

src/onegov/form/validators.py
```diff
@@ -302,6 +299,39 @@ def __call__(self, form: 'Form', field: 'Field') -> None:
...
+                           "allowed.")
+
+    def __call__(self, form: 'Form', field: 'Field') -> None:
+        from onegov.form.fields import MultiCheckboxField
+
+        super().__call__(form, field)
+
+        # limit the definition to MultiCheckboxField, RadioField which can
+        # be used for filter definition
+        parsed_form = self._parse_form(field, enable_indent_check=False)
```
Daverball, line 322: This will cause the form to get parsed twice, I would change `ValidFormDefinition.__call__` so it returns either the parsed form or `None` so you can reuse it in your own `__call__` method.

```python
    parsed_form = super().__call__(form, field)
    if parsed_form is None:
        return None
```

Tschuppi81: It now only gets parsed once, but the last error found in the form will be displayed twice. Don't know why as the 'errors' only contains 3 items...
Screenshot from 2023-09-12 14-40-44

src/onegov/org/forms/event.py
```diff
@@ -415,6 +433,20 @@ def process_obj(self, model):
...
+            from onegov.form.parser.core import CheckboxField, RadioField
+            keywords = model.filter_keywords
+
+            for field in self.request.app.org.event_filter_fields:
+                form_field = getattr(self, field.id, None)
+
+                if form_field is None:
+                    continue
+
+                if isinstance(field, (CheckboxField, RadioField)):
```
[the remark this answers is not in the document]
Tschuppi81: removed

src/onegov/org/views/occurrence.py
```diff
@@ -82,9 +163,12 @@ def view_occurrences(self, request, layout=None):
         'start': self.start.isoformat() if self.start else '',
         'ranges': ranges,
         'tags': tags,
+        'filters': filters,
         'locations': locations,
         'title': _('Events'),
         'search_widget': self.search_widget,
+        'show_tags': filter_type in ['tags', 'tags_and_filters'],
+        'show_filters': filter_type in ['filters', 'tags_and_filters'],
```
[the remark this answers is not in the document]
Tschuppi81: Let me know if there is a better way than ../view/utils.py

src/onegov/org/models/organisation.py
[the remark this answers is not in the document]
Tschuppi81: Good catch! I now implemented using `lru_cache` and it works!

src/onegov/form/validators.py
```diff
@@ -302,6 +299,47 @@ def __call__(self, form: 'Form', field: 'Field') -> None:
...
+                error = field.gettext(self.invalid_field_type.format(
+                    label=field.label.text))
+                errors = form['definition'].errors
+                if not isinstance(errors, list):
+                    errors = form['definition'].process_errors
+                    assert isinstance(errors, list)
+                errors.append(error)
+
+        if errors:
+            raise ValidationError(error)
```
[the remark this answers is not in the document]
Daverball: If `ValidationError` contains a message it will be added to the field's error list, but since we already manually added the error so we can report multiple errors instead of just the first one, we don't need to add a new error message, we just need to tell WTForms that the validation failed.

Tschuppi81:
Are we ready to merge?

Daverball:
> Are we ready to merge?

Looks like your merge with master broke some things. Once that is sorted out you can merge.

Tschuppi81:
Cannot reproduce issue locally...

Daverball:
Actually I see you've only made your changes for org, you will also need to make them for town6. I don't think that's the source of the issue, but most of the failing tests seem to be in town6 with views having the wrong layout.

Tschuppi81:
Totally agree on your comment but after commit 'adjust annotations' (https://github.com/OneGov/onegov-cloud/pull/990/commits/245a3df8a314c919a0ee08114418003c20fbf7c1) the tests were successful!

src/onegov/org/models/organisation.py
```diff
@@ -237,3 +242,13 @@ def title_lines(self):
...
     def excluded_person_fields(self, request):
         return [] if request.is_logged_in else self.hidden_people_fields
+
+    @property
+    def event_filter_fields(self):
+        return flatten_event_filter_fields_from_definition(
+            self.event_filter_definition)
+
+
+@lru_cache(maxsize=1)
```
Daverball, line 252: I would not set a maxsize this small, since this cache is shared between all tenants, you can do it if you move this cache to a method on the app, because then the maxsize means you can only have one of these per app and the app sticks around, otherwise you will need to go for something larger, like maybe 64/128?

Tschuppi81: Makes sense

Tschuppi81:
I can still not figure out what causes the tests to fail! Locall `pytest tests/onegov/town6` passes, no issue. On the server many fail. What is different on the server? Also @cyrillkuettel checked the commits again and could not find the root cause. Do you have an idea @Daverball?

Daverball:
> I can still not figure out what causes the tests to fail! Locall `pytest tests/onegov/town6` passes, no issue. On the server many fail. What is different on the server? Also @cyrillkuettel checked the commits again and could not find the root cause. Do you have an idea @Daverball?

I couldn't really figure it out either at a cursory glance, definitely appears to be something subtle. Have you tried running the tests locally with `pytest-split` and the same group number? It might be some subtle bug in the test suite that only rears its ugly head if the tests get split in a certain way. Since you added new tests, the tests will be split slightly differently compared to the master branch.

Tschuppi81:
Yeah, 8 splits worked!

cyrillkuettel:
@Daverball Saved the day  🎉 🎉 🎉

Daverball:
This doesn't actually invoke the tests that previously failed, because  you are only testing the first six groups in the split, if you want to test 8 splits, then you need to add groups 7 & 8 to the `matrix` configuration of the `step`.

Daverball:
That's good enough for now to merge, if you mirror that change to this branch, we just have to keep in mind that this issue may crop up again in the future. It's probably some fixture that doesn't properly get initialized/reset if you skip over one of the tests, creating a weird inter-dependency between multiple tests.

Tschuppi81:
I totally agree, it will pop-up again some when in the future. I hope not to soon ;-)

Tschuppi81:
Thank you for your help, guys!

PR #1000, opened 2023-09-14
Handle raised roadwork errors

Tschuppi81:
winterthur: handle roadwork internal client connection errors

TYPE: Bugfix
LINK: ogc-1260


Tschuppi81:
@freinhard did you have a change to review?

treinhard (commented):
I think it's better to let the exception uncaught instead of silently ignoring it so that we have something in the logs.

PR #1007, opened 2023-09-18
Town6: Sort by pages in person card.

cyrillkuettel:
## Commit message

Town6: Sort by pages in person card.

<Optional Description>

TYPE: Bugfix
LINK: OGC-1257

## Checklist

### DIFF
--- a/src/onegov/org/views/people.py
+++ b/src/onegov/org/views/people.py
@@ -1,7 +1,6 @@
 from collections import namedtuple
-
 import morepath
-
+import time
 from morepath.request import Response
 from onegov.core.security import Public, Private
 from onegov.org import _, OrgApp
@@ -11,6 +10,12 @@ from onegov.org.layout import PersonLayout, PersonCollectionLayout
 from onegov.org.models import AtoZ, Topic
 from onegov.people import Person, PersonCollection
 from markupsafe import Markup
+
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from onegov.page import Page
+    from collections.abc import Iterable, Iterator
 
 
 @OrgApp.html(model=PersonCollection, template='people.pt', permission=Public)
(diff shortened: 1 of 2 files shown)


src/onegov/org/views/people.py
```diff
@@ -34,10 +40,21 @@ def get_items(self):
...
+    def visit_topics_with_people(
+        pages: 'Iterable[Page]',
+        root_id: int | None = None
+    ) -> 'Iterator[Topic]':
+        for page in pages:
+
+            if root_id is None:
+                root_id = page.id
 
+            if isinstance(page, Topic) and page.people:
```
Daverball, line 52: Since `page.people` does a bunch of work that has to be repeated every time the property is accessed it might be better to check for `page.content.get('people')` here. Since `person_functions_by_organization` accesses this attribute again.

src/onegov/org/views/people.py
```diff
@@ -46,36 +63,46 @@ def view_person(self, request, layout=None):
...
+        ),
+        key=lambda pair: pair.topic.title,
+    )
+
+    return (
+        Markup("<span>{0}: {1}</span>".format(
+            Markup('<a href="{0}">{1}</a>').format(request.link(pair.topic),
+                                                   pair.topic.title),
+            pair.function
+        ))
```
Daverball, line 103:
```suggestion
        Markup('<span><a href="{url}">{title}</a>: {function}</span>'.format(
            url=request.link(pair.topic),
            title=pair.topic.title,
            function=pair.function
        ))
```

src/onegov/org/views/people.py
```diff
@@ -46,36 +63,46 @@ def view_person(self, request, layout=None):
...
-                        func = person.context_specific_function
-                        if func:
-                            page = f"<a href=\"{request.link(topic)}\">" \
-                                   f"{topic.title}</a>"
-                            organization_to_function.append(
-                                Markup(f"<span>{page}: {func}</span>"))
-                except AttributeError:
-                    continue
-    return organization_to_function
+    TopicFunctionPair = namedtuple("TopicFunctionPair", ["function", "topic"])
```
Daverball, line 80: Might as well use `typing.NamedTuple`. That way we don't have to change it once we add type annotations.

Daverball:
```suggestion
    class TopicFunctionPair(NamedTuple):
        function: str
        topic: Topic
```

Daverball: You will also need to change the import

Daverball (commented):
Looks good, I just have a couple of nitpicks.

src/onegov/org/views/people.py
cyrillkuettel: ah yes, we can save ourselves this call.

PR #1009, opened 2023-09-19
Org: Use cached topics for more speed.

cyrillkuettel:
## Commit message

Org: Use cached topics for more speed.

In most cases, `session.query(Topic).all()` is redundant because all pages 
are pre-loaded with `@orm_cached` anyway in the `OrgApp`.

TYPE: Bugfix

## Checklist


src/onegov/org/app.py
```diff
@@ -193,23 +193,25 @@ def ticket_permissions(self):
...
         ) -> 'Iterator[tuple[int, Topic]]':
             for page in pages:
+
                 if isinstance(page, Topic):
                     yield root_id or page.id, page
 
                 yield from visit_topics(
                     page.children,
-                    root_id=root_id or page.id
+                    root_id=root_id
```
Daverball, line 207: This will not work correctly, since the widget for homepage_pages expects all the nested topics to exist in one combined flat list per root page.

The `or page.id` is necessary to forward the root_id to the children tree. So either leave it like it was or use a new variable `current_root_id = root_id or page.id` and use it for the `yield` above and down here, if you want to avoid the duplicated code.

src/onegov/org/app.py
```diff
@@ -193,23 +193,25 @@ def ticket_permissions(self):
         return result
 
     @orm_cached(policy='on-table-change:pages')
-    def homepage_pages(self) -> dict[int, list[Topic]]:
-
+    def topics_cached(self) -> list[tuple[int, 'Topic']]:
         def visit_topics(
-            pages: 'Iterable[Page]',
-            root_id: int | None = None
+            pages: 'Iterable[Page]', root_id: int | None = None
```
Daverball, line 198: Just a small style nit, but when the whole signature doesn't fit on one line, I would always prefer if each argument had its own line.

src/onegov/org/app.py
```diff
@@ -193,23 +193,25 @@ def ticket_permissions(self):
...
                 if isinstance(page, Topic):
                     yield root_id or page.id, page
 
                 yield from visit_topics(
                     page.children,
-                    root_id=root_id or page.id
+                    root_id=root_id
                 )
 
+        return list(visit_topics(self.root_pages))
```
Daverball, line 210:
```suggestion
        return tuple(visit_topics(self.root_pages))
```
`list` is mutable, so it can't be cached properly (it will work in the second-tier redis cache, but not the first tier request cache for subsequent calls in the same request), you need to use a hashable type, such as `tuple`.

src/onegov/org/app.py
```diff
@@ -193,23 +193,25 @@ def ticket_permissions(self):
         return result
 
     @orm_cached(policy='on-table-change:pages')
-    def homepage_pages(self) -> dict[int, list[Topic]]:
-
+    def topics_cached(self) -> list[tuple[int, 'Topic']]:
```
Daverball, line 196:
```suggestion
    def topics_cached(self) -> tuple[tuple[int, 'Topic'], ...]:
```
As a consequence from changing the return statement below

Daverball (changes requested):
Looks good for the most part, but there's a regression and a subtle bug due to the use of `orm_cached`

Daverball (approved):

PR #1016, opened 2023-09-22
Move each filter element in separate tag preventing layout issues (in…

Tschuppi81:
Events: Every filter element goes into its own field to prevent layout issues in case when many filter elements are selected

TYPE: Bugfix
LINK: ogc-1219


Tschuppi81:
@BreathingFlesh Could you have a look at my css? The 'margin-right' has no effect although if enter it manually in the browser dev tools it works...

PR #1027, opened 2023-09-26
Validate if filter keywords not used in EventForm

Tschuppi81:
Events: Validate event filter definition does not use names used in `EventForm`

TYPE: Feature
LINK: ogc-1219

### DIFF
--- a/src/onegov/org/forms/event.py
+++ b/src/onegov/org/forms/event.py
@@ -345,6 +345,7 @@ class EventForm(Form):
             'end_time',
             'weekly',
             'email',
+            # tschupre ?
         })
 
         # clear the recurrence to avoid updating all occurrences too much
@@ -653,3 +654,20 @@ class EventConfigurationForm(Form):
             'class_': 'formcode-select',
             'data-fields-include': 'radio,checkbox'
         })
+
+    def validate(self):
+        """
+        Ensures none of the active keywords is used in the `EventForm` to
+        prevent overriding data accidentally by defining filters.
+        """
+        result = super().validate()
+
+        for keyword in [k.lower() for k
+                        in self.keyword_fields.data.splitlines()]:
+            if keyword in dir(EventForm):
+                error = f'Keyword \'{keyword}\' is already in use!'
+                self.definition.errors.append(error)
+                self.keyword_fields.errors.append(error)
+                result = False
+
+        return result
--- a/src/onegov/org/forms/event.py
+++ b/src/onegov/org/forms/event.py
@@ -345,7 +345,6 @@ class EventForm(Form):
             'end_time',
             'weekly',
             'email',
-            # tschupre ?
         })
 
         # clear the recurrence to avoid updating all occurrences too much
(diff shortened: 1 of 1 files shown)


Tschuppi81:
Now will not save the filters if already existing names are used
Screenshot from 2023-09-26 14-02-07

src/onegov/org/forms/event.py
```diff
@@ -653,3 +653,20 @@ class EventConfigurationForm(Form):
...
+    def validate(self):
+        """
+        Ensures none of the active keywords is used in the `EventForm` to
+        prevent overriding data accidentally by defining filters.
+        """
+        result = super().validate()
+
+        for keyword in [k.lower() for k
+                        in self.keyword_fields.data.splitlines()]:
+            if keyword in dir(EventForm):
```
Daverball, line 666:
```suggestion
            if keyword in EventForm:
```
`Form` implements `__contains__` for field names, `dir` should be unnecessary. `dir` is a little bit expensive, so it's better if we can get away with not using it.

src/onegov/org/forms/event.py
```diff
@@ -653,3 +653,20 @@ class EventConfigurationForm(Form):
...
+                error = f'Keyword \'{keyword}\' is already in use!'
```
Daverball, line 667:
```suggestion
                error = self.definition.gettext(_(
                    'Field name \'${name}\' is already in use!',
                    mapping={'name': keyword})
                ))
```
i18n support

Daverball: Also custom validators are generally defined in a separate function prefixed by `ensure_` you don't need to override the `validate` method.

Daverball: Also does `display` actually matter? Shouldn't `definition` be merged into `EventForm` regardless? I think you only really need to check definition, in which case it makes more sense to put this validation into `ValidFilterFormDefinition`

Daverball: @Tschuppi81 Actually `ValidFormDefinition` already has `reserved_fields` arguments, so you can override `__init__` in `ValidFilterForm` to call `self.reserved_fields.update(name for name, __ in EventForm._unbound_fields)`. You will need to move the validator from `onegov.form` to `onegov.org` which makes more sense to me anyways, since you are only using this validator there.

Tschuppi81: Ok, I' ll try

Tschuppi81: The `reserved_fields` of `ValidFormDefinition` did all I needed!

src/onegov/org/forms/event.py
```diff
@@ -642,6 +642,8 @@ class EventConfigurationForm(Form):
             ValidFilterFormDefinition(
                 require_email_field=False,
                 require_title_fields=False,
+                reserved_fields={name for name, _ in
+                                 EventForm()._unbound_fields}
```
Daverball, line 646:
```suggestion
                                 EventForm._unbound_fields}
```
You don't need to create an instance, `_unbound_fields` is a class attribute.

Daverball: You could also use `onegov.form.utils.get_fields_from_class` which works even if no instance has been created yet.

Tschuppi81: Better, then creating an instance!

Daverball (approved):

PR #1043, opened 2023-10-16
Org: Automatic archiving and deletion of tickets.

cyrillkuettel:
## Commit message

Org: Automatic archiving and deletion of tickets.

TYPE: Feature
LINK: OGC-58

## Checklist


src/onegov/org/forms/settings.py
```diff
@@ -1132,3 +1133,39 @@ class EventSettingsForm(Form):
...
+def generate_timespans() -> (
+    tuple[tuple[str | datetime.timedelta, str], ...]
+):
+    return (
+        ('disabled', _('Disabled')),
+        (datetime.timedelta(days=180), _('6 months')),
+        (datetime.timedelta(days=365), _('1 year')),
+        (datetime.timedelta(days=365 * 2), _('2 years')),
+        (datetime.timedelta(days=365 * 3), _('3 years')),
+    )
```
Daverball, line 1147: Why is this a function? As far as I can tell there's nothing dynamic being generated here.

I would move this into a global constant, e.g. `TIMESPAN_CHOICES`. 

Also it's worth noting that this doesn't actually work because both `SelectField` and `RadioField` default their `coerce` argument to `str`, i.e. all the values are turned into strings.

I think the sanest way to solve this would be to pass `coerce=int` and use integers for the choices, then you can use `0` for disabled. Then you can either override `process_obj`/`populate_obj` as you have or simplify things by leaving the setting as an integer and convert it to a timedelta in the cronjob.

src/onegov/org/forms/settings.py
```diff
@@ -1132,3 +1133,39 @@ class EventSettingsForm(Form):
...
+
+    auto_delete_timespan = RadioField(
+        label=_('Duration from archived state until deleted automatically'),
+        validators=[InputRequired()],
+        default='disabled',
+        choices=generate_timespans()
+    )
+
+    def process_obj(self, model):
+        super().process_obj(model)
```
Daverball, line 1167: You don't actually need to call `super().process_obj(model)` here, since you are manually setting the only fields in the form.

But by adressing the above comment you may be able to get rid of this entirely anyways.

src/onegov/org/models/ticket.py
```diff
@@ -376,6 +374,13 @@ def reservations(self):
 
         return tuple(query)
 
+    def has_future_reservation(self):
+        exists = self.session.query(Reservation).filter(
+            Reservation.start > func.now()
+        ).limit(1).all()
+
+        return bool(exists)
```
Daverball, line 382:
```suggestion
        exists = self.session.query(Reservation).filter(
            Reservation.start > func.now()
        ).exists()

        return self.session.query(exists).scalar()
```
An exists subquery is much more efficient than a query limited to 1 record. (It also avoids the ORM overhead)

But also, shouldn't you filter the reservations to the ones attached to the ticket? Otherwise you will only ever be able to delete tickets if no future reservations exist at all.

src/onegov/org/models/ticket.py
```diff
@@ -415,6 +420,28 @@ def undecided(self):
 
         return True
 
+    def prepare_delete_ticket(self):
+        if self.reservations:
+            for reservation in self.reservations:
+                self.session.delete(reservation)
```
Daverball, line 426:
```suggestion
        for reservation in self.reservations or ():
            self.session.delete(reservation)
```

src/onegov/org/models/ticket.py
```diff
@@ -663,6 +690,23 @@ def extra_data(self):
     def undecided(self):
         return self.event and self.event.state == 'submitted'
 
+    def prepare_delete_ticket(self):
+        if self.event:
+            self.session.delete(self.event)
```
Daverball, line 695: Have we fully thought this through, should past events actually ever be deleted?

src/onegov/org/views/ticket.py
```diff
@@ -1069,6 +1074,72 @@ def action_link(ticket):
...
+def delete_tickets_and_related_data(
+    request: 'CoreRequest', tickets: 'Query[Ticket]'
+) -> tuple[list['Ticket'], list['Ticket']]:
+
+    not_deletable, successfully_deleted = [], []
+
+    for ticket in tickets:
+        if (hasattr(
+                ticket.handler, 'ticket_deletable')
+                and not ticket.handler.ticket_deletable):
```
Daverball, line 1107:
```suggestion
        if not getattr(ticket.handler, 'ticket_deletable', True):
```
`hasattr` + attribute access can be simplified to `getattr` with default.

Although I think the default should be flipped, it doesn't really seem safe to me to delete a ticket when the handler doesn't specify whether it should be deleted or not.

Daverball: But also I would honestly just add `ticket_deletable` with a sane default to the base handler class, then you don't need `getattr`/`hasattr` at all.

src/onegov/org/views/ticket.py
```diff
@@ -1069,6 +1074,72 @@ def action_link(ticket):
...
+            not_deletable.append(ticket)
+
+            ticket.redact_data()
+            continue
+
+        delete_messages_from_ticket(request, ticket.number)
+
+        if hasattr(ticket.handler, 'prepare_delete_ticket'):
```
Daverball, line 1115: Same here, if this is added to the base handler, then we don't need this `hasattr` check.

src/onegov/org/views/ticket.py
```diff
@@ -1069,6 +1074,72 @@ def action_link(ticket):
...
+    for message in messages.query():
+        messages.delete(message)
+
+
+def delete_files_and_submissions_from_ticket(request: 'CoreRequest',
+                                             ticket: 'Ticket'):
+    submission = getattr(ticket.handler, 'submission', None)
+    if submission:
+        for file in submission.files:
+            request.session.delete(file)
```
Daverball, line 1139:
```suggestion
```
I think you can get rid of this, `cascade` should take care of this for you.

Daverball (changes requested):
Looks good for the most part, but there are still some bugs and rough edges here and there.

src/onegov/agency/models/ticket.py
```diff
@@ -55,6 +55,18 @@ def mutation(self):
     def deleted(self):
         return self.agency is None
 
+    def prepare_delete_ticket(self):
+        pass
```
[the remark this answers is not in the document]
Daverball: Actually I would go a step further and just add this to `Handler`, maybe even get rid of `TicketDeletionMixin`. Are there any handlers that don't use this mixin?

src/onegov/org/forms/settings.py
cyrillkuettel: ah yes probably better than handling the strings. This itched me anyway, having to depend on string parsing of datetime for this critical part of the code.

src/onegov/org/models/ticket.py
cyrillkuettel: Very good catch here, this got removed when I refactored. the previous version I had was using `self.reservations` which is what we probably want here.

src/onegov/org/models/ticket.py
cyrillkuettel: It would only apply to those which have no future occurrence, else it would not be `deletable`. 

`future_occurrences. (https://github.com/OneGov/onegov-cloud/blob/76cd47f707f42b0f41f0a8d4e369fba985c5a966/src/onegov/org/models/ticket.py#L705C34-L705C34)

But maybe we should reconsider it anyway...

src/onegov/org/views/ticket.py
cyrillkuettel: Probably yes, I'll have a test for that.

src/onegov/org/forms/settings.py
cyrillkuettel: And  it's a function because we would have to duplicate it two times for the both fields of this form....

src/onegov/org/models/ticket.py
Daverball: Yes, but it still seems weird, since events are public information, tickets are not.

src/onegov/org/views/ticket.py
cyrillkuettel: Seems obvious in hindsight. There's not really a reason not to add it to the base handler.

src/onegov/org/models/ticket.py
cyrillkuettel: I still can't believe I missed that. Might have to add a test for that

src/onegov/agency/models/ticket.py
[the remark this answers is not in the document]
cyrillkuettel: The mixin is everywhere. I just wanted to be "extra safe" like what if someone happens to write a handler in the future without `TicketDeletionMixin`. That was the original idea. But moving it up the inheritance hierarchy seems like the way to go.

src/onegov/org/models/ticket.py
cyrillkuettel: Yeah. I think we should not do that. What's also just struck me just now is that `view_event` makes a query for the associated ticket. If it does not exist, it will open a new one !?  This might result in unexpected things happening. (https://github.com/OneGov/onegov-cloud/blob/9af704438ccae09b3a416704b4fcc02d99c19d23/src/onegov/org/views/event.py#L256)

This is all getting a bit convoluted.

src/onegov/org/forms/settings.py
Daverball: you don't need a function to use the same choices in multiple fields. We have other fields that share common choices, such as `WEEKDAYS`, those are defined in a global variable as well.

cyrillkuettel: ah right, it only needs computed once.

src/onegov/org/cronjobs.py
```diff
@@ -459,3 +460,53 @@ def form(definition):
...
+
+    archive_timespan = parse_to_timedelta(archive_timespan)
+
+    diff = utcnow() - archive_timespan
+    query = session.query(Ticket)
+    query = query.filter(Ticket.state == 'closed')
+    query = query.filter(Ticket.created <= diff)
+
+    for ticket in query:
+        ticket.archive_ticket()
```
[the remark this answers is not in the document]
cyrillkuettel: Done!

src/onegov/feriennet/models/activity.py
```diff
@@ -138,6 +138,10 @@ def undecided(self):
 
         return self.activity.state == 'proposed'
 
+    @property
+    def ticket_deletable(self):
+        return self.activity.state != 'archived' and super().ticket_deletable()
```
Daverball, line 143: Reversing the order of these conditions might be slightly more efficient, since I'm not sure if `self.activity` is eagerly loaded or not.

src/onegov/org/cronjobs.py
```diff
@@ -47,6 +48,8 @@ def hourly_maintenance_tasks(request):
     publish_files(request)
     reindex_published_models(request)
     send_scheduled_newsletter(request)
+    # for testing
+    # archive_old_tickets(request)
```
Daverball, line 52: Is the idea here that we don't enable this part of the cronjob until a later date?

src/onegov/org/cronjobs.py
```diff
@@ -459,3 +462,50 @@ def form(definition):
...
+
+    archive_timespan = timedelta(days=archive_timespan)
+
+    diff = utcnow() - archive_timespan
+    query = session.query(Ticket)
+    query = query.filter(Ticket.state == 'closed')
+    query = query.filter(Ticket.created <= diff)
+
+    for ticket in query:
+        if getattr(ticket.handler, 'has_future_reservation', False):
```
Daverball, line 488: This is not correct, a method is always truthy, you actually need to call the method:
```suggestion
        if getattr(ticket.handler, 'has_future_reservation', lambda: False)():
```

src/onegov/org/models/organisation.py
```diff
@@ -179,6 +179,10 @@ class Organisation(Base, TimestampMixin):
     gever_password = meta_property()
     gever_endpoint = meta_property()
 
+    # data retention policy
+    auto_archive_timespan = meta_property()
+    auto_delete_timespan = meta_property()
```
Daverball, line 184: If you change these to have a default value then you don't need a custom `process_obj` method in the form.
```suggestion
    auto_archive_timespan = meta_property(default=0)
    auto_delete_timespan = meta_property(default=0)
```

src/onegov/org/models/ticket.py
```diff
@@ -366,15 +337,25 @@ def resource(self):
         return query.one()
 
     @cached_property
-    def reservations(self):
+    def reservations_query(self):
```
Daverball, line 340: It's probably not worth to make the query a `cached_property`. I would change this to `property`

src/onegov/org/models/ticket.py
```diff
@@ -366,15 +337,25 @@ def resource(self):
...
         query = query.filter(Reservation.token == self.id)
         query = query.order_by(Reservation.start)
+        return query
 
-        return tuple(query)
+    @cached_property
+    def reservations(self):
+        return tuple(self.reservations_query)
+
+    def has_future_reservation(self):
```
Daverball, line 353: Maybe we should make this a `cached_property`, unless you're confident we only call this once per request. This will also mean changing all the calls to simple attribute accesses.

src/onegov/org/models/ticket.py
```diff
@@ -415,6 +396,14 @@ def undecided(self):
 
         return True
 
+    def prepare_delete_ticket(self):
+        for reservation in self.reservations or ():
+            self.session.delete(reservation)
+
+    @property
+    def ticket_deletable(self):
+        return not self.has_future_reservation() and super().ticket_deletable()
```
Daverball, line 405: I'm not sure how `super()` interacts with properties, but I think this should still be `super().ticket_deletable` without the call.

src/onegov/org/templates/cautionary_form.pt
Daverball, line 1: Wouldn't a `request.alert(msg)` in the view be sufficient here? Is it actually worth adding a new template for this?

src/onegov/ticket/handler.py
```diff
@@ -212,6 +212,26 @@ def get_links(self, request: 'CoreRequest') -> 'Sequence[_LinkOrCallback]':
...
+            # needed for exports
+            return False
+        if self.undecided:
+            return False
+        return True
+
+    def prepare_delete_ticket(self) -> None:
+        """The handler knows best what to do when a ticket is called for
+        deletion. """
+        assert self.ticket_deletable
```
Daverball, line 232: This assertion suggest to me that we might want `ticket_deletable` to change to a `cached_property` in some cases, otherwise we do the work of figuring out if the ticket is deletable twice.

Daverball (changes requested):
Looks good for the most part, but there are still a couple of small problems.

src/onegov/org/cronjobs.py
Daverball: If you decide to change this to a `cached_property`, like suggested further below, then this doesn't need to change.

src/onegov/org/cronjobs.py
cyrillkuettel: This needs to be removed, was just for local testing. I have configured both jobs to run in the night (e.g. once every 24 hours)

src/onegov/org/cronjobs.py
cyrillkuettel: Good catch here. If we change `has_future_reservation` to be cached_property, this ` lambda: False)()`  would not be necessary ?

Daverball: correct

src/onegov/org/templates/cautionary_form.pt
cyrillkuettel: yes, actually that's right. New template is overkill

src/onegov/org/cronjobs.py
cyrillkuettel: That would also be a good case for a linter. I'm actually kind of surprised

Daverball: mypy would catch this, but as long as this doesn't have type annotations it will not.

Daverball: Actually nevermind the `getattr` just inherently makes this difficult to statically analyze.

cyrillkuettel: I was contemplating using `isinstance` anyway. Do you know if there's not much of a performance difference between that and `getattr` ?

src/onegov/ticket/handler.py
cyrillkuettel: I'm kind of questioning if the assertion is even worth it to write. That was there before in  `prepare_delete_ticket`, I did not add this

Daverball: It's a pretty good sanity check to make sure this only gets called for tickets that are actually deletable.

src/onegov/org/cronjobs.py
Daverball: shouldn't be a big hit, there's some things like runtime checkable protocols which would be slow in an `isinstance` check and there are other classes that implement their own `__instancecheck__` which has a potential to be slow, but this shouldn't apply here.

cyrillkuettel:
Anything I missed, are we all set? @Daverball

Daverball:
@cyrillkuettel You will need to do another merge to resolve conflicts, otherwise I think we're good to go, but I will double check after the merge just in case.

Daverball (approved):

PR #1055, opened 2023-10-30
Adds CC field

cyrillkuettel:
## Commit message

Org: Adds CC field and attachments to Messages

TYPE: Feature
LINK: OGC-982

## Checklist

### DIFF
--- a/src/onegov/core/mail.py
+++ b/src/onegov/core/mail.py
@@ -155,7 +155,7 @@ def format_address(addresses: 'SequenceOrScalar[Address | str]') -> str:
 
 class Attachment:
     """
-    Represents a mail attachment that can be passed to prepare_mail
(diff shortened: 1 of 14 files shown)


BreathingFlesh (commented):
The mails with attachment somehow don't seem to be sent when I test it.

cyrillkuettel:
> The mails with attachment somehow don't seem to be sent when I test it.

I was able to reproduce. Happens only *sometimes*. But the mails are actually sent.  The sent emails are also here:
```
find /usr/local/var/onegov/mail/smtp -type f -printf '%T+ %p\n' | sort -r | head -n 1
```
Then you can search in this file for "Bcc" and "Attachments". I suppose this is a bug in smtp4dev itself?

PR #1060, opened 2023-11-06
Ogc 631 map

BreathingFlesh:

## Commit message

Org: Add Zugmap

TYPE: Feature
LINK: OGC-631
HINT: Change map settings of all instances from Zug

## Checklist


Daverball (commented):
Should we consider adding a migration? `geo-zugmap-luftbild` is no longer a valid selection, it would make sense to me to replace old values on existing`Organisation` with one of the two new values.

In case only one or two tenants are using this particular map provider, we could do it manually, but then I would add an upgrade hint to the commit, so we don't forget to do it during rollout.

PR #1061, opened 2023-11-06
Ogc 1287 landsgemeinde person

BreathingFlesh:

## Commit message

Landsgemeinde: Add Dropdown for filling out people from person directory

TYPE: Feature
LINK: OGC-1287

## Checklist

- [x] I have added new modules to the docs


src/onegov/landsgemeinde/cli.py
```diff
@@ -0,0 +1,107 @@
...
+            abort("{} already contains an organisation".format(
+                group_context.selector))
+
+        app.settings.org.create_new_organisation(app, name, locale=locale)
+
+        click.echo("{} was created successfully".format(name))
+
+    return add_org
+
+
```
Daverball, line 41:
```suggestion
```
You don't need to duplicate this command from `onegov.org` to `onegov.landsgemeinde`, since you inherit from `TownsApp` the CLI command should work just fine, you just need to use a landsgemeinde specific selector.

src/onegov/landsgemeinde/cli.py
```diff
@@ -0,0 +1,107 @@
...
+@cli.command('fill-person-info')
+@pass_group_context
+@click.option('--dry-run', is_flag=True, default=False)
+def fill_person_info(group_context, dry_run=False):
+    """ Takes the first two rows of the votum text and fills them into the
+        name and function field of the person.
+
+    Example:
+    onegov-landsgemeinde --select '/onegov_landsgemeinde/gl' fill-person-info
+    """
```
Daverball, line 51: Extend the commit message with a `HINT:` so people know to execute this command after rollout.

src/onegov/landsgemeinde/forms/votum.py
```diff
@@ -117,13 +118,33 @@ def next_number(self):
...
+                f'{p.first_name} {p.last_name}, {p.function}, '
+                f'{p.political_party}, {p.location_code_city}, '
+                f'{p.picture_url}',
+                f'{p.first_name} ' + ', '.join(filter(None, [
+                    p.last_name,
+                    p.function,
+                    p.political_party,
+                    p.location_code_city])))
+        ) for p in people]
+        people_choices.insert(0, (', , , ', '...'))
```
Daverball, line 134: Is this the correct number of `,`? The generated choices appear to use 4, rather than just 3. The same would apply to the current `default`/`choices` parameters on the field.

Daverball (commented):
Looks good, just a couple of smaller things I noticed.

src/onegov/landsgemeinde/cli.py
BreathingFlesh: Thanks for noticing. I actually merged this issue from another (unmerged) branch, this command will probably never be used, so I'm going to remove it from here again.

src/onegov/landsgemeinde/forms/votum.py
BreathingFlesh: You're right, i's not the correct number. I'll fix that.

PR #1065, opened 2023-11-07
Cleanup/Remove cli cmd and data migration script for agency address f…

Tschuppi81:
People: Cleanup/Remove cli cmd and data migration script for agency address parsed from portrait field

TYPE: Feature
LINK: ogc-1053


Tschuppi81:
@Daverball is there an easy way to verify if a migration script has been invoked on an instance similar to upgrade tasks in `upgrades` table? I want to make sure they got executed before removing them...

Daverball:
> @Daverball is there an easy way to verify if a migration script has been invoked on an instance similar to upgrade tasks in `upgrades` table? I want to make sure they got executed before removing them...

Not unless you explicitly stored something into the database within the migration script that would let you determine that.

PR #1101, opened 2023-12-12
Chat archive

BreathingFlesh:

## Commit message

Town6: Add Chat Archive

TYPE: Feature

## Checklist


src/onegov/org/models/message.py
```diff
@@ -97,7 +97,7 @@ def create(  # type:ignore[override]
         cls,
         ticket: Ticket,
         request: 'OrgRequest',
-        text: str,
+        text: str | None,
```
Daverball, line 100: As far as I could tell, there are places that rely on `TicketNote` never being `None`, so you shouldn't change this and either assert that a variable doesn't contain `None` or pass an empty string rather than `None` in places where that's a problem e.g. change from `text=foo` to `text=foo or ''`.

Currently we get the worst of both worlds, because the override in the `TYPE_CHECKING` block above claims that text is `Column[str]` rather than `Column[str | None]`, so we get neither an error for `None` when passing it in or reading it back out.

src/onegov/org/views/people.py
```diff
@@ -62,7 +63,9 @@ def visit_topics_with_people(
     }
 
 
-def person_functions_by_organization(subject_person, topics, request):
+def person_functions_by_organization(subject_person: Person,
+                                     topics: 'Iterator[Topic]',
```
Daverball, line 67: `Iterator` is a fairly strict protocol and is more commonly seen in return values rather than arguments (e.g. most generator functions and all generator comprehensions are `Iterator`s), `Iterable` is usually the better protocol, all it requires is a `__iter__` method, whereas `Iterator` also requires `__next__`.

Since you're not calling `next(topics)` anywhere `Iterable` is the correct thing to use.

src/onegov/org/views/ticket.py
```diff
@@ -445,11 +445,11 @@ def email_iter():
...
     template='ticket_note_form.pt', form=TicketNoteForm
 )
-def handle_new_note(self, request, form, layout=None):
+def handle_new_note(self: Ticket, request: 'OrgRequest', form: TicketNoteForm,
+                    layout: TicketNoteLayout | None = None):
 
     if form.submitted(request):
-        message = form.text.data,
-        form: TicketNoteForm
+        message = form.text.data
```
Daverball, line 452: Since `TicketNoteForm` uses `InputRequired` on `text` you can assert here that the message is not `None`, so this solves our mystery from above and you can change the argument back and instead add:

`assert message is not None`

Daverball (changes requested):
A couple of small things need to change, but for the most part it looks good.

Daverball (approved):

Tschuppi81 (approved):

PR #1103, opened 2023-12-13
Directory: Avoids AttributeError on entry with multiple files.

treinhard:
TYPE: Bugfix
LINK: OGC-1378

### DIFF
--- a/src/onegov/town6/layout.py
+++ b/src/onegov/town6/layout.py
@@ -2144,11 +2144,12 @@ class DirectoryEntryBaseLayout(DefaultLayout):
         thumbnail = self.thumbnail_field_id
         if not thumbnail:
             return
-        if thumbnail in entry.values:
-            item = entry.values[thumbnail]
-            if 'data' in item:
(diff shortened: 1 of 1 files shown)


src/onegov/town6/layout.py
```diff
@@ -2144,7 +2144,11 @@ def thumbnail_file_id(self, entry):
         thumbnail = self.thumbnail_field_id
         if not thumbnail:
             return
-        return (entry.values.get(thumbnail) or {}).get('data', '').lstrip('@')
+        if thumbnail in entry.values:
+            item = entry.values[thumbnail]
+            if 'data' in item:
+                return item['data'].lstrip('@')
+        return
```
Daverball, line 2151:
```suggestion
        value = entry.values.get(thumbnail)
        if isinstance(value, list) and value:
            value = value[0]            
        if not value:
            return
        return value.get('data', '').lstrip('@')
```
It probably makes sense to use the first picture as a thumbnail if it's a multiple file field.

PR #1118, opened 2023-12-21
Adds option to show people names in western ordered per topic

Tschuppi81:

## Commit message

<Module>: <Message>

<Optional Description>

TYPE: <Feature|Bugfix>
LINK: <Ticket-Number>
HINT: <Optional upgrade hint>

## Checklist

### DIFF
--- a/src/onegov/org/models/extensions.py
+++ b/src/onegov/org/models/extensions.py
@@ -404,6 +404,9 @@ class PersonLinkExtension(ContentExtension):
                 self.apply_model(obj)
 
             def update_model(self, model):
+                model.show_western_name_order = self._fields[
+                    'western_ordered'].data
+
                 previous_people = model.content.get('people', [])
 
                 if self.is_ordered_people(previous_people):
@@ -438,6 +441,9 @@ class PersonLinkExtension(ContentExtension):
                     model.content['people'] = old_people + new_people
 
             def apply_model(self, model):
+                self._fields['western_ordered'].data = (
+                    model.show_western_name_order)
+
                 fields = self.get_people_fields(with_function=False)
                 people = dict(model.content.get('people', []))
 
@@ -452,12 +458,25 @@ class PersonLinkExtension(ContentExtension):
         builder = WTFormsClassBuilder(PeoplePageForm)
         builder.set_current_fieldset(fieldset_label)
 
-        for person in self.get_selectable_people(request):
+        selectable_people = self.get_selectable_people(request)
+        if selectable_people:
+            builder.add_field(
+                field_class=BooleanField,
+                field_id='western_ordered',
+                label=request.translate(_("Use Western ordered names e.g. "
+                                          "Franz Müller instead of Müller "
+                                          "Franz")),
+                required=False,
+                default=True if self.show_western_name_order else False,
+            )
+        for person in selectable_people:
             field_id = fieldset_id + '_' + person.id.hex
+            name = ' '.join([person.first_name, person.last_name]) if (
+                self.show_western_name_order) else person.title
             builder.add_field(
                 field_class=BooleanField,
                 field_id=field_id,
-                label=person.title,
+                label=name,
                 required=False,
                 id=person.id,
             )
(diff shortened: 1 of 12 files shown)


Daverball (commented):
You don't need to ignore `note` messages, they're just hints and won't make mypy fail. The reason for the mypy failure is the upgrade to 1.8.0 which caused the latest version of typeshed to be pulled which changed the definition of the logging kwargs, for now you can just pin mypy to `<1.8.0`, unless you want to try changing the import to the suggested name instead, but that's probably going to cause some other failures, because we actually only want the optional keyword arguments, so we will probably need to define those ourselves, so that's why pinning mypy for now is probably the quickest solution.

src/onegov/org/models/extensions.py
```diff
@@ -452,12 +461,25 @@ def apply_model(self, model):
...
+                field_id='western_ordered',
+                label=_("Use Western ordered names"),
+                description=_("For instance Franz Müller instead of Müller "
+                              "Franz"),
+                required=False,
+                default=self.show_western_name_order,
+            )
+        for person in selectable_people:
             field_id = fieldset_id + '_' + person.id.hex
+            name = ' '.join([person.first_name, person.last_name]) if (
```
Daverball, line 477:
```suggestion
            name =f '{person.first_name} {person.last_name}' if (
```
Don't use a join with a static sequence, just write the f-string or use `str.format`.

src/onegov/org/templates/macros.pt
```diff
@@ -562,7 +562,8 @@
                     <h2 i18n:translate>People</h2>
                     <ul class="panel-links" data-sortable data-sortable-url="${layout.move_person_url_template}">
                         <li tal:repeat="person people" data-sortable-id="${request.is_manager and person.id}" class="${repeat.person.index > 2 and 'hidden hideable' or ''}">
-                            <a href="${request.link(person)}">${person.title}</a>
+                            <a tal:condition="western_ordered" href="${request.link(person)}">${person.first_name} ${person.last_name}</a>
+                            <a tal:condition="not western_ordered" href="${request.link(person)}">${person.title}</a>
```
Daverball, line 566: This seems pretty fragile, requiring every view to pass this in. The point of extensions is, that you can rely on every model that uses the extension to have the required data accessible from the same attribute, so you don't have to do quite as much plumbing, you should be able to just do `page.show_western_name_order` here, please don't add additional template data that isn't actually required, it makes template rendering more error prone.

src/onegov/org/models/page.py
```diff
@@ -84,6 +84,26 @@ def allowed_subtraits(self):
...
+    def show_western_name_order(self, value):
+        """
+        Sets whether people of current topic shall be shown in western
+        order.
+
+        Example: "John Doe" (western) instead of "Doe John" (eastern)
+
+        """
+        self.content['western_name_order'] = value
+
```
Daverball, line 106: Why not add this as part of the extension? Also why not use `western_name_order = content_property(default=False)`?

tests/onegov/org/test_extensions.py
```diff
@@ -92,6 +92,7 @@ def test_person_link_extension():
 
     class Topic(PersonLinkExtension):
         content = {}
+        show_western_name_order = False
```
Daverball, line 95: You wouldn't need to do this if you added this to the extension

Daverball (changes requested):
Looks fine for the most part, although the naming is a bit confused and inconsistent, try to stick with one name for the same thing throughout. Also you can simplify this a whole bunch more if you actually take advantage of model extensions and move the added content property to the extension itself and always access the data from the model, rather than passing it around manually.

src/onegov/org/models/page.py
```diff
@@ -42,6 +42,8 @@ class Topic(Page, TraitInfo, SearchableContent, AccessExtension,
     # Show the lead on topics page
     lead_when_child = content_property(default=True)
 
+    western_name_order = content_property(default=False)
```
Daverball, line 45: Put this in `PersonLinkExtension` not `Topic`, `Topic` inherits from `PersonLinkExtension`, but other models do too, the form will break with those models. E.g. `News` won't work correctly now.

Tschuppi81: Yes, got it!

Daverball (approved):

PR #1120, opened 2024-01-05
Bugfix/ogc 1282 prevent move topic under latest news

Tschuppi81:
Topics: Prevents topics can be moved under a news page

TYPE: Bugfix
LINK: ogc-1282


Tschuppi81:
@BreathingFlesh had to revert 'moving import to top of file' due to circuar import problem...

PR #1122, opened 2024-01-08
Town6: Make newlines possible in lead.

cyrillkuettel:
## Commit message

Town6: Make newlines possible in lead.

The `str()` in directory.pt is used for deep copy to avoid side effects

TYPE: Feature
LINK: OGC-1328

## Checklist


src/onegov/org/templates/directory.pt
```diff
@@ -41,7 +41,7 @@
                             <tal:b define="custom_lead search_widget.lead(layout, entry)|nothing">
                                 <div class="list-lead">
                                     <tal:b condition="custom_lead" replace="structure custom_lead" />
-                                    <tal:b condition="not:custom_lead" replace="entry.lead" />
+                                    <tal:b condition="not:custom_lead" replace="structure python: str(entry.lead).replace('\\n', '<br>')" />
```
Daverball, line 44: This is not safe, you are making `entry.lead` vulnerable to HTML injection.
```suggestion
                                    <tal:b condition="not:custom_lead">
                                        <tal:b tal:repeat"line entry.lead.splitlines()">${line}<br tal:condition="not repeat.line.end"/></tal:b>
                                    </tal:b>
```
This is one potential of way making this safe inside pt. Another way is to add `Markup` and `escape` to the globals dictionary in our custom BaseTemplate class, so you can use `escape(entry.lead).replace('\\n', Markup('<br>')`

src/onegov/town6/templates/directory.pt
```diff
@@ -55,7 +55,7 @@
                                     <a tal:attributes="href request.link(entry)">
                                         <p class="list-lead preview">
                                             <tal:b condition="custom_lead" replace="structure custom_lead" />
-                                            <tal:b condition="not:custom_lead" replace="entry.lead" />
+                                            <tal:b condition="not:custom_lead" replace="structure python: str(entry.lead).replace('\\n', '<br>')" />
```
Daverball, line 58: Same thing here

src/onegov/org/templates/directory.pt
cyrillkuettel: Thank you for your review. It's great that you caught that.

cyrillkuettel: @Daverball By the way, I was curious that `Markup` seems to be imported (we don't have to import it in the template). From where is the import actually?

cyrillkuettel:
This is maybe a bit overkill, but I added a test which attempts to put a `<script>` in the html.

src/onegov/org/layout.py
```diff
@@ -2837,6 +2837,10 @@ def breadcrumbs(self):
             Link(_(self.model.title), self.request.link(self.model))
         ]
 
+    def linkify(self, text):
+        linkified = super().linkify(text)
+        return linkified.replace('\\n', '<br>') if linkified else linkified
```
Daverball, line 2842: Is the double backslash necessary here? I don't see why it would be, I feel like this actually should produce the wrong result (i.e. replacing the literal string `\n` instead of the newline character).

src/onegov/town6/layout.py
```diff
@@ -2373,6 +2373,10 @@ def breadcrumbs(self):
             Link(_(self.model.title), self.request.link(self.model))
         ]
 
+    def linkify(self, text):
+        linkified = super().linkify(text)
+        return linkified.replace('\\n', '<br>') if linkified else linkified
```
Daverball, line 2378: Same thing here.

Daverball (commented):
Looks good. The test is indeed a bit overkill, especially since the CSP Policy should protect us against XSS as well, but it can't hurt to have at least one test making sure this all works.

There's one small thing about `\\n` vs `\n` in python string literals vs. chameleon templates.

src/onegov/org/layout.py
cyrillkuettel: Yes, that's correct. If you look the ticket, we wanted to create a symbol to add newlines. It's the literal string we want.

Daverball: I see

PR #1127, opened 2024-01-10
Election Day: Adds cli util to extract the mapping

cyrillkuettel:
## Commit message

Election Day: Adds cli util to extract the mapping

TYPE: Feature
LINK: OGC-1354

## Checklist

### DIFF
--- a/do/test.py
+++ b/do/test.py
@@ -1,0 +1,67 @@
+# checks to see if all urls are valid
+# Usage example:
+# python check_urls.py /path/to/file.csv --column 2
+
+
+import argparse
+import csv
+import concurrent.futures
+import requests
+import time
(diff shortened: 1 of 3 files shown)


msom:
Why not just include our internal ID to the export?

cyrillkuettel:
> Why not just include our internal ID to the export?

I would love to. However this would create very long filenames for some cases.

Or do you mean separate? Yes we should do that. Best would be if the filenames and urls have 1:1 relationship.

msom:
No I mean add the ID to the into the file content. I will add this for votes, elections and compounds....

cyrillkuettel:
Closing this because we'll add it as a column

PR #1144, opened 2024-01-22
Resolves `FIXME`

cyrillkuettel:

## Commit message

Org: Don't assume occurrence.start exists always

<Optional Description>

TYPE: Bugfix

## Checklist

### DIFF
--- a/src/onegov/org/models/search.py
+++ b/src/onegov/org/models/search.py
@@ -86,6 +86,12 @@ class Search(Pagination[_M]):
         This methods is a wrapper around `batch.load()`, which returns the
         actual search results form the query. """
 
+        def get_sort_key(event):
+            if event.latest_occurrence and hasattr(event.latest_occurrence,
+                                                   'start'):
+                return True, event.latest_occurrence.start
+            return False, float('inf')
+
         assert self.batch is not None
         batch = self.batch.load()
         events = []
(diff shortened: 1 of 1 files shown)


src/onegov/org/models/search.py
```diff
@@ -86,6 +86,12 @@ def load_batch_results(self) -> list['Hit']:
         This methods is a wrapper around `batch.load()`, which returns the
         actual search results form the query. """
 
+        def get_sort_key(event: Event) -> tuple[bool, float]:
+            if event.latest_occurrence and hasattr(event.latest_occurrence,
+                                                   'start'):
```
Daverball, line 91:
```suggestion
            if event.latest_occurrence:
```
This should be good enough, `start` is defined on the base class and not nullable.

src/onegov/org/models/search.py
```diff
@@ -86,6 +86,12 @@ def load_batch_results(self) -> list['Hit']:
...
+                return True, event.latest_occurrence.start.timestamp()
+            return False, float('inf')
```
Daverball, line 93:
```suggestion
                return event.latest_occurrence.start.timestamp()
            return float('-inf')
```
Why not just this?

Daverball (commented):
Looks good, just a couple of minor nits.

PR #1148, opened 2024-01-23
Translator Directory: Fixes page search results resetting query

cyrillkuettel:
## Commit message

Translator Directory: Fixes page search results resetting query

Switching to a new page in search reset the search filter.

TYPE: Bugfix
LINK: OGC-1425

## Checklist


BreathingFlesh (commented):
Seems to work now, but changing the filters should probably also result in starting at page 1 again. But maybe this should be handled as a separate issue?

PR #1157, opened 2024-02-01
Feature/ogc 1408 videos in directories

Tschuppi81:
Directories: Adds support for videos

TYPE: Feature
LINK: ogc-1408


src/onegov/form/display.py
```diff
@@ -121,6 +122,78 @@ def __call__(self, field: 'Field') -> Markup:
...
+    url_video_template = """<a href=${url}>${url}</a>"""
+
+    video_page = PageTemplate(video_template)
+    url_video_page = PageTemplate(url_video_template + video_template)
+
+    def __call__(self, field: 'Field', **kwargs: Any) -> Markup:
+        url = None
+        data = self.escape(field.data)
+
+        if kwargs:
```
Tschuppi81, line 150: @Daverball I did not find a good way to distinguish between editing and displaying the video url.
My goal is to show the 'plain url' in a `StringField` when editing the directory entry and the embedded video otherwise.

Tschuppi81: TODO: Support multiple videos

src/onegov/form/display.py
```diff
@@ -121,6 +122,78 @@ def __call__(self, field: 'Field') -> Markup:
...
+            <div class="videowrapper">
+                <iframe allow="fullscreen"
+                frameborder="0" src="${url}"></iframe>
+            </div>
+        </div>
+    """
+    url_video_template = """<a href=${url}>${url}</a>"""
+
+    video_page = PageTemplate(video_template)
+    url_video_page = PageTemplate(url_video_template + video_template)
```
Daverball, line 144: For such simple templates Chameleon seems like complete overkill, just use `Markup` directly. You will need an additional replacement variable for the translation, but otherwise lingua won't pick up `Video` as a translation string anyways.

```suggestion
    video_template = Markup("""
        <div class="video">
            <h2 class="visually-hidden">{title}</h2>
            <div class="videowrapper">
                <iframe allow="fullscreen"
                frameborder="0" src="{url}"></iframe>
            </div>
        </div>
    """)
    url_video_template = Markup("""<a href="{url}">{url}</a>""")
```
Either way you don't need to manually escape, both Chameleon and Markup escape format strings automatically.

Also it's worth noting that the use of iframe is a bit questionable due to tracking, especially in non-public directories, can be sandbox the iframe for both YouTube and Vimeo to a degree where we don't have to worry about this?

src/onegov/form/display.py
Daverball: Field display is strictly for display, it should never be editable, so it does not make sense to render an input field here.

Daverball: You'll want to create a separate Widget for rendering the input, but considering the widget is just a regular url field anyways, you don't need a custom widget.

src/onegov/form/fields.py
```diff
@@ -460,6 +462,11 @@ class TextAreaFieldWithTextModules(TextAreaField):
     widget = TextAreaWithTextModules()
 
 
+class VideoURLField(URLField):
+
+    widget = VideoURLFieldRenderer()
```
Daverball, line 467:
```suggestion
    pass
```
The original widget should do the job just fine, if you need to adjust the display, then you should create a widget that derives from `URLInput` and applies any necessary changes.

src/onegov/form/display.py
Tschuppi81: Agree, just had mypy complaining about it 'MS001 Markup is used in a dangerous way' :-)

Daverball: That's a flake8 plugin, not mypy. Although mypy could potentially catch this error in the future via `LiteralString`.

src/onegov/form/display.py
```diff
@@ -160,11 +152,11 @@ def __call__(self, field: 'Field', **kwargs: Any) -> Markup:
             url = self.ensure_vimeo_embedded_url(data)
 
         if url:
-            return Markup(self.video_page.render(url=self.escape(url)))
+            return self.video_template.format(url=self.escape(url))
 
         # for other sources we try to render video url but also provide
         # the link
```
Daverball, line 158: I think this is incredibly ill-advised, since directories can accept submissions from the public it would be quite easy to inject harmful pages into iframes that would get loaded immediately without a user having to first click a link. It's already kind of debatable whether embedding youtube/vimeo videos is a good idea unless we can sandbox the iframe to a degree where it can't track where it was embedded from.

Tschuppi81: I will try to sandbox the iframe.
Regarding 'other' non-youtube, non-vimeo sources, would you not render the video at all, just show the link?

Daverball: just show the link in that case, yes

src/onegov/form/display.py
```diff
@@ -121,6 +119,71 @@ def __call__(self, field: 'Field') -> Markup:
...
+            <div class="videowrapper">
+                <iframe allow="fullscreen" frameborder="0" src="{url}"
+                sandbox="allow-scripts allow-same-origin
+                allow-presentation" referrerpolicy="no-referrer"></iframe>
+            </div>
+        </div>
+    """)
+    url_template = Markup('<a href="{url}">{url}</a>')
+
+    def __call__(self, field: 'Field', **kwargs: Any) -> Markup:
```
Daverball, line 140:
```suggestion
    def __call__(self, field: 'Field') -> Markup:
```
You no longer need kwargs here

src/onegov/form/display.py
```diff
@@ -121,6 +119,71 @@ def __call__(self, field: 'Field') -> Markup:
...
+
+        # vimeo
+        if any(x in data for x in ['vimeo']):
+            url = self.ensure_vimeo_embedded_url(data)
+
+        if url:
+            return self.video_template.format(url=self.escape(url))
+
+        # for non-vimeo and non-youtube sources we just display the url
+        return self.url_template.format(url=self.escape(data))
```
Daverball, line 156:
```suggestion
        data = field.data

        # youtube
        if any(x in data for x in ['youtube', 'youtu.be']):
            url = self.ensure_youtube_embedded_url(data)

        # vimeo
        if any(x in data for x in ['vimeo']):
            url = self.ensure_vimeo_embedded_url(data)

        if url:
            return self.video_template.format(url=url)

        # for non-vimeo and non-youtube sources we just display the url
        return self.url_template.format(url=data)
```
No need to be this defensive about escaping, `Markup` already calls `escape` on any parameter of the `format` method.

PR #1176, opened 2024-02-09
Fix mypy incompatible type error

Tschuppi81:
<Module>: <Message>

<Optional Description>

TYPE: <Feature|Bugfix>
LINK: <Ticket-Number>


Daverball:
@Tschuppi81 Update your version of libres, there was an incorrect annotation, it's already fixed, no need to change the code here.

PR #1177, opened 2024-02-09
Feature/ogc 1408 videos in directories

Tschuppi81:
Org: Adds support for videos in directory entries

TYPE: Feature
LINK: ogc-1408


Tschuppi81:
@Daverball I reverted the falsly merged previous PR (wrong merge message and lint error). So here it is again...

src/onegov/form/parser/core.py
```diff
@@ -798,6 +820,11 @@ class UrlField(Field):
     type: ClassVar[Literal['url']] = 'url'
 
 
+@final
+class VideoURLField(Field):
```
Daverball, line 824: Make sure to add `VideoURLField` to the `BasicParsedField` union above, I think you can also change the `ParsedField` union to `BasicParsedField | FileParsedField`, not sure why I haven't done that.

src/onegov/form/parser/form.py
```diff
@@ -185,6 +186,18 @@ def handle_field(
             description=field.field_help
         )
 
+    elif field.type == 'video_url':
+        # FIXME: I don't understand mypy complaints here
```
Daverball, line 190: That's exactly because it's missing from the `ParsedField` tagged union, this unreachable error goes away once you fix that.

tests/onegov/form/test_grammar.py
```diff
@@ -26,7 +26,7 @@
     time,
     url,
     valid_date_range,
-    with_whitespace_inside,
+    with_whitespace_inside, video_url,
```
Daverball, line 29:
```suggestion
    with_whitespace_inside,
    video_url,
```
Formatting nit, it's fine to have multiple imports on one line, but then you should reformat everything to remain consistent, so either everything on its own line or everything split roughly evenly among a fewer number of lines, although for avoiding merge conflicts it's usually better to have separate lines.

src/onegov/form/parser/grammar.py
```diff
@@ -306,6 +306,18 @@ def url() -> ParserElement:
...
 
+def video_url() -> ParserElement:
+    """ Returns an video url field parser.
+
+    Example::
+
+        vimeo-youtube
+
+    """
+    return Suppress(Regex(r'vimeo-youtube')).setParseAction(tag(
```
Daverball, line 317: I'm not crazy about the syntax for this field, but I also don't really have a better suggestion. Maybe ask around if someone has a good idea.

Daverball (commented):
Looks good for the most part, just a small typing change and a couple of nits.

src/onegov/form/parser/form.py
Tschuppi81: Nice!

src/onegov/form/parser/grammar.py
Tschuppi81: `video-url` was proposed by @BreathingFlesh. As we use the keyword `http://` for urls we may go for `http://video` resp. `https://video`?

Tschuppi81: Ok, will go for `video-url` and finalize on Thursday

PR #1192, opened 2024-02-16
Feature/ogc 508 replace elastic search by postgres search v2 postgres indexer

Tschuppi81:
Core: Introducing postgres full text search (fts) columns, indexer and orm event handler as well as db upgrade

TYPE: Feature
LINK: OGC-508


Daverball (changes requested):
I don't think we ever need to drop the index column, we recompute the column for every model on a re-index. So the drop + add seems wasteful.

I also don't think we need two columns, just directly convert the search data into a `TSVECTOR`. You are not really making use of the work we already do for elastic search.

All I would've done is override `index` and `delete` and use the values in `task` to perform the equivalent action in Postgres, i.e. for `index` take the `task['properties']` and convert it to a `TSVECTOR` using `task['language']` and for `delete` don't do anything. I wouldn't even really use `ensure_index` anymore, I would just write a migration to add the index column for all the models inheriting from `Searchable`.

You will need to remove `es_suggestions` from `properties` before converting them to `TSVECTOR` (and maybe there are some other transformations that make sense). Eventually we maybe want to build something that makes use of suggestions by storing them in a separate table.

I would re-use the existing `ORMEventTranslator`. We can refactor and simplify things once we get rid of elastic search. Temporarily, to make things easier you could add the table name and the name of the id column to the task, so it's easier to generate the update statement, I wouldn't use the ORM for the update statement, just issue a raw update on the engine, that way we should also not get an ORM event and don't have to worry about infinite cycles.

Tschuppi81:
Reply to: 
`I also don't think we need two columns, just directly convert the search data into a TSVECTOR. You are not really making use of the work we already do for elastic search.`
Postgres does not support creating an index on row level, just on columns. Therefore I used the second column 'fts_idx_data' the index is based on.

Tschuppi81:
Reply to: `I don't think we ever need to drop the index column, we recompute the column for every model on a re-index. So the drop + add seems wasteful.`
Agree, if there is a way to make an upgrade_task executing on every upgrade call. We just need to ensure that also newly added 'searchable' models will get the index column.

Daverball:
> Reply to: `I don't think we ever need to drop the index column, we recompute the column for every model on a re-index. So the drop + add seems wasteful.` Agree, if there is a way to make an upgrade_task executing on every upgrade call. We just need to ensure that also newly added 'searchable' models will get the index column.

I don't think this is necessary, we already have to write upgrade tasks when we add a new column to an existing table, this should be no different. We could consider adding a CLI command (or a flag for the reindex command) if you're concerned about people forgetting to add the column without anybody noticing, so we can add any missing FTS columns.

Tschuppi81:
@Daverball I reworked the entire stuff to reuse the `Indexer` class for postgres only overwriting `index`. Overall the number of changes dropped.
Open: 
- Not all rows in table `files` get indexed as some contain stuff like '&ƌŽŶůĞŝĐŚŶĂŵͲ^ĞĞƉƌŽǌĞƐƐŝŽŶ \x18ŽŶŶĞƌƐƚĂŐ͕\x03Ϭϴ͘\x03:ƵŶŝ\x03ϮϬϮϯ\x03 \x18Ğƌ\x03\x04ŶůĂƐƐ\x03ĨŝŶĚĞƚ\x03'
- Type annotation and mypy

src/onegov/search/indexer.py
```diff
@@ -307,84 +311,106 @@ def delete(self, task):
...
+            id = task['id']
+            id_key = task['id_key']
+            where_stmt = f"WHERE {id_key} = '{id}'"
+            connection = self.engine.connect()
+            trans = connection.begin()
+            connection.execute(sqlalchemy.text(f"""
+                UPDATE "{schema}".{tablename}
+                SET {self.TEXT_SEARCH_COLUMN_NAME} = {tsvector}
+                {where_stmt}
+            """))
```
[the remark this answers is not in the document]
Daverball:
```suggestion
            table = sqlalchemy.table(
                tablename,
                (id_col := sqlalchemy.column(id_key)),
                sqlalchemy.column(self.TEXT_SEARCH_COLUMN_NAME),
                schema=schema
            )
            tsvector_expr = text('to_tsvector(:language, :data)').bindparams(
                sqlalchemy.bindparam('language', type_=sqlalchemy.String),
                sqlalchemy.bindparam('language', type_=JSONB)
            )
            stmt = (
                sqlalchemy.update(table)
                .where(id_col == id)
                .values({self.TEXT_SEARCH_COLUMN_NAME: tsvector_expr})
            )
            connection = self.engine.connect()
            trans = connection.begin()
            connection.execute(stmt, {'language': language, 'data': data})
```
Something like this should do the trick and be safe from SQL injection

src/onegov/search/indexer.py
```diff
@@ -307,84 +311,106 @@ def delete(self, task):
...
+            data = []
+            languages = ['simple']
+
+            properties = task['properties']
+            if task['language'] in self.language_mapping:
+                languages.append(self.language_mapping[task['language']])
+
+            data = self.get_data_from_properties(data, properties)
+            if not data:
+                return True
```
Daverball, line 344: You can't actually early out here, what if we had data before and now we don't? In this case we need to set an empty `TSVECTOR`, which should be guaranteed if `data` happens to be an empty dictionary.

src/onegov/search/indexer.py
```diff
@@ -307,84 +311,106 @@ def delete(self, task):
...
-                # endless loop
-                obj.fts_idx_data = data
+            data = []
+            languages = ['simple']
+
+            properties = task['properties']
+            if task['language'] in self.language_mapping:
+                languages.append(self.language_mapping[task['language']])
+
+            data = self.get_data_from_properties(data, properties)
```
Daverball, line 342: Don't try to recompute the data, we already have the data embedded in the task:
```suggestion
            data = task['properties'].copy()
            # TODO: handle suggestions
            del data['es_suggestion']
```

src/onegov/search/indexer.py
```diff
@@ -307,84 +311,106 @@ def delete(self, task):
...
-                # this check prevents unnecessary updates causing 'update'
-                # events triggering reindexing the object leading to an
-                # endless loop
-                obj.fts_idx_data = data
+            data = []
+            languages = ['simple']
+
+            properties = task['properties']
+            if task['language'] in self.language_mapping:
+                languages.append(self.language_mapping[task['language']])
```
Daverball, line 340:
```suggestion
            language = self.language_mapping.get(task['language'], 'simple')
```

Daverball (changes requested):
We're getting pretty close now. There is a safety issue with the generated SQL statement however and there's still a bit of code that can be simplified away.

Daverball:
>     * Not all rows in table `files` get indexed as some contain stuff like '&ƌŽŶůĞŝĐŚŶĂŵͲ^ĞĞƉƌŽǌĞƐƐŝŽŶ \x18ŽŶŶĞƌƐƚĂŐ͕\x03Ϭϴ͘\x03:ƵŶŝ\x03ϮϬϮϯ\x03 \x18Ğƌ\x03\x04ŶůĂƐƐ\x03ĨŝŶĚĞƚ\x03'

That's to be expected, since we extract the text from the PDF which may use some weird symbol fonts to draw things, or it's also possible that some binary data is accidentally interpreted as text. In either case there's not much we can do. I'm not sure if the errors were due to your unsafe SQL generation or because it actually had an issue with the unicode data. The same goes for any size limits, which should be less strict for bind params.

So only do some kind of workaround, if it's still needed after the changes and I'd always try to commit the data unchanged first and then we can try to apply some transformations, we'll just have to come up with some sane heuristics, like dropping everything from the data that isn't ASCII and/or limiting individual data keys to a length an order of magnitude below the maximum length.

Tschuppi81:
Thank you for all your suggestions and code snippets @Daverball!
Finally no errors during indexing, not event for files :-)

I was thinking about logging 'indexing' issues to a separate log file, which shall be overwritten with each `reindex` call. This would help us to identify indexing issues on production servers. What do you think about that?

src/onegov/search/indexer.py
Tschuppi81: agree

src/onegov/search/indexer.py
Tschuppi81: I adapted it to remove all keys starting with 'es_'

Tschuppi81:
> > I was thinking about logging 'indexing' issues to a separate log file, which shall be overwritten with each `reindex` call. This would help us to identify indexing issues on production servers. What do you think about that?
> 
> It certainly can't hurt, although how would you configure that? Write it to the same directory that the log config specifies for the default log file (if it even has one configured)? It seems a little bit clunky.
> 
> I suppose we can also just extend the `onegov.yml` with a log config section for `onegov.search` that writes to a different log file than the root logger. This would be a change in the puppet repository and not in OneGov itself. Or add another sublogger `onegov.search.reindex` if you want it to be even more granular, but how would distinguish between a reindex and just a regular update to a model?

Yes, I was thinking of a different logger in search/indexer and maybe in search/integration which would log to a different file

Daverball:
> Yes, I was thinking of a different logger in search/indexer and maybe in search/integration which would log to a different file

Yes, you can entirely achieve this through `onegov.yml` just add a new file handler to `handlers` and add a  `onegov.search` entry for `loggers` that uses that new handler.

src/onegov/search/__init__.py
```diff
@@ -2,6 +2,8 @@
 log = logging.getLogger('onegov.search')
 log.addHandler(logging.NullHandler())
 
+index_log = logging.getLogger('onegov.search.index')
+
```
Daverball, line 6: You should still add the `NullHandler` here too, so you don't get logs for it unless you turn it on via the config file.

src/onegov/search/integration.py
```diff
@@ -375,24 +401,30 @@ def reindex_model(model):
...
-                    for model in searchable_sqlalchemy_models(base)
-                )
+                reindex_model, (model for model in models)
             )
             if fail:
-                tuple(results)
+                print(tuple(results))
 
         self.es_indexer.bulk_process()
+        self.psql_indexer.process()
```
Daverball, line 425: Eventually we may want to implement `bulk_process` for the `psql_indexer` in order to improve performance of the complete re-index, if you feel like it you can take a stab at it right now:

You would gather all the `index` tasks grouped by model (you can use `task['type_name']` as the groupby key, they're already sorted by model, because that's how we generated the tasks in the first place) and use `executemany` instead of `execute` to reindex all the relevant rows at once. You'd probably have to refactor `index` slightly, so the part that generates the query can be reused, you'd need to turn `id` into a substitution value like `language` and `data` so the three values can be passed as a list of dictionaries to `executemany`.

Tschuppi81: ok, will try that later

src/onegov/search/indexer.py
```diff
@@ -306,8 +307,105 @@ def delete(self, task):
...
+            index_log.error(f'Error \'{ex}\' indexing object id \'{_id}\' '
+                            f'in schema \'{schema}\' of table \'{tablename}\'')
+            return False
+
+        return True
+
+    def delete(self, task):
+        return True
 
+    def bulk_process(self):
```
Tschuppi81, line 387: Should I index in batches of e.g. 1000  tasks?

For Meggen the number of documents/tasks are:
```
2024-03-07 13:31:51 - INFO: Indexing schema onegov_town6-meggen..               
2024-03-07 13:31:51 - INFO: Number of models to be indexed: 39                  
2024-03-07 13:32:14 - INFO: Indexing 3 tasks for directories                    
2024-03-07 13:32:14 - INFO: Indexing 334 tasks for pages                        
2024-03-07 13:32:14 - INFO: Indexing 7 tasks for external_links                 
2024-03-07 13:32:14 - INFO: Indexing 36 tasks for users                         
2024-03-07 13:32:14 - INFO: Indexing 4 tasks for filesets                       
2024-03-07 13:32:14 - INFO: Indexing 211 tasks for people                       
2024-03-07 13:32:14 - INFO: Indexing 3 tasks for resources                      
2024-03-07 13:32:14 - INFO: Indexing 167 tasks for directory_entries            
2024-03-07 13:32:14 - INFO: Indexing 315 tasks for events                       
2024-03-07 13:32:14 - INFO: Indexing 64 tasks for forms                         
2024-03-07 13:32:14 - INFO: Indexing 2969 tasks for tickets                     
2024-03-07 13:32:14 - INFO: Indexing 357 tasks for files                        
2024-03-07 13:32:14 - INFO: Done  
``

File: src/onegov/search/indexer.py [hunk 8]
```diff
@@ -306,8 +307,105 @@ def delete(self, task):
...
+                sqlalchemy.bindparam('data', type_=sqlalchemy.JSON)
+            )
+            stmt = (
+                sqlalchemy.update(table)
+                .where(id_col == sqlalchemy.bindparam('_id'))
+                .values({self.TEXT_SEARCH_COLUMN_NAME: tsvector_expr})
+            )
+            connection = self.engine.connect()
+            trans = connection.begin()
+            connection.execute(stmt, content)  # executemany supported with
```
Tschuppi81, line 374: Unfortunately `executemany` is not supported in sqlalchemy 1.3

src/onegov/search/indexer.py
Daverball: Good question, not sure if it will be worth batching, but considering that on a complete re-index we will receive the tasks pre-sorted by table it might be more memory-efficient to use `itertools.groupby` and immediately call `index` on each group, rather than first collect all of the tasks into a dictionary.

So do something closer to the elasticsearch version, where you define a generator function, then use `itertools.groupby` to group the tasks returned by the generator and submit the group as a batch. If SQLAlchemy struggles with large batches we can always use `onegov.core.utils.batched` to split each group into smaller batches.

src/onegov/search/indexer.py
Daverball: I believe SQLAlchemy will automatically use either `execute` or `executemany` on the underlying DBAPI 2.0 cursor underneath depending on whether you pass a sequence of dictionaries or a single dictionary. So I don't think we need to worry about this.

I don't think they ever matched DBAPI 2.0 and separated it into two functions.

src/onegov/search/indexer.py
```diff
@@ -323,27 +323,41 @@ def __init__(self, mappings, queue, es_client, engine, session):
...
-            id = task['id']
-            id_key = task['id_key']
+            for task in tasks:
+                language = (
+                    self.idx_language_mapping.get(task['language'], 'simple'))
+                data = {
+                    k: str(v)
+                    for k, v in task['properties'].items()
+                    if not k.startswith('es_')}
+                _id = str(task['id'])
```
Daverball, line 349: Have you tried leaving `task['id']` unmodified? I worry SQLAlchemy will do the wrong thing here when binding the parameter, because it doesn't know the type of the id column in the table. Technically the id should already have the correct Python type and be automatically converted to the correct database type.

It's possible automatic type coercion makes this a non-issue, but it's better to check with some models that use `int` and `UUID` primary keys, that this actually works (starting from an empty index column).

src/onegov/search/indexer.py
```diff
@@ -392,17 +394,13 @@ def bulk_process(self):
...
-            tasks = {}
-
-            for _ in range(self.queue.qsize()):
-                task = self.queue.get(block=False, timeout=None)
-                if task['action'] == 'index':
-                    tasks.setdefault(task['tablename'], []).append(task)
-                self.queue.task_done()
+        def task_generator():
+            while not self.queue.empty():
+                yield self.queue.get(block=False, timeout=None)
```
Daverball, line 400: You accidentally got rid of `self.queue.task_done()`.

Tschuppi81: Absolutely!

src/onegov/search/indexer.py
Tschuppi81: Yes, unmodified works now

src/onegov/search/indexer.py
Daverball: I'm not sure if you're allowed to call `task_done` before pulling it off the queue, so I'd reorder these two statements, if you're worried about the final task of a group hanging around while it's being processed you can pull the task out into a local variable before yielding it:
```python
                task = self.queue.get(block=False, timeout=None)
                self.queue.task_done()
                yield task
```

Daverball:
@Tschuppi81 Ich habe gesehen es gibt teilweise Orte wo die automatische Verarbeitung von ORM events deaktiviert wird und `app.es_indexer.process()` manuell aufgerufen wird, z.B. beim FSI in der CLI. Da müsstest du dann allenfalls den postgres indexer auch noch manuell aufrufen. Mit `fgrep 'es_indexer.process' src -R` findest du ja schnell raus, wo du noch etwas anpassen musst.

src/onegov/search/integration.py
```diff
@@ -375,24 +408,64 @@ def reindex_model(model):
...
+                      f'being executed?)')
+                success = False
+            else:
+                print(f'Percentage of indexed documents of model '
+                      f'{model} is {percentage}%')
+
+        if not success:
+            print('\x1b[0;31;40m' + 'Indexing status check NOT OK' + '\x1b[0m')
+        else:
+            print('\x1b[0;32;40m' + 'Indexing status check OK' + '\x1b[0m')
```
Daverball, line 468: I would consider moving this entirely to the CLI command, so you can take advantage of `click.secho` for the styling and don't have to use `print`.

Daverball (approved):
This looks good to me now as a first version, I just have a small suggestion for the `psql_index_status` method.

Tschuppi81:
Testing single Town
onegov-core --select /onegov_town6/meggen upgrade
onegov-search --select /onegov_town6/meggen reindex
onegov-search --select /onegov_town6/meggen index-status

Testing all towns
onegov-core --select /onegov_town6/* upgrade
onegov-search --select /onegov_town6/* reindex
onegov-search --select /onegov_town6/* index-status

Testing single org

PR #1193, opened 2024-02-19
Fix type errors in Occasion

msom:
Resolves the following messages:
```
 src/onegov/activity/models/occasion.py:328: error: Unsupported operand types for >= ("int" and "None")  [operator]
src/onegov/activity/models/occasion.py:328: note: Right operand is of type "float | None"
src/onegov/activity/models/occasion.py:332: error: Unsupported operand types for - ("None" and "int")  [operator]
src/onegov/activity/models/occasion.py:332: note: Left operand is of type "float | None"
src/onegov/activity/models/occasion.py:351: error: Unsupported operand types for - ("None" and "int")  [operator]
src/onegov/activity/models/occasion.py:351: note: Left operand is of type "float | None"
src/onegov/activity/models/occasion.py:351: error: Incompatible return value type (got "float | int", expected "int")  [return-value]
src/onegov/activity/models/occasion.py:421: error: Argument "min_age" to "is_too_young" of "AgeBarrier" has incompatible type "float | None"; expected "int"  [arg-type]
src/onegov/activity/models/occasion.py:428: error: Unsupported operand types for - ("None" and "int")  [operator]
src/onegov/activity/models/occasion.py:428: note: Left operand is of type "float | None"
src/onegov/activity/models/occasion.py:428: error: Argument "max_age" to "is_too_old" of "AgeBarrier" has incompatible type "float | int"; expected "int"  [arg-type]
```

## Commit message

Feriennet: Fix type errors. 

TYPE: Bugfix

## Checklist

### DIFF
--- a/src/onegov/activity/models/occasion.py
+++ b/src/onegov/activity/models/occasion.py
@@ -6,7 +6,7 @@ from onegov.activity.models.occasion_date import DAYS
 from onegov.core.orm import Base, observes
 from onegov.core.orm.mixins import TimestampMixin
 from onegov.core.orm.types import UUID
-from psycopg2.extras import NumericRange
+from onegov.activity.types import BoundedIntegerRange
 from sqlalchemy import Boolean
 from sqlalchemy import case
 from sqlalchemy import Column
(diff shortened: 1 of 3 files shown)


msom:
@Daverball I am not sure about this.
- NumericRange might actually be None: https://github.com/jkehler/awslambda-psycopg2/blob/master/psycopg2-3.9/_range.py#L56, so defaulting it to zero might be OK? I guess we should do this in all the other cases (which are somehow not reported by mypy, e.g. https://github.com/OneGov/onegov-cloud/blob/a65d3fd5f8e68baa9a4034fa2399fbd3e1259f38/src/onegov/activity/models/occasion.py#L338), too?
- Type annotation `NumericRange` might be too broad, could we use a more specific annotation (INT4RANGE)?

Daverball:
I think I would prefer if we just disallow empty and unbounded ranges, since they don't really make any sense in this context (or do we allow occasions with unlimited spots?)

`INT4Range` isn't really valid as a value type, `NumericRange` is what SQLAlchemy will use under the hood as the actual value type with psycopg2, SQLAlchemy provides some functions to generate specific ranges but not actual types. I think this was remedied in 2.0, but we can't rely on this yet.

We may be able to define our own subclass that doesn't allow omitting start/end. If psycopg2 isn't happy with that we can pretend we have a subclass for type checking purposes like so:

```python
if TYPE_CHECKING:
    class BoundedNumericInterval(NumericInterval):
        def __init__(self, lower: float, upper: float, bounds: str = '[)') -> None: ...
        @property
        def upper(self) -> float: ...
        @property
        def lower(self) -> float: ...
else:
    def BoundedNumericInterval(lower: float, upper: float, bounds: str = '[)') -> NumericInterval:
        assert lower is not None and upper is not None
        return NumericInterval(lower, upper, bounds)
```

PR #1217, opened 2024-03-05
Delete invisible links before populating object, cli for deleting all…

BreathingFlesh:
… invisible links in topics

## Commit message

Org: Bugfix of deleting files

Some files in the FileLinkExtension could not be deleted, they will return to the list if they are still linked in the text. Some files were invisibly linked in the text. This change will delete all invisible links on save and adds a command for deleting all current invisible links. 

TYPE: Bugfix
LINK: PRO-1248

## Checklist


src/onegov/org/models/extensions.py
```diff
@@ -829,6 +830,7 @@ class GeneralFileLinkExtension(ContentExtension):
         id: Any
         files: relationship[list[File]]
         access: dict_property[str]
+        text: str
```
Daverball:
```suggestion
```
Take advantage of `content_fields_containing_links_to_files`, it tells you what you need to process, rather than hardcoding `text`.

So do something like:
```python
for field_name in obj.content_fields_containing_links_to_files:
    if text := obj.content.get(field_name):
        obj.content[field_name] = remove_empty_links(text)
```
And extract the parsing and removal of link tags into a utility function.

Maybe you could also first do a `self[field_name].object_data != self[field_name].data` to only do this step if the user changed the contents of the field.

src/onegov/org/cli.py
```diff
@@ -884,3 +886,54 @@ def mark_as_published(request: 'OrgRequest', app: 'OrgApp') -> None:
...
+@cli.command(name="delete-invisible-links")
+def delete_invisible_links() -> 'Callable[[OrgRequest, OrgApp], None]':
+    """ Deletes all the data associated with a period, including:
+    Example:
+        onegov-org --select /foo/bar delete-invisible-links
+    """
+
+    def delete_invisible_links(request: 'OrgRequest', app: 'OrgApp') -> None:
+        session = request.session
+        pages = session.query(Page).all()
```
Daverball, line 900: Technically you would need to do this for any model that uses the `GeneralFileLinkExtension`, not just `Page`, if you wanted to be thorough.

src/onegov/org/cli.py
```diff
@@ -884,3 +886,54 @@ def mark_as_published(request: 'OrgRequest', app: 'OrgApp') -> None:
...
+                        click.echo(f"Deleting invisible link: {link}")
+                        if all(tag.name == 'br' for tag in link.contents):
+                            link.replace_with(
+                                BeautifulSoup("<br/>", "html.parser")
+                            )
+                        else:
+                            link.decompose()
+
+                    # Save the modified HTML back to page.text
+                    page.text = str(soup)  # type:ignore
```
Daverball, line 929:
```suggestion
            if not isinstance(page, GeneralFileLinkExtension):
                continue
                
            for field_name in page.content_fields_containing_links_to_files:
                if text := page.content.get(field_name):
                   page.content[field_name] = remove_empty_links(text)
```
Mostly reuse the code from the Form in the Mixin and use a common utility function.

Daverball (commented):
A  few refactoring suggestions

src/onegov/org/models/extensions.py
```diff
@@ -872,13 +872,17 @@ def populate_obj(self, obj: 'GeneralFileLinkExtension',
...
                 for field_name in obj.content_fields_containing_links_to_files:
-                    if self[field_name].data == self[field_name].object_data:
-                        continue
-
-                    if (
-                        (text := obj.content.get(field_name))
-                        and (cleaned_text := remove_empty_links(text)) != text
-                    ):
-                        obj.content[field_name] = cleaned_text
+                    if hasattr(self, field_name):
```
Daverball, line 875:
```suggestion
                    if field_name in self:
```
That's the proper way to check for whether or not a field exists in WTForms.

src/onegov/org/cli.py
```diff
@@ -884,3 +890,65 @@ def mark_as_published(request: 'OrgRequest', app: 'OrgApp') -> None:
...
+        onegov-org --select /foo/bar delete-invisible-links
+    """
+
+    def delete_invisible_links(request: 'OrgRequest', app: 'OrgApp') -> None:
+        session = request.session
+        query = QueryChain(
+            (session.query(Page),
+             session.query(DaypassResource),
+             session.query(ItemResource),
+             session.query(RoomResource),
```
Daverball, line 908:
```suggestion
             session.query(Resource),
```
Just like with `Page` you can simplify this down to a single query using the common base class.

src/onegov/org/utils.py
```diff
@@ -62,7 +62,7 @@
 # note that this tag should be safe - i.e. it may not accept '<' or the like
 # to avoid js script injection further down the line
 HASHTAG = re.compile(r'#\w{3,}')
-IMG_URLS = re.compile(r'<img src="(.*?)"')
+IMG_URLS = re.compile(r'<img.*?src="(.*?)"')
```
Daverball, line 65:
```suggestion
IMG_URLS = re.compile(r'<img[^>]*?src="(.*?)"')
```
You can't allow all characters between img and src, but you can allow any except `>`.

BreathingFlesh: Right, thank you for noticing

src/onegov/org/cli.py
```diff
@@ -884,3 +889,63 @@ def mark_as_published(request: 'OrgRequest', app: 'OrgApp') -> None:
...
+                        click.echo(f"Deleting invisible link: {link}")
+                        if all(tag.name == 'br' for tag in link.contents):
+                            link.replace_with(
+                                BeautifulSoup("<br/>", "html.parser")
+                            )
+                        else:
+                            link.decompose()
+
+                    # Save the modified HTML back to page.text
+                    page.content[field] = str(soup)
```
Daverball, line 941: I'd replace this with a call to `remove_empty_links` as well and also add the check to see if the content changed before setting it, so you don't emit quite as many `UPDATE`statements, otherwise this looks good to me now.
```suggestion
                cleaned = remove_empty_links(page.content[field])
                if page.content[field] != cleaned:
                    page.content[field] = str(soup)
```

PR #1230, opened 2024-03-13
Use closure to capture free variables

cyrillkuettel:
## Commit message

Use closure to capture free variables

TYPE: Bugfix
LINK: OGC-1255

## Checklist

Tested and works. See linear for xml if needed.

### DIFF
--- a/src/onegov/town6/homepage_widgets.py
+++ b/src/onegov/town6/homepage_widgets.py
@@ -536,7 +536,7 @@ class JobsWidget:
                     'jobs_rss_feed',
                     creator=lambda: requests.get(rss_feed_url, timeout=4),
                     expiration_time=3600,
-                    should_cache_fn=self.should_cache_fn,
+                    should_cache_fn=lambda response: response.status_code == 200,
                 )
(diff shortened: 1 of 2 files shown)


Daverball (approved):

src/onegov/town6/homepage_widgets.py
```diff
@@ -520,49 +520,31 @@ class JobsWidget:
...
+            Note that this is called within the <?python> tag in the macro.
+            This is done so we can get the ``rss_feed_url`` which itself is a
+            dependency to build the actual widget.
+            """
+            try:
+                response = layout.app.cache.get_or_create(
+                    'jobs_rss_feed',
+                    creator=lambda: requests.get(rss_feed_url, timeout=4),
+                    expiration_time=3600,
+                    should_cache_fn=self.should_cache_fn,
```
Daverball, line 539:
```suggestion
                    should_cache_fn=lambda response: response.status_code == 200,
```
One minor nit, you can probably replace this with a lambda, we don't really need to make this a method either

PR #1256, opened 2024-03-25
Ogc 1429 i frames

BreathingFlesh:

## Commit message

Org: Add option for adding iFrames

TYPE: Feature
LINK: OGC-1429

## Checklist


tests/onegov/org/test_views_pages.py
```diff
@@ -576,3 +577,27 @@ def test_view_page_as_member(client):
...
+    page.form['url'] = "https://www.seantis.ch/success-stories/"
+    page = page.form.submit()
+    assert 'iframe' in page
+    assert 'https://www.seantis.ch/success-stories/' in page
+
+    page = client.get('/topics/organisation').click('iFrame')
+    page.form['title'] = "Failure"
+    page.form['url'] = "https://www.organisation.org/success-stories/"
+    page = page.form.submit()
+    assert 'Die Domäne der URL ist für iFrames nicht zulässig.' in page
```
msom, line 603: I would also test if the CSP headers are set correctly by the tween, see for example https://github.com/OneGov/onegov-cloud/blob/master/tests/onegov/election_day/views/test_views.py#L94

src/onegov/org/forms/page.py
```diff
@@ -61,6 +62,62 @@ class PageForm(PageBaseForm):
...
+        label=_("Display as card"),
+        description=_("Display the iframe as a card with a border"),
+        fieldset=_('Display')
+    )
+
+    def on_request(self) -> None:
+        self.allowed_domains = (
+            self.request.app.allowed_iframe_domains)  # type: ignore
+        if self.allowed_domains:
+            self.domain_hint.text = (
```
msom, line 102: There are hardcoded  domains such as youtube, vimeo etc.(https://github.com/OneGov/onegov-cloud/blob/master/src/onegov/org/app.py#L516). I guess it would make sense to mention these here as well.

You could even inspect the default content security policy here (`getattr(self.request.app.settings.content_security_policy.default, 'child_src', set())`).

msom (changes requested):
I would also mention the hard-coded domains and also check if the headers are set in the test. Other than that, looks good.

PR #1264, opened 2024-03-28
Bugfix/pro 1271 attribute error business name

Tschuppi81:
Pay: Fix attribute error 'business_name' for stripe

TYPE: Bugfix
LINK: pro-1271

### DIFF
--- a/src/onegov/pay/models/payment_providers/stripe.py
+++ b/src/onegov/pay/models/payment_providers/stripe.py
@@ -269,7 +269,9 @@ class StripeConnect(PaymentProvider[StripePayment]):
     def public_identity(self) -> str:
         account = self.account
         assert account is not None
-        if hasattr(account, 'business_name'):
+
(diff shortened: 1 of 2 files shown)


src/onegov/pay/models/payment_providers/stripe.py
```diff
@@ -269,7 +269,7 @@ def url(self) -> str:
     def public_identity(self) -> str:
         account = self.account
         assert account is not None
-        if account.business_name:
+        if 'business_name' in account:
```
msom, line 272: I would probably go for `getattr(account, 'business_name', '')` since it might exist but be empty...

Tschuppi81: done

msom:
With using has_attr you still have the case that business might exists but is empty. How about:
```
business_name = getattr(account, 'business_name', '')
if business_name:
  return f'{business_name} / {account.email}'
```

PR #1269, opened 2024-04-02
Web 42 i frames

BreathingFlesh:

## Commit message

Org: iFrame button generating iFrame-Code

TYPE: Feature
LINK: WEB-42

## Checklist

- [x] I have added new modules to the docs


src/onegov/org/models/traitinfo.py
```diff
@@ -124,7 +134,7 @@ def get_form_class(
     def get_editbar_links(
         self,
         request: 'OrgRequest'
-    ) -> 'Sequence[Link | LinkGroup]':
+    ) -> 'Sequence[Link | LinkGroup | iFrameLink]':
```
Daverball, line 137: You shouldn't need to change this union, since `iFrameLink` is a subclass of `Link`

src/onegov/org/models/traitinfo.py
```diff
@@ -158,7 +168,7 @@ def get_add_links(
     def get_edit_links(
         self,
         request: 'OrgRequest'
-    ) -> 'Iterator[Link | LinkGroup]':
+    ) -> 'Iterator[Link | LinkGroup | iFrameLink]':
```
Daverball, line 171: same here

src/onegov/org/elements.py
```diff
@@ -276,3 +276,50 @@ def __init__(
     'LinkGroup',
     'QrCodeLink'
 )
+
+
+class iFrameLink(BaseLink):
```
Daverball, line 281:
```suggestion
class IFrameLink(BaseLink):
```
Classes should always start with an uppercase letter, this is also more consistent with `IFrameForm`.

src/onegov/org/models/traitinfo.py
Daverball: I see the problem, it inherits from `BaseLink`. So you should change this to `BaseLink | BaseLinkGroup` or change `IFrameLink` and `QrCodeLink` to inherit from `Link` instead of `BaseLink`. I'm not entirely sure why we use different link classes in Org to begin with, part of it seems to be `access`, but we could make this work in the base class. I'm also not sure why we're using templates in one place and a markup builder in the other. Ideally we'd move all of that logic back into the base class.

Daverball: We're also not very consistent about which `Link` class we use, sometimes we use the base class and sometimes the Org class, regardless of whether the module is using an OrgApp or not.

src/onegov/org/models/traitinfo.py
```diff
@@ -134,7 +135,7 @@ def get_form_class(
     def get_editbar_links(
         self,
         request: 'OrgRequest'
-    ) -> 'Sequence[Link | LinkGroup | IFrameLink]':
+    ) -> 'Sequence[Link | LinkGroup | BaseLink]':
```
Daverball, line 138: Could you change this to `BaseLink | LinkGroup` in the other pull request after merging with master?

`Link` is redundant because it is a subclass of `BaseLink`. Same thing for `get_edit_links`.

PR #1275, opened 2024-04-09
Ogc 1564 timestamp video

BreathingFlesh:

## Commit message

Landsgemeinde: Convert Timestamp to seconds

YouTube Requires the timestamp as seconds. Added a utils-method to convert timestamps into seconds.

TYPE: Bugfix
LINK: OGC-1563

## Checklist


src/onegov/org/utils.py
```diff
@@ -1185,3 +1185,23 @@ def widest_access(*accesses: str) -> str:
...
+
+
+def timestamp_to_seconds(t: str) -> int:
+    """Convert a timestamp to seconds.
+       Examples:
+       '1m30s' -> 90
+       '30s' -> 30
+       '1h2m30s' -> 3750"""
+
+    time = 0
```
Tschuppi81, line 1197: Minor: rename `time` to `seconds`

Tschuppi81 (approved):

src/onegov/org/utils.py
```diff
@@ -1185,3 +1185,23 @@ def widest_access(*accesses: str) -> str:
...
+    """Convert a timestamp to seconds.
+       Examples:
+       '1m30s' -> 90
+       '30s' -> 30
+       '1h2m30s' -> 3750"""
+
+    seconds = 0
+    if 'h' in t:
+        h, t = t.split('h')
+        seconds += int(h) * 3600
```
msom, line 1200: Since this timestamp comes from a user input, we should probably catch errors coming from misspelling. Maybe one try/catch with defaulting to 0?

msom: BTW wouldn't this be much easier with regex? You could also add a validator to the form using the same regex

msom: Something like ` r'(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s)?'`

src/onegov/landsgemeinde/models/votum.py
```diff
@@ -129,8 +130,9 @@ def agenda_item_number(self) -> int:
     @property
     def video_url(self) -> str | None:
         video_url = self.agenda_item.assembly.video_url
+        ts = timestamp_to_seconds(str(self.video_timestamp))
```
msom, line 133: this will lead to an error is video_timestamp is None ('None')

msom:
`{self.video_url_base}&amp;start={seconds}` makes no sense. I guess start would be a query param, so `&` should not be quoted but the URL needs to be parsed properly to ensure a `?` is present...

msom (approved):

PR #1278, opened 2024-04-11
Adds test reproducing issue

Tschuppi81:
Org: Fix adjusting registration window (end date) after first attendee confirmed

TYPE: Bugfix
LINK: ogc-1557

### DIFF
--- a/src/onegov/org/forms/form_registration.py
+++ b/src/onegov/org/forms/form_registration.py
@@ -206,7 +206,7 @@ class FormRegistrationWindowForm(Form):
 
         assert self.limit.data is not None  # but may be 0 / limit inactive
 
-        if self.limit.data and self.limit.data < self.claimed_spots:
+        if 0 < self.limit.data < self.claimed_spots:
             raise ValidationError(_(
                 "The limit cannot be lower than the already confirmed "
                 "number of attendees (${claimed_spots})",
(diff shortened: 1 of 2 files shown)


Tschuppi81:
Adding @msom as @BreathingFlesh is out

src/onegov/org/forms/form_registration.py
```diff
@@ -204,8 +204,9 @@ def validate_limit(self, field: IntegerField) -> None:
         if not self.limit_attendees.data:
             return
 
-        assert self.limit.data is not None
-        if self.limit.data < self.claimed_spots:
+        assert self.limit.data is not None  # but may be 0 / limit inactive
```
msom, line 207: Is this only for typing or does this (being 0) actually occur? Then we should probably also raise a ValidationErro

msom (approved):

src/onegov/org/forms/form_registration.py
Tschuppi81: Yes it can happened, https://linear.app/onegovcloud/issue/OGC-1557.

Tschuppi81: Actually it cannot happen from the user side, it could happen if a programmer does a mistake as the value will be set to 0 or the limit provided in the form during `populate_obj`
https://github.com/OneGov/onegov-cloud/blob/master/src/onegov/org/forms/form_registration.py#L131

So I guess I will keep the assert statement not switching to a ValidationError

PR #1292, opened 2024-04-15
News are now deletable when expired

Tschuppi81:
News: News are now deletable when end date passed

TYPE: Bugfix
LINK: ogc-1560


Tschuppi81:
Missed to add tests... will do

BreathingFlesh (approved):
Maybe move the deletable option near the publication in the form.

PR #1293, opened 2024-04-15
Make an occurrence deletable after reaching end date

Tschuppi81:
Event: Occurrences are now deletable when end date passed

TYPE: Feature
LINK: ogc-1560

### DIFF
--- a/src/onegov/event/models/occurrence.py
+++ b/src/onegov/event/models/occurrence.py
@@ -6,6 +6,7 @@ from onegov.core.orm.mixins import TimestampMixin
 from onegov.core.orm.types import UUID
 from onegov.event.models.mixins import OccurrenceMixin
 from onegov.gis import Coordinates
+from onegov.org.models.extensions import DeletableContentExtension
 from pytz import UTC
 from sedate import to_timezone
 from sqlalchemy import Column
@@ -20,7 +21,8 @@ if TYPE_CHECKING:
     from onegov.event.models import Event
 
 
-class Occurrence(Base, OccurrenceMixin, TimestampMixin):
+class Occurrence(Base, OccurrenceMixin, TimestampMixin,
+                 DeletableContentExtension):
     """ Defines an occurrence of an event. """
 
     __tablename__ = 'event_occurrences'


Tschuppi81:
My goal is to make an extension to the Occurrence model in order to make it 'deletable' when the end date has passed. Now, when importing the extension I get a 'circular import'. @msom Any idea how to resolve it?

msom:
> My goal is to make an extension to the Occurrence model in order to make it 'deletable' when the end date has passed. Now, when importing the extension I get a 'circular import'. @msom Any idea how to resolve it?
> 
> Btw: Did the same for `News`, see #1292

Just import it in the function (search for `# XXX circular import` comments for an example)

msom:
> My goal is to make an extension to the Occurrence model in order to make it 'deletable' when the end date has passed. Now, when importing the extension I get a 'circular import'. @msom Any idea how to resolve it?
> 
> Btw: Did the same for `News`, see #1292

I think the occurence is the wrong place for this, as they get generated by rrule from the event. You probalby want to extend the event and use this in the occurrence. Also make sure `occurrence_dates` does not generate past occurrences.

Another point is that events are most likely defined by submission and always public, so does this setting on a single event really make sense? Why not make it a general setting in the event configuration (/event-settings) and use it for all events on an instance? But if not, you probably need to adjust the `EventForm` by hand while also making sure, it's only available for admins/editors and not when submitting an event as visitor i guess...

src/onegov/org/cronjobs.py
```diff
@@ -687,24 +691,25 @@ def delete_content_marked_deletable(request: 'OrgRequest') -> None:
...
             query = query.filter(model.delete_when_expired == True)
             for obj in query:
-                deletable = False
-
                 # delete entry if end date passed
                 if isinstance(obj, (News, ExtendedDirectoryEntry)):
-                    if obj.publication_end and obj.publication_end < utc_now:
-                        deletable = True
-                    else:
-                        if (not obj.publication_end
```
msom, line 697: Why did you delete this code?

msom (approved):

PR #1312, opened 2024-04-30
Ogc 1596 edit bar

BreathingFlesh:

## Commit message

Town6: Move the save button for edit forms to edit bar

TYPE: Feature
LINK: OGC-1596

## Checklist


src/onegov/core/elements.py
```diff
@@ -173,6 +173,35 @@ def __repr__(self) -> str:
...
+    id = 'back_link'
+
+    __slots__ = ()
+
+    def __init__(
+        self,
+        text: str = '',
+        **props: Any
+    ):
+        super().__init__(text, 'javascript:history.back()', **props)
```
Daverball, line 199: We want to get rid of inline JavaScript to tighten up our CSP, so I would move this function into a common JavaScript file and globally attach it to a selector e.g. `a.go-back-link`. Or you can also use a data attribute `a[data-back-link]`

src/onegov/org/layout.py
Daverball, line 1: The extra union member is not ideal, maybe add
```python
    from onegov.core.elements Link as BaseLink
```
in the `if TYPE_CHECKING` block and change the union to `'BaseLink | LinkGroup'`

src/onegov/town6/locale/de_CH/LC_MESSAGES/onegov.town6.po
```diff
@@ -328,6 +331,9 @@ msgstr "Art:"
 msgid "The chat request got accepted."
 msgstr "Die Chat-Anfrage wurde angenommen."
 
+msgid "<dynamic element> <dynamic element>"
```
Daverball, line 334: This indicates that you either accidentally put a `i18:translate` somewhere or forgot `i18:name` for a replacement variable.

src/onegov/town6/templates/macros.pt
```diff
@@ -389,23 +389,25 @@
...
-            <input type="submit" value="${button_text}" class="button" role="button">
+        <tal:b condition="not: layout.edit_mode|False">
+            <tal:b condition="cancel|nothing">
+                <a class="button secondary cancel-button" href="${cancel}" i18n:translate="" role="button">
+                    <tal:b condition="not:cancel_text|nothing">
+                        Cancel
+                    </tal:b>
+                    <tal:b condition="cancel_text|nothing">
+                        ${cancel_text}
+                    </tal:b>
```
Daverball, line 400: Looks like it was here:
```suggestion
                <a class="button secondary cancel-button" href="${cancel}" role="button">
                    <tal:b condition="not:cancel_text|nothing" i18n:translate="">
                        Cancel
                    </tal:b>
                    <tal:b condition="cancel_text|nothing">
                        ${cancel_text}
                    </tal:b>
```

Daverball (changes requested):
Looks good for the most part. There is a CSP issue that we should probably resolve now, while we have the opportunity, but we can also leave it for a follow-up pull request if you'd prefer that.

Other than there's a misplaced `i18n:translate` in one of the templates and a minor typing nitpick.

PR #1326, opened 2024-05-16
Update submmission title

Tschuppi81:
Submission: Update submission title after editing

TYPE: Bugfix
LINK: ogc-1576

### DIFF
--- a/src/onegov/form/models/submission.py
+++ b/src/onegov/form/models/submission.py
@@ -233,6 +233,7 @@ class FormSubmission(Base, TimestampMixin, Payable, AssociatedFiles,
     def update_title(self, form: 'Form') -> None:
         title_fields = form.title_fields
         if title_fields:
+            # FIXME: Reconsider using unescape when consistently using Markup.
             self.title = extract_text_from_html(', '.join(
                 html.unescape(render_field(form._fields[id]))
                 for id in title_fields
--- a/src/onegov/form/models/submission.py
+++ b/src/onegov/form/models/submission.py
@@ -32,7 +32,6 @@ if TYPE_CHECKING:
     from onegov.form import Form
     from onegov.form.models import FormDefinition, FormRegistrationWindow
     from onegov.form.types import RegistrationState, SubmissionState
-    # from onegov.form.utils import extract_text_from_html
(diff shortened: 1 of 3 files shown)


src/onegov/form/models/submission.py
```diff
@@ -31,6 +32,7 @@
     from onegov.form import Form
     from onegov.form.models import FormDefinition, FormRegistrationWindow
     from onegov.form.types import RegistrationState, SubmissionState
+    # from onegov.form.utils import extract_text_from_html
```
Daverball, line 35:
```suggestion
```
leftover comment

Daverball (approved):
Just a comment that needs to be cleaned up.

You could add a test to make sure `html.unescape` works by putting one of `<`, `>`, `&`, `"` in one of the name fields, but I don't think it's necessary.

tests/onegov/form/test_collection.py
```diff
@@ -101,6 +101,35 @@ def test_submission_extra_data(session):
...
+    submitted_form = form.form_class(data)
+    submission = collection.submissions.add(
+        'whatever', submitted_form, state='complete')
+    assert submission.title == 'Bill, Gate.'
+
+    edited_form = submitted_form
+    edited_form.first_name.data = '<Billy>Bill'
+    edited_form.last_name.data = 'Gates'
+    collection.submissions.update(submission, edited_form)
+    assert submission.title == 'Bill, Gates'
```
Daverball, line 130: Technically `<Billy>` being stripped from the title is a bug, since it will be rendered as `&lt;Billy&gt;` and only unescape makes it look like a tag, but it would be fairly difficult to avoid it. So it's probably not worth fixing.

I'm frankly not quite sure why we unescape here anyways, unless there are some fields that are allowed to contain Markup.

Daverball: It's possible that we render the title with `structure` for whatever reason, so you can leave it like it for now but I'd make a `FIXME` comment to remind me to investigate this when I look at consistent use of `Markup`.

Tschuppi81: Agree, I played/tried quite a bit but also do not fully understand why its done this way.
I will revert the last commit and leave this alone...

PR #1328, opened 2024-05-17
Fixme Makes the xml export generic

Tschuppi81:
Export: Implement a generic xml export

TYPE: Bugfix
LINK: None | Fixme

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
     - id: sass-lint
       files: '^src/.*\.scss'
   - repo: https://github.com/pre-commit/mirrors-eslint
-    rev: v9.1.1
+    rev: v9.2.0
     hooks:
     - id: eslint
       files: '^src/.*\.jsx?$'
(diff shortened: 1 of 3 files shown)


Tschuppi81:
Replaces the obscure occurrence xml export (I did earlier) with a generic implementation. Tested for event export and newsletter recipients

Tschuppi81:
@Daverball Do I need to implement stubs for 'dicttoxml' to fix the mypy errors?

Daverball:
It's probably good enough to just add `type:ignore[import-untyped]` to the import for now. Not sure whether it's worth adding stubs for this one function.

Daverball (approved):
Since you wrote the original XML export I assume the original use-case is still supported after this change and provides equivalent functionality.

Tschuppi81:
> Since you wrote the original XML export I assume the original use-case is still supported after this change and provides equivalent functionality.

Yes, it was for the event export :+1:

Tschuppi81:
grrr, mis-clicked resulting in a bad commit message...

PR #1331, opened 2024-05-21
Bugfix/fix psql index status

Tschuppi81:
Search: Fix search cli for index status

TYPE: Bugfix
LINK: None

### DIFF
--- a/src/onegov/search/cli.py
+++ b/src/onegov/search/cli.py
@@ -4,10 +4,10 @@ import click
 
 from onegov.core.cli import command_group, pass_group_context
 from sedate import utcnow
-from sqlalchemy import func
+from sqlalchemy import func, inspect, Column
 
+from typing import TYPE_CHECKING, Any
 
-from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


src/onegov/search/cli.py
```diff
@@ -39,7 +53,9 @@ def psql_index_status(app: 'Framework') -> None:
             success = 3
             continue
 
-        q = session.query(func.count(model.fts_idx))
+        count_column = get_non_nullable_columns(model)[0]
+        q = session.query(func.count(count_column))
```
Daverball, line 57: Have you tried:

```suggestion
        q = session.query(model).with_entities(func.count(text('*'))
```

That seems a lot more robust to me than inspecting the table, that should generate `SELECT count(*) FROM ...`

Daverball (commented):
I'd try to get rid of the introspection overhead, since it doesn't seem necessary.

src/onegov/search/cli.py
Tschuppi81: Will do so

Tschuppi81: Works as expected, just commited

Daverball (approved):

PR #1340, opened 2024-05-28
Feature/ogc 1634 use directories for faq

Tschuppi81:
Directories: Adds FAQ directories for common questions and answers

TYPE: Feature
LINK: ogc-1634

## Checklist


src/onegov/org/forms/directory.py
```diff
@@ -412,6 +415,10 @@ def ensure_public_fields_for_submissions(self) -> bool | None:
         are not meant to be private repositories.
 
         """
+        # Skip check for FAQ directories
+        if hasattr(self, 'is_faq') and self.is_faq:
```
Daverball, line 419:
```suggestion
        if getattr(self, 'is_faq', False):
```

src/onegov/org/forms/directory.py
```diff
@@ -757,3 +764,186 @@ def validate_name(self, field: StringField) -> None:
             raise ValidationError(
                 _("An entry with the same name exists")
             )
+
+
+class DirectoryFAQForm(DirectoryBaseForm):
```
Daverball, line 769: I think it would be better to add an `is_faq` boolean field at the top and add field dependencies to hide the fields that don't matter for FAQs. We can still have two dedicated views by pre-setting and hiding the field if we wish, although I'm not sure even that is worth it, considering the separate views are just all copies of one another.

src/onegov/org/views/directory.py
```diff
@@ -57,6 +57,9 @@ def get_directory_form_class(
     model: object,
     request: 'OrgRequest'
 ) -> type[DirectoryForm]:
+    if 'faq' in request.path_info:
```
Daverball, line 60: This is very fragile, if we collapse the two forms into one we can get rid of this.

Daverball (changes requested):
I would prefer to not split this into two logical units and instead make use of field dependencies.

src/onegov/org/forms/directory.py
Tschuppi81: Yes, let me do so

src/onegov/org/views/directory.py
Tschuppi81: I tried hard to implement the field dependencies but I didn't really manage :-( 
1) It seems the `DirectoryFAQForm` will be shown for regular directories as well as for faq directories
2) I didn't find a good way to change the form values in case of faq directories

src/onegov/org/forms/directory.py
```diff
@@ -334,9 +375,24 @@ class DirectoryBaseForm(Form):
             ('submitter_address', _("Address")),
             ('submitter_phone', _("Phone")),
         ),
-        fieldset=_("Submitter")
+        fieldset=_("Submitter"),
+        depends_on=('is_faq', 'n'),
     )
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
Tschuppi81, line 382: I don't like the ctor approach as users can not adapt the `structure` of their faq directories as the ctor will overwrite the values upon submit.

Daverball: You could set `field.default` in `on_request` instead. I would avoid overwriting `__init__` since we have plenty of other ways to hook control flow.

src/onegov/org/views/directory.py
Daverball: 1. If you don't actually make the form dynamic, i.e. you have separate views it might be better to just use `on_request` and delete all the fields/fieldsets that should be hidden. Also don't use a field for `is_faq` if you go that route, just set/read it in the view. I personally think it would be better to just keep things simple and keep doing it all in one view. We don't gain that much from pre-populating two fields with a suggestion and if we really want to have that, we could still do it with JavaScript.
2. Don't change form values if you can avoid it, you can dynamically set a default, but I'd avoid setting data, it gets messy and is not really intended by WTForms.
3. Seems like a bug, although it will be a bit tricky to fix, since we will have to check in JavaScript whether the corresponding fieldset is empty after hiding a field and then hide the section. And do the reverse when a field is unhidden.

Tschuppi81: Regarding point 1: I would prefer to have one view only but not sure how to accomplish this to distinguish e.g. 'new directory' from 'new faq directory'

Daverball: My question is, do we even need to make this as user-friendly? The ones setting up new directories are usually us. I'd honestly keep this as dead-simple as possible and add a checkbox "Allow anonymous submissions" which turns off the requirement of having an E-Mail field for user submissions.

As long as it's easy to submit new questions and add answers for those questions that's honestly good enough in my book. There may be other use cases where anonymous submissions make sense.

Daverball: And the display of individual entries as an accordion can be another checkbox.

Daverball: Or better yet just a generic radio select called `layout` with the two options, in case we want to add more layouts in the future.

Tschuppi81: > My question is, do we even need to make this as user-friendly? The ones setting up new directories are usually us and this only needs to be done once per tenant. I'd honestly keep this as dead-simple as possible and add a checkbox "Allow anonymous submissions" which turns off the requirement of having an E-Mail field for user submissions.
> 
> As long as it's easy to submit new questions and add answers for those questions that's honestly good enough in my book. There may be other use cases where anonymous submissions make sense.

The goal is to have Q&A lists provided by the community. Personally I would not allow anyone submitting a new question, I would rather collect the most frequent one and publish them...

Daverball: The way I read the original issue the only thing they really care about is having a different layout anyways. I misread some of your code, that's where the anonymous submissions bit came from.

Daverball: I'd pretty much revert all of the changes to the form and views, the only thing I'd keep is the changes to the template and SCSS, and change the condition for the accordion layout to `directory.layout == 'accordion'` and add one radio field for layout to the directory form, I don't think we need any other changes. It will be much easier to maintain.

Point in case: the changes were so complicated I couldn't really figure out what the goal of the change was without reading the original ticket.

You also haven't ported the changes to the Org Template and SCSS yet, so the setting would not do anything in Org, which I'd give a higher priority than making it easy to configure.

Tschuppi81: Right, not yet ported to org. 
My goal was to prepare the `form.structure` so anybody could start an Q&A catalog/directory ;-) But let me try your suggested layout setting.

Tschuppi81:
Much better with less changes, I like it!
Open items: template for org, tests

Daverball (approved):
This looks good now, so just needs porting to Org.

PR #1344, opened 2024-06-02
Use explicit ubuntu-22.04 instead of ubuntu-latest image

msom:


msom:
Already in the master

PR #1357, opened 2024-06-05
Feature/ogc 508 replace elastic search by postgres v2

Tschuppi81:

## Commit message

<Module>: <Message>

<Optional Description>

TYPE: <Feature|Bugfix>
LINK: <Ticket-Number>
HINT: <Optional upgrade hint>

## Checklist


src/onegov/org/models/ticket.py
```diff
@@ -67,6 +71,9 @@ def reference(self, request: 'OrgRequest') -> str:
     def reference_group(self, request: 'OrgRequest') -> str:
         return request.translate(self.group)
 
+    # FIXME: extra localized text is used for suggestions in the search but
+    # does not work with the ranking of the search results
+    # @hybrid_property
     @cached_property
     def extra_localized_text(self) -> str:
```
Tschuppi81, line 78: `extra_localized_text` is used for search suggestions but do not work when applying the ranking for search results. I changed to `hybrid_property` which throws
```
  File "/home/reto/workspace/onegov-cloud/src/onegov/org/models/ticket.py", line 93, in extra_localized_text
    q = object_session(self).query(Message)
        ^^^^^^^^^^^^^^^^^^^^
  File "/home/reto/workspace/onegov-cloud/venv/lib/python3.11/site-packages/sqlalchemy/orm/session.py", line 3464, in object_session
    util.raise_(
  File "/home/reto/workspace/onegov-cloud/venv/lib/python3.11/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
sqlalchemy.orm.exc.UnmappedInstanceError: Class 'sqlalchemy.ext.declarative.api.DeclarativeMeta' is not mapped; was a class (onegov.org.models.ticket.FormSubmissionTicket) supplied where an instance was required?
```
How can I transform it to a hybrid property?

src/onegov/org/models/search.py
```diff
@@ -159,5 +163,207 @@ def subset_count(self) -> int:
...
+                            vector,
+                            ts_query,
+                            0  # normalization, ignore document length
+                        )
+                        query = query.filter(
+                            model.fts_idx.op('@@')(ts_query)
+                        )
+                        # FIXME: sorting by rank_expression does not work
+                        # with OrgTicketMixin::extra_localized_text
+                        # query = query.order_by(rank_expression.desc())
```
Tschuppi81, line 314: This is the counter part of the issue described in the above comment.

Tschuppi81:
@Daverball I am working on the tests in `test_views_search.py` and can't make them pass. I narrowed the issue down to the hybrid_property resp. expression of `GeneralFile` and `News`.
While debugging for `News` I see that `es_public` is `True` for all the three elements in the DB. Same for `access` and `published`. Once I apply the `query.filter(model.es_public == True)` no elements are left.
Screenshot from 2024-10-14 06-29-53
Any idea why my filter statement or my hybrid expression does not work?

Daverball:
> @Daverball I am working on the tests in `test_views_search.py` and can't make them pass. I narrowed the issue down to the hybrid_property resp. expression of `GeneralFile` and `News`. While debugging for `News` I see that `es_public` is `True` for all the three elements in the DB. Same for `access` and `published`. Once I apply the `query.filter(model.es_public == True)` no elements are left. Screenshot from 2024-10-14 06-29-53 Any idea why my filter statement or my hybrid expression does not work?

You're relying a lot on implicit type conversions here and hoping that SQLAlchemy is doing the right thing for you. You could try `.is_(True)` rather than `== True`, which is the recommended way for boolean and none checks in SQLAlchemy anyways. Although you can probably just filter on the column itself, since a boolean column can be used as a filter so `filter(model.es_public)` should technically work as well.

Either way to properly debug this, you would have to look at the final SQL statement SQLAlchemy generates for you. Sometimes you will just have to manually define an `@{property_name}.expression` for hybrid properties to get something correct. There's already several hybrid properties that define a manual expression, like `published`.

Daverball:
Also as a side note `query.count` is quite inefficient for checking whether or not you have any results, since Postgres will have to find and count all the entries. It's usually better to do `if session.query(query.exists()).scalar():` which will create an `EXISTS` subquery, which only has to find a single row that satisfies the predicate before returning a result.

src/onegov/org/models/search.py
```diff
@@ -323,7 +323,9 @@ def generic_search(self) -> list['Searchable']:
                         query = (query.filter(model.fts_idx.op('@@')(ts_query))
                                  .add_columns(rank_expression))
 
-                        results.extend(list(query.all()))
+                        res = list(query.all())
```
Daverball, line 325: `query.all()` already returns a list so you don't need to convert to a `list`.

tests/onegov/org/test_views_search.py
```diff
@@ -147,13 +188,82 @@ def test_search_recently_published_object(client_with_es):
...
+    assert 'is pretty awesome' not in anom.get('/search-postgres?q=fulltext')
+
+    # postgres
+    assert 'fulltext' in client.get('/search-postgres?q=wow')
+    assert 'fulltext' not in anom.get('/search-postgres?q=wow')
+    assert 'is pretty awesome' in client.get('/search-postgres?q=fulltext')
+    assert 'is pretty awesome' not in anom.get('/search-postgres?q=fulltext')
+
+
+def test_search_recently_published_object_freeze(client_with_es):
```
Tschuppi81, line 213: I don't understand why test_search_recently_published_object and tests/onegov/org/test_views_search.py::test_search_recently_published_object_freeze are failing. I introduced the later just to test the same in a different way but it fails as well.
Doing the steps on the UI, the News are only listed when they should, as long as it is within the publication window or after publication start date/time (in case of no end date).
My assumption is that the sql expressions for es_public are not a fully correct causing this issue but couldn't figure it out.
Don't forget to use search-postgres for search queries to use postgres

tests/onegov/agency/test_views.py
```diff
@@ -765,8 +767,26 @@ def test_basic_search(client_with_es):
     assert 'Nick Rivera (Doctor)' in client.get('/search/suggest?q=Nic').json
     assert 'Rivera Nick (Doctor)' in client.get('/search/suggest?q=Riv').json
     assert '8911 Rivera Nick (Doctor)' in client.get(
-        '/search/suggest?q=89'
-    ).json
+        '/search/suggest?q=89').json
+
+    # postgres
```
Tschuppi81, line 772: The same is valid for this test. Any ideas @Daverball?

src/onegov/fsi/models/course_attendee.py
```diff
@@ -152,7 +166,7 @@ def role(self) -> str:
         assert self.user is not None
         return self.user.role
 
-    @property
+    @hybrid_property
```
Tschuppi81, line 169: Since I converted `email` to a `hybrid_property` the test `test_add_edit_external_attendee` fails with `TypeError: 'email' is an invalid keyword argument for CourseAttendee`.
If I revert back the test passed but the `search-postgres?q=martin` fails due to the missing hybrid property...

Daverball: It's very possible that `setter` has to be defined before the `expression`. Although I'm still not sure if the auto-generated `__init__` for SQLAlchemy will contain settable hybrid properties, it might not.

But this whole thing seems to be heading in very much the same direction as your first attempt at implementing the indexing. Maybe it would be more robust to store the computed es properties like `es_public` in the database at the same time we store the index.

This way we don't have to add all these new potentially subtly incorrect hybrid properties just to use them within a SQL query and are closer to the original elastic search implementation. If we want more of these things to happen online down the road, we can always improve it later.

tests/onegov/org/test_views_search.py
Daverball: I believe there are some cronjobs that have to run in order for recently published or unpublished models to behave correctly, although I can't recall if that is something elastic search specific or if postgres fts would also be affected by this cronjob.

So only freezing the time might not be enough, you may also need to invoke the cronjob. However, it is very likely that some of the hybrid properties are incorrect. But it will be difficult to test them all and find the culprit(s). It might be more fruitful to try a different approach, as described in my other comment.

Tschuppi81: The cronjob only triggers a re-indexing of the object (meant for ES) but does not impact the postgres data. So no, this is not helping...

Tschuppi81:
I will try the approach of calculating and storing the properties in a separate column while indexing and use them for searching.

PR #1364, opened 2024-06-10
Bugfix/pro 1289 notification template allow links

Tschuppi81:
Feriennet: Allow links in notification template

TYPE: Bugfix
LINK: pro-1289

### DIFF
--- a/src/onegov/feriennet/models/notification_template.py
+++ b/src/onegov/feriennet/models/notification_template.py
@@ -155,7 +155,7 @@ class TemplateVariables:
         if len(paragraphs) <= 1:
             result = text
         else:
-            result = '\n'.join(as_paragraphs(text))
+            result = '\n'.join(paragraphs)
 
         result = self.expand_storage_links(Markup(result))
 
--- a/src/onegov/feriennet/models/notification_template.py
+++ b/src/onegov/feriennet/models/notification_template.py
@@ -1,6 +1,7 @@
 import re
 
-from markupsafe import Markup, escape
+from bleach import clean
+from markupsafe import Markup
 from onegov.activity import BookingCollection, InvoiceCollection
 from onegov.core.orm import Base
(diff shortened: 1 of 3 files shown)


src/onegov/feriennet/models/notification_template.py
```diff
@@ -145,9 +155,16 @@ def render(self, text: str) -> Markup:
         if len(paragraphs) <= 1:
             result = text
         else:
-            result = Markup('\n').join(as_paragraphs(text))
+            result = '\n'.join(paragraphs)
 
-        return self.expand_storage_links(result)
+        result = self.expand_storage_links(Markup(result))
+
+        return clean(
```
Tschuppi81, line 162: I thought about using `linkify` from [core/utils.py](https://github.com/OneGov/onegov-cloud/blob/master/src/onegov/core/utils.py#L406) but it does not allow <p> and <br>. I could easily add them but not sure if its a good idea.

src/onegov/feriennet/models/notification_template.py
```diff
@@ -134,8 +119,33 @@ def bind(
...
+    def render(self, text: str) -> str:
+        """
+        Escapes and renders the given text, replacing all tokens with the
+        result of the bound methods and 'linkifying' the result.
+
+        :param text:
+        :return: clean text
+        """
+
+        def as_paragraphs(text: str) -> Iterator[str]:
```
Tschuppi81, line 131: NOTE: I moved `as_paragraphs` inside `render` method as I reverted using `Markup` and `as_paragraphs` does not return safe html code. I finally use `bleach.clean` to make it safe.

Daverball: I'd go back to the safe `as_paragraphs` function, there's still plenty of ways to add links this way.

Tschuppi81:
New solution found, closing

PR #1369, opened 2024-06-11
Improves CI performance by using pytest-xdist

Daverball:
Since #1366 should have potentially fixed the deadlocks caused by forks, this might work now

## Checklist


Daverball:
Looks like we'll first need to clean up the tests that aren't properly isolated from one another, but this is at least a good way to identify some of those tests

Daverball:
Actually, it looks like the observers don't always trigger correctly with pytest-xdist involved.

PR #1372, opened 2024-06-11
Bugfix/pro 1289 html field for  notification template

Tschuppi81:
Feriennet: Switch to html for notification templates

TYPE: Bugfix
LINK: pro-1289

### DIFF
--- a/src/onegov/feriennet/forms/notification_template.py
+++ b/src/onegov/feriennet/forms/notification_template.py
@@ -1,5 +1,6 @@
 from functools import cached_property
-from markupsafe import escape, Markup
+from markupsafe import Markup, escape
+
 from onegov.activity import Activity
 from onegov.activity import Booking, BookingCollection
 from onegov.activity import Occasion, OccasionCollection
@@ -53,6 +54,14 @@ class NotificationTemplateForm(Form):
             raise ValidationError(
                 _("A notification with this subject exists already")
             )
+
+    def process_obj(self, model: NotificationTemplate) -> None:
+        self.subject.data = Markup.escape(model.subject)
+        self.text.data = Markup.escape(model.text)
+
+    def populate_obj(self, model):
+        model.subject = Markup.unescape(self.subject.data)
+        model.text = Markup.unescape(self.text.data)
 
 
 class NotificationTemplateSendForm(Form):
--- a/src/onegov/feriennet/forms/notification_template.py
+++ b/src/onegov/feriennet/forms/notification_template.py
@@ -55,11 +55,15 @@ class NotificationTemplateForm(Form):
                 _("A notification with this subject exists already")
             )
 
-    def process_obj(self, model: NotificationTemplate) -> None:
+    def process_obj(
+        self,
+        model: NotificationTemplate  # type:ignore[override]
+    ) -> None:
+
         self.subject.data = Markup.escape(model.subject)
         self.text.data = Markup.escape(model.text)
 
-    def populate_obj(self, model):
+    def populate_obj(self, model: NotificationTemplate) -> None:
         model.subject = Markup.unescape(self.subject.data)
         model.text = Markup.unescape(self.text.data)
 
--- a/src/onegov/feriennet/forms/notification_template.py
+++ b/src/onegov/feriennet/forms/notification_template.py
@@ -10,11 +10,11 @@ from onegov.feriennet.collections import BillingCollection
 from onegov.feriennet.layout import DefaultLayout
 from onegov.feriennet.models import NotificationTemplate
 from onegov.form import Form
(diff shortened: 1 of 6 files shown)


src/onegov/feriennet/models/notification_template.py
```diff
@@ -135,19 +118,15 @@ def bind(
         self.bound[token] = method
 
     def render(self, text: str) -> Markup:
-        text = escape(text)
```
Daverball, line 138: This still needs to happen, we also use `render` for the subject. You will need to change the notification template views to wrap `self.text` in `Markup`.

Ideally we'd add a new database type that automatically creates `Markup` instead of `str` for fields that contain `Markup`. But we can do that in a separate step, we would also have to figure out how to get the same functionally in `data_property`.

src/onegov/feriennet/forms/notification_template.py
```diff
@@ -54,6 +55,18 @@ def validate_subject(self, field: StringField) -> None:
...
+        model: NotificationTemplate  # type:ignore[override]
+    ) -> None:
+
+        self.subject.data = Markup.escape(model.subject)
+        self.text.data = Markup.escape(model.text)
+
+    def populate_obj(self, model: NotificationTemplate) -> None:
+        model.subject = Markup.unescape(self.subject.data)
+        model.text = Markup.unescape(self.text.data)
+
```
Daverball, line 69:
```suggestion
```
This change makes no sense to me. `def render` is still the correct place for the `escape`. You just need to pass `Markup(self.text)` instead of `self.text` in the view that calls the `render` function. Eventually we may want to add a `Markup` column type, so we don't need to do this conversion for columns, that are supposed to store HTML content.

Tschuppi81: Ok, done. But now getting again `Markup is used in a dangerous way`

Daverball: That's a flake8 plugin and it's correct that it would be dangerous, you can add a `noqa: MS001` comment to silence the error.

src/onegov/feriennet/views/notification_template.py
```diff
@@ -230,8 +232,8 @@ def email_iter() -> 'Iterator[EmailJsonDict]':
         'title': _("Mailing"),
         'layout': layout,
         'form': form,
-        'preview_subject': variables.render(self.subject),
-        'preview_body': variables.render(self.text),
+        'preview_subject': variables.render(Markup(self.subject)),
```
Daverball, line 235: don't wrap the subject, the subject needs to be escaped.
```suggestion
        'preview_subject': variables.render(self.subject),
```

src/onegov/feriennet/forms/notification_template.py
Tschuppi81: And why would I need `variables.render(Markup(self.subject))` in my view as `render` takes a string returning Markup?

Daverball: Because otherwise the markup will be escaped by `escape`, wrapping in `Markup` prevents that. `escape(Markup("<br>")) == "<br>"`, whereas `escape("<br>") == "&lt;br&gt;`

Daverball: You could also get rid of the `escape` and change the signature to `Markup -> Markup` and pass `escape(self.subject)` and `Markup(self.text)`, if that is easier to understand for you.

Daverball: I see you haven't even restored the `escape` yet, so you can either do that or change the signature. It's up to you.

Tschuppi81: I did `escape` when returning but figured out that its not doing what we expect as `expand_storage_links` returns Markup

```
def render(self, text: str) -> Markup:
        ...
        result = self.expand_storage_links(text)
        return Markup.escape(result)
```
I will follow your suggestion and change the signature to `Markup -> Markup`

Daverball: The escape needed to happen at the start, just like before, but you can also move it outside the function by changing the signature, yes.

src/onegov/feriennet/views/notification_template.py
```diff
@@ -232,8 +234,8 @@ def email_iter() -> 'Iterator[EmailJsonDict]':
         'title': _("Mailing"),
         'layout': layout,
         'form': form,
-        'preview_subject': variables.render(self.subject),
-        'preview_body': variables.render(Markup(self.text)),  # noqa: MS001
+        'preview_subject': variables.render(escape(self.subject)),
```
Daverball:
```suggestion
        'preview_subject': subject,
```

Daverball (approved):

PR #1378, opened 2024-06-14
Build docker image

msom:
Build docker image for all tags (potentially allowing test builds for development).


Daverball:
@msom I would definitely do it for release builds, to make sure they're not broken. Ideally we switch the test/lint workflows to using the latest docker image as well instead of the generic ubuntu runner, then we don't have to install quite as many things, since most of the things would already be installed, which should speed up the tests. It doesn't really matter if the latest image doesn't match our branch, since we will still download the source and pull the newest versions of all the installed packages before testing.

msom:
https://github.com/OneGov/onegov-cloud/pull/1382

PR #1379, opened 2024-06-15
Adds a role option to cron jobs to set the requet' current role if re…

Tschuppi81:
Newsletter: Timed newsletter may have different content if restricted

`Cronjobs` are now executed with a role if specified

TYPE: Bugfix
LINK: ogc-1691

## Checklist

### DIFF
--- a/src/onegov/core/cronjobs.py
+++ b/src/onegov/core/cronjobs.py
@@ -149,6 +149,7 @@ class Job(Generic[_JobFunc]):
         'offset',
         'once',
         'url',
+        'as_role',
     )
 
     app: Framework
@@ -160,6 +161,7 @@ class Job(Generic[_JobFunc]):
     offset: float
     once: bool
     url: str | None
+    as_role: str | None
 
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


Tschuppi81:
Cronjobs get executed with no role nor user. This lead to stripped content for newsletters configured to be sent at a specific date/time.

Daverball:
This seems wrong to me. I don't see why we would ever want to send out newsletter content that should be invisible to logged out users. If anything we should be using the role of the user we're sending the newsletter to, if they have an account, but definitely don't send everyone a newsletter as if they're an admin user.

Also the whole `as_role` thing for cronjobs seems like overkill to me, I don't think it's necessary. If there's code that's hard to re-use because of privileges then that could should be fixed, we shouldn't mess with the privileges, it's a recipe for disaster.

Tschuppi81:
I agree for 'Private' content. 'Secret' content is visible if you know the link therefore I think it should be able to be sent/rendered in a newsletter.
 
Screenshot from 2024-06-17 10-04-54

However, newsletters sent by a logged-in person, will render and send according her privileges per item where as the cron job only sends out the public content. In my eyes, the cron job shall behave the same way as sending the newsletter by a admin/editor/member - I will limit this.

Daverball:
This sounds just inherently broken to me. The manual sending view is incorrect, not the cronjob. You can make an argument for including `Secret` content, but I don't think it's necessarily what people would expect to happen by default. This sounds like something that should explicitly be configured per tenant based on their preference. I think the expectation is that the newsletter people receive looks identical to the one they can see on the website.

You can also make the argument that both the cronjob and the manual view are broken, since the newsletters that are sent to users that have an account, should contain everything they can see when they're logged in. But I'm not sure whether people actually care about that use-case.

Tschuppi81:
See ticket for details

PR #1384, opened 2024-06-19
Also build image on master branch

msom:


.github/workflows/build-and-push.yml
```diff
@@ -3,8 +3,10 @@ name: Create and publish a Docker image
 on:
   workflow_dispatch:  # allow to run manually
   push:
-   tags:
-    - '**'
+    branches:
+      - master
```
Daverball, line 7:
```suggestion
      - '**'
```
I would do the build step on all branches, just like we currently do with buildkite, but I would avoid tagging the release with `head` or `latest` outside the `master` branch.

Daverball (approved):

PR #1388, opened 2024-06-20
Bugfix/files fix timestamp comparison

Tschuppi81:

## Commit message

<Module>: <Message>

<Optional Description>

TYPE: <Feature|Bugfix>
LINK: <Ticket-Number>
HINT: <Optional upgrade hint>

## Checklist

### DIFF
--- a/src/onegov/org/models/file.py
+++ b/src/onegov/org/models/file.py
@@ -64,9 +64,9 @@ class GroupFilesByDateMixin(Generic[FileT]):
     ) -> 'Iterator[DateInterval]':
 
         today = standardize_date(today, 'UTC')
-
-        month_end = today + relativedelta(day=31)
-        month_start = today - relativedelta(day=1)
+        month_start = today.replace(
+            day=1, hour=0, minute=0, second=0, microsecond=0)
+        month_end = month_start + relativedelta(months=1, microseconds=-1)
 
         yield DateInterval(
             name=_("This month"),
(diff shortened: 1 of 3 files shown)


src/onegov/org/models/file.py
```diff
@@ -313,7 +314,13 @@ def for_order(self, order: str) -> 'Self':
...
     @cached_property
     def intervals(self) -> tuple[DateInterval, ...]:
-        return tuple(self.get_date_intervals(today=sedate.utcnow()))
+        """
+        Returns the date intervals as local time in reverse order ('this
+        month', 'last month', 'this year', 'last year', 'older').
+
+        """
+        return tuple(self.get_date_intervals(
+            today=datetime.now().astimezone()))
```
Daverball, line 323: This doesn't work, all our servers are in UTC, not in local time, since we could very well have multiple tenants with different timezones on the same server.

I'd personally treat this as a "won't fix", since it doesn't seem worth the cost of refactoring this. All you're getting is a small subset of additional files being put in the "wrong" bucket per month/year. If we had a "Yesterday" it's be a different story, but we don't.

Also it's fairly unlikely people will upload files around midnight anyways.

Daverball: I'm also not sure it's worth snapping the time interval to the start/end of the day, if anything not doing it helps us, since generally people will check these around the same time as they're created.

Tschuppi81: Got it.

Tschuppi81: So, where are we converting database/utc timestamps back to localtime what the user sees?

Daverball: `layout.format_date` and related functions

PR #1395, opened 2024-06-21
Feature/ogc 1691 newsletter note secret and private content not beeing sent

Tschuppi81:
Newsletter: Adds note for secret and private content not being sent. Scheduled newsletter contains same content as if sent by logged-in person.

TYPE: Feature
LINK: ogc-1691


src/onegov/org/views/newsletter.py
```diff
@@ -99,6 +112,51 @@ def news_by_newsletter(
     return request.exclude_invisible(query.all())
```
Daverball, line 112: This will still exclude invisible content for the cronjob, so you should skip this step now.

But you should also take publication into account. Right now even news that are set to be published in the future or have ended publication would be included. You should be able to filter it within the query via `News.published`.

src/onegov/org/views/newsletter.py
```diff
@@ -82,15 +82,28 @@ def get_newsletter_form(
 
 def news_by_newsletter(
     newsletter: Newsletter,
-    request: 'OrgRequest'
+    request: 'OrgRequest',
+    visibility: str | tuple[str, ...] = 'public',
```
Daverball, line 86: Rename `visibility` to `access`. We should use the same name everywhere.

It also seems weird to pass in a list, just use one value and keep a list of access levels, if the index of the access of the News is lower than the index of the access provided, then it should be hidden.

I've used a similar idea for the `widest_access` in `onegov.org.utils`. You can reuse the `ORDERED_ACCESS` tuple and include everything below the index of the access given, i.e. `valid_accesses = ORDERED_ACCESS[ORDERED_ACCESS.index(access):]`

src/onegov/org/views/newsletter.py
Tschuppi81: I thought I covered it with a manual test, but no. The added test reproduces what you say

Tschuppi81:
I now split newsletter preview and sending from compiling newsletter content as they use different filters

src/onegov/org/views/newsletter.py
```diff
@@ -80,13 +80,62 @@ def get_newsletter_form(
     return form
 
 
-def news_by_newsletter(
+def newsletter_news_by_access(
     newsletter: Newsletter,
-    request: 'OrgRequest'
+    request: 'OrgRequest',
+    access: str = 'public',
+    include_wider_access: bool = True,
```
Daverball, line 87: I don't think we ever want to exclude wider access levels. So until we have a use-case for it I would get rid of this argument and the corresponding if/else below.

src/onegov/org/views/newsletter.py
```diff
@@ -80,13 +80,62 @@ def get_newsletter_form(
...
+
+    Raises:
+    - ValueError: If an invalid access level is provided.
+    """
 
     news_ids = newsletter.content.get('news')
+    if not news_ids:
+        return None
+
+    if access not in ('public', 'secret', 'private'):
```
Daverball, line 110:
```suggestion
    if access not in ORDERED_ACCESS:
```
There's no reason to restrict this function to only those three access levels

Daverball (approved):
Looks good. Just a couple of nits about the `newsletter_news_by_access` function.

Also to get the CI lint to pass you need to merge master. (and also run `make update` locally)

src/onegov/org/views/newsletter.py
Tschuppi81: Absolutely!

PR #1417, opened 2024-06-28
Feature/ogc 1701 delete resource including future reserverations

Tschuppi81:

## Commit message

<Module>: <Message>

<Optional Description>

TYPE: <Feature|Bugfix>
LINK: <Ticket-Number>
HINT: <Optional upgrade hint>

## Checklist

### DIFF
--- a/src/onegov/org/views/resource.py
+++ b/src/onegov/org/views/resource.py
@@ -27,6 +27,7 @@ from onegov.org.models.external_link import (
 from onegov.org.utils import group_by_column, keywords_first
 from onegov.reservation import ResourceCollection, Resource, Reservation
 from onegov.ticket import Ticket, TicketCollection
+from onegov.pay import PaymentCollection
 from operator import attrgetter, itemgetter
 from purl import URL
 from sedate import utcnow, standardize_date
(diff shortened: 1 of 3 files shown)


Tschuppi81:
Basically, I can delete resources with existing reservations now. To do so, I extended traits and delete related payments.

src/onegov/org/views/resource.py
```diff
@@ -595,20 +596,21 @@ def view_resource(
...
 
     def handle_reservation_tickets(reservation: 'BaseReservation') -> None:
         ticket = tickets.by_handler_id(reservation.token.hex)
         if ticket:
             assert request.current_user is not None
+
             close_ticket(ticket, request.current_user, request)
-            ticket.create_snapshot(request)
+            if not ticket.snapshot:
+                ticket.create_snapshot(request)
```
Daverball, line 608: Shouldn't we always create a snapshot? Otherwise the snapshot will start to diverge from the actual resource.

src/onegov/org/views/resource.py
```diff
@@ -595,20 +596,21 @@ def view_resource(
 def handle_delete_resource(self: Resource, request: 'OrgRequest') -> None:
 
     request.assert_valid_csrf_token()
-
-    # FIXME: should we move this attribute to the base class?
-    #        it could just be statically True or False by default
-    if not self.deletable:  # type:ignore[attr-defined]
```
Daverball, line 601: I think this condition should stay, didn't we discuss, that resources with future reservation should still remain undeletable? In that case you need to change the `deletable` property, rather than remove this condition.

Daverball (changes requested):

PR #1419, opened 2024-07-01
Moves user default active filter from path to navigation

Tschuppi81:
User Admin: Show 'active' users by default 

Move implementation from path to navigation

TYPE: Feature
LINK: ogc-1710


Tschuppi81:
Here is the improvement from your comment https://github.com/OneGov/onegov-cloud/commit/d5fc407b74484d9e05c5714b43de518be08a98a6

Daverball (approved):

PR #1431, opened 2024-07-08
Ogc 1612 surveys

BreathingFlesh:

## Commit message

Org: Surveys

Add option to create and view results of surveys

TYPE: Feature
LINK: OGC-1612

## Checklist


src/onegov/form/assets/js/snippets.jsx
```diff
@@ -61,9 +62,10 @@ var FormSnippet = React.createClass({
     },
     render: function() {
         var name = this.props.snippet[0];
+        var id = this.props.snippet[2].toLowerCase().replace(/[^a-z0-9]/g, '-');
 
         return (
-            <div className="formcode-snippet">
+            <div className="formcode-snippet" id={id}>
```
Daverball, line 68: same thing here

src/onegov/form/models/definition.py
```diff
@@ -214,3 +221,176 @@ def for_new_name(self, name: str) -> 'Self':
...
+    __tablename__ = 'surveys'
+
+    # # for better compatibility with generic code that expects an id
+    # # this is just an alias for `name`, which is our primary key
+    @hybrid_property
+    def id(self) -> str:
+        return self.name
+
+    #: the name of the form (key, part of the url)
+    name: 'Column[str]' = Column(Text, nullable=False, primary_key=True)
```
Daverball, line 239: Rather than copying form definitions here I'd suggest specifying a standard `id` field with a UUID or integer. The fewer special cases like `FormDefinition` we have, the better. You can keep the name as a URL key by making it `unique=True` instead of `primary_key=True`.

Daverball: It may also be worth considering putting these all in the same table with the same abstract polymorphic base-class. So when we add new features to forms, they will copy over to surveys without needing to duplicate all the code. Especially because the `RegistrationWindow` and `FormSubmission` probably don't need their own subclass for surveys, any changes in logic can be made based on the type of the linked definition. If it's all in the same table we don't need to duplicate all the related tables.

src/onegov/form/validators.py
```diff
@@ -344,6 +344,43 @@ def __call__(self, form: 'Form', field: 'Field') -> 'Form | None':
...
+        if parsed_form is None:
+            return None
+
+        # Exclude fields that are not allowed in surveys
+        errors = None
+        for field in parsed_form._fields.values():
+            if isinstance(field, (UploadField, DateField, TimeField,
+                                  DateTimeLocalField)):
+                error = field.gettext(self.invalid_field_type.format(
+                    label=field.label.text))
```
Daverball, line 371:
```suggestion
    invalid_field_type = _("Invalid field type for field '${label}'. Please "
                           "use the plus-icon to add allowed field types.")

    def __call__(self, form: 'Form', field: 'Field') -> 'Form | None':
        from onegov.form.fields import UploadField

        parsed_form = super().__call__(form, field)
        if parsed_form is None:
            return None

        # Exclude fields that are not allowed in surveys
        errors = None
        for field in parsed_form._fields.values():
            if isinstance(field, (UploadField, DateField, TimeField,
                                  DateTimeLocalField)):
                error = field.gettext(self.invalid_field_type %
                    {'label': field.label.text})
```
This is the proper way to do dynamic string interpolation with translation strings. Your way would've worked too but it was backwards, you needed to format after calling `gettext`, not before, otherwise it remain untranslated because the interpolated translation string does not exist. Remember to update the translations with the missing $ as well.

Daverball (changes requested):
Looks pretty good. I'm not thrilled about the amount of code duplication between forms and surveys however. I think we may be able to reduce some of it by using a common abstract base class that both models inherit from. Submissions and registration windows can be shared between forms and surveys, I don't think they each need their own version.

Separate collections do make sense however, hence the introduction of a common abstract base class, rather than making surveys a subclass of forms or vice versa.

PR #1444, opened 2024-07-18
Feature/ogc 1160 formcode warning for empty fieldsets

Tschuppi81:
Form: Raise error for empty field sets

TYPE: Feature
LINK: ogc-1160

### DIFF
--- a/src/onegov/form/parser/core.py
+++ b/src/onegov/form/parser/core.py
@@ -1156,18 +1156,19 @@ class CheckboxField(OptionsField, Field):
 @lru_cache(maxsize=1)
 def parse_formcode(
     formcode: str,
-    enable_indent_check: bool = False
+    enable_edit_checks: bool = False
 ) -> list[Fieldset]:
     """ Takes the given formcode and returns an intermediate representation
     that can be used to generate forms or do other things.
 
     :param formcode: string representing formcode to be parsed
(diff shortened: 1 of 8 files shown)


src/onegov/form/parser/core.py
```diff
@@ -1188,6 +1188,8 @@ def parse_formcode(
             parse_field_block(block, field_classes, used_ids, fs)
             for block in (fieldset[label] or ())
         ]
+        if not fs.fields:
+            raise errors.EmptyFieldsetError(label)
```
Daverball, line 1192: I'm a little bit concerned about putting this here. Since there almost certainly will be old forms that have this issue and rather than continue to work, they will now stop working. If you put this here you also need to write a migration to clean out all the empty field sets from existing form definitions.

For now it might be easier to make this check optional, just like the indent check. Then you can only enable it within the validator. Alternatively you can do the check entirely in the validator by iterating over the parsed form's fieldsets.

Daverball (changes requested):
Looks good, but there's a pitfall about the current placement of the check.

src/onegov/form/parser/core.py
Tschuppi81: Valid point!

Tschuppi81: done

Daverball (approved):

PR #1446, opened 2024-07-22
Reduce newlines by replacing carriage returns and newlines in once

Tschuppi81:
Directories: Adjust string field rendering to prevent many newlines

TYPE: Bugfix
LINK: ogc-1746


src/onegov/form/display.py
```diff
@@ -89,7 +89,7 @@ def __call__(self, field: 'Field') -> Markup:
             if field.render_kw.get('data-editor') == 'markdown':
                 return render_untrusted_markdown(field.data)
 
-        return self.escape(field.data or '').replace('\n', Markup('<br>'))
+        return self.escape(field.data or '').replace('\r\n', Markup('<br>'))
```
Tschuppi81, line 92: This certainly helps but it's not what you changed when you removed the `structure` keyword. So still investigating...

Daverball: This is less reliable than replacing `\n`, since regardless of which style of newlines are used `\n` or `\r\n`, it will result in correct line splitting. Splitting on `\r\n` is only correct in the latter case.

You could argue for using `splitlines` and `join`, although that's less efficient than this solution.

Daverball:
Also for the mypy errors on CI you can pin mypy to `<1.11`, I'll fix these tomorrow.

src/onegov/form/display.py
Tschuppi81: Then we could go for replacing both newline styles, first `\r\n` then `\n`
`return self.escape(field.data or '').replace('\r\n', Markup('<br>')).replace('\n', Markup('<br>'))`

Daverball: That wouldn't really change anything however. Yes, you may end up with some lonely `\r` in your HTML, but that shouldn't really matter. Neither `\n`, `\r` nor `\r\n` do anything in HTML unless you use `<pre>` or enable the equivalent setting through CSS for arbitrary elements.

Generally this shouldn't be an issue with form display, if the newlines are present in the data, they're there and should be rendered. From the sound of it the issue is in the actual form, where the `TextArea` input doesn't properly render the newlines, so it's not obvious they're there until you see the rendered output of the data (this part may be a Windows specific issue, where the data was originally entered with UNIX style newlines, but the people trying to edit the data use Windows, so everything looks like it's on one line instead, although I'm not aware that this is actually ever an issue with HTML forms, so I'd be very surprised if it were). Although I'm not at my work computer so I can't check the Linear issue. I can take a look tomorrow what this is about.

Tschuppi81: No worries, we can check tomorrow!

Daverball: Alright I found the issue, the problem is that we linkify the output using `layout.linkify_field`, which normalizes the generated HTML, so it inserts a second set of newlines after the `<br>` we inserted which then once again get replaced by `<br>`.

I think the most sane solution is to check in `layout.linkify` if `text` is an instance of `Markup`, and if so, then skip the `replace('\n', '<br>')` so we don't insert additional newlines into text that's already been turned into `Markup`.

Daverball: The reason your fix appeared to work, is because `\r\n` won't ever be present, since HTML forms always use UNIX style newlines. So you essentially just skipped one of the two redundant line splitting steps unconditionally, but this will produce incorrect output everywhere that relies on the line splitting happening in the field display, before it's passed to `linkify_field`.

Daverball: As to why this suddenly started happening: `linkify` used to unconditionally sanitize the output, so it removed the `<br>` tags we had added in the field display, but due to the normalization it reinserted `\n`, which then once again got replaced with `<br>`. 

This is no longer true. When we pass `Markup` into `linkify`, it will no longer sanitize `Markup`, since the content is already sanitized (this replaced your old hack of having to manually pass a boolean argument to that function, to skip the sanitation step).

Daverball: Long story short `onegov.org.layout.Layout.linkify` should be changed to:

```python
    def linkify(self, text: str | None) -> Markup | None:
        if text is None:
            return None

        if isinstance(text, TranslationString):
            # translate the text before applying linkify if it's a
            # translation string
            text = self.request.translate(text)

        linkified = linkify(text)
        if isinstance(text, Markup):
            # we don't need to insert br tags, since we started with `Markup`
            return linkified
        return linkified.replace('\n', Markup('<br>'))
```

Tschuppi81: Oh gosh, quite complex the replacing and re-adding of newlines, br tags you name it :-D

I will check with your proposed `linkify` implementation

Tschuppi81: And thank you for your analysis!

Daverball:
You need to remove the mypy pin again, since it has been removed on master.

Tschuppi81:
My bad, the linkify tests are here https://github.com/OneGov/onegov-cloud/blob/master/tests/onegov/core/test_utils.py#L120

Daverball:
That's technically only testing the core function and not the org layout one. But anything that's either part of the layout or templates is generally not very well unit tested. We could write unit tests to avoid further regressions here, but it's also not a high priority.

PR #1450, opened 2024-07-25
Bugfix/pro 1295 support 26 character reference in booking text for pofi

Tschuppi81:
Feriennet: Import bank statements now supports 27 character reference number in booking text (POFI)

Also we do not break the import if one entry fails

TYPE: Bugfix
LINK: ogc-1295

Open item:
Add detailed tests later

### DIFF
--- a/src/onegov/activity/iso20022.py
+++ b/src/onegov/activity/iso20022.py
@@ -1,4 +1,5 @@
 import re
+import stdnum.ch.esr as esr
 
 from collections import defaultdict
 from functools import cached_property
@@ -9,7 +10,7 @@ from onegov.activity.collections import InvoiceCollection
 from onegov.activity.models import Invoice
 from onegov.activity.models import InvoiceItem
 from onegov.activity.models import InvoiceReference
-from onegov.activity.models.invoice_reference import FeriennetSchema, ESRSchema
+from onegov.activity.models.invoice_reference import FeriennetSchema
 from onegov.user import User
 from sqlalchemy import func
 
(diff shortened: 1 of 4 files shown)


src/onegov/activity/iso20022.py
```diff
@@ -128,6 +128,45 @@ def transaction_entries(root: 'etree._Element') -> 'Iterator[etree._Element]':
...
+    match = re.search(pattern_27, booking_text)
+    if match:
+        return re.sub(r'\s', '', match.group(1))
+
+    # If no 27-digit match, try 26-digit pattern
+    match = re.search(pattern_26, booking_text)
+    if match:
+        ref26 = re.sub(r'\s', '', match.group(1))
+        # append checksum as it is stored like this in the database
+        return ref26 + ESRSchema().checksum(ref26)
```
cyrillkuettel, line 164: If the booking text contains a number with a length == 26, it seems to me that there are two possibilities. 
1. They don't send the checksum.
2. They strip one leading zero.

Currently we only account for (1), but not (2). Is (2) impossible and I'm missing something?
If not, this is actually non-trivial.

Tschuppi81: Right, shorter reference numbers are possible (leading zeros), https://github.com/OneGov/onegov-cloud/blob/master/src/onegov/activity/models/invoice_reference.py#L284
But then the `get_esr()` would also fail when extracting it....

cyrillkuettel (approved):

PR #1464, opened 2024-08-05
Make nationality a multi select field

Tschuppi81:
Translator Directory: Make nationality a multi select field with a complete list of countries

TYPE: Feature
LINK: ogc-1755

onegov-translator --select /translator_directory/* migrate-nationalities --dry-run

## Checklist


src/onegov/translator_directory/utils.py
```diff
@@ -201,3 +205,13 @@ def geocode_translator_addresses(
...
 
     return trs_total, total, geocoded, skipped, coords_not_found
+
+
+def nationality_choices(locale: str | None) -> list['_Choice']:
+    assert locale
+    _locale = Locale.parse(locale)
+    nationalities = [_locale.territories.get(code) for code in
+                     _locale.territories if len(code) == 2]
+    nationalities = [(v, v) for v in sorted(nationalities)]
```
Daverball, line 215: Is there a good reason not to store the code as the value? We already have to format the value for display anyways. Storing the code is more robust, since it would allow for mixed language scenarios in the future.

```suggestion
    nationalities = sorted((
        (code, name) for code, name in
        _locale.territories.items() if len(code) == 2
    ), key=itemgetter(1))
``

You may also want to consider pinning the most likely choices to the top (`CH` and its neighbors), so you don't need to use the search for them.

Comment (Tschuppi81) [reply src/onegov/translator_directory/utils.py:215]: Agree, for multi language support this makes absolutely sense.

File: src/onegov/translator_directory/utils.py [hunk 2]
```diff
@@ -207,12 +210,41 @@ def geocode_translator_addresses(
...
+        nationalities.index(('CH', 'Schweiz'))))
+    nationalities.insert(2, nationalities.pop(
+        nationalities.index(('DE', 'Deutschland'))))
+    nationalities.insert(3, nationalities.pop(
+        nationalities.index(('FR', 'Frankreich'))))
+    nationalities.insert(4, nationalities.pop(
+        nationalities.index(('IT', 'Italien'))))
+    nationalities.insert(5, nationalities.pop(
+        nationalities.index(('AT', 'Österreich'))))
+    nationalities.insert(6, ('', '------'))
```
Daverball, line 231:
```suggestion
    country_names = country_code_to_name(locale)
    pinned = ('CH', 'DE', 'FR', 'IT', 'AT', 'LI')
    nationalities = [(code, name) for code, name in
                     country_names.items() if code not in pinned]
    # pin common countries on top of the list
    nationalities.insert(0, ('', '------'))  # add divider
    for code in reversed(pinned):
        nationalities.insert(0, (code, country_names.get(code, code)))
    nationalities.insert(0, ('', ''))  # add empty choice
```
Your code breaks if the locale isn't German, this should be more robust

Tschuppi81: Yes, have it on my todo list, but l like your implementation!

Tschuppi81:
I don't get the mypy issues. IMO the type `list[tuple[str, str]]` is more narrow/specific than `list[tuple[Any, str]]`. Therefore mypy should not blame me...
```
src/onegov/translator_directory/forms/accreditation.py:619: error: Incompatible types in assignment (expression has type "list[tuple[str, str]]", variable has type "list[str | tuple[Any, str] | tuple[Any, str, dict[str, Any]]] | dict[str, Any]")  [assignment]
src/onegov/translator_directory/forms/translator.py:449: error: Incompatible types in assignment (expression has type "list[tuple[str, str]]", variable has type "list[str | tuple[Any, str] | tuple[Any, str, dict[str, Any]]] | dict[str, Any]")  [assignment]
src/onegov/translator_directory/forms/mutation.py:85: error: Incompatible types in assignment (expression has type "str | None", variable has type "str")  [assignment]
src/onegov/translator_directory/forms/mutation.py:92: error: Incompatible types in assignment (expression has type "list[tuple[str, str]]", variable has type "list[str | tuple[Any, str] | tuple[Any, str, dict[str, Any]]] | dict[str, Any]")  [assignment]
```

Daverball:
> I don't get the mypy issues. IMO the type `list[tuple[str, str]]` is more narrow/specific than `list[tuple[Any, str]]`. Therefore mypy should not blame me...
> 
> ```
> src/onegov/translator_directory/forms/accreditation.py:619: error: Incompatible types in assignment (expression has type "list[tuple[str, str]]", variable has type "list[str | tuple[Any, str] | tuple[Any, str, dict[str, Any]]] | dict[str, Any]")  [assignment]
> src/onegov/translator_directory/forms/translator.py:449: error: Incompatible types in assignment (expression has type "list[tuple[str, str]]", variable has type "list[str | tuple[Any, str] | tuple[Any, str, dict[str, Any]]] | dict[str, Any]")  [assignment]
> src/onegov/translator_directory/forms/mutation.py:92: error: Incompatible types in assignment (expression has type "list[tuple[str, str]]", variable has type "list[str | tuple[Any, str] | tuple[Any, str, dict[str, Any]]] | dict[str, Any]")  [assignment]
> ```

That's because you assign to choices directly and `list` is invariant, so the `tuple` needs to match exactly. You can use `list[_Choice]` as an annotation and import `_Choice` from `wtforms.fields.choices` inside an `if TYPE_CHECKING` block. I uses that in various other places.

PR #1511, opened 2024-09-06
Fix space between slider and content

Tschuppi81:
Town6: No padding after slider on homepage

TYPE: Feature
LINK: ogc-1810

Have manually checked all instances on `aether` with the fix.
For one customer we need a slite css adjustment, see details in ticket.


BreathingFlesh (changes requested):
There could be something else instead of a slider in the first container, for example the focus widget. In that case the space is needed. If the slider is the first element on the website it should not be in a container at all. I think I already fixed this in the template for new instances.

PR #1559, opened 2024-10-25
Ogc 508 replace elastic search by postgres v3

Tschuppi81:

## Commit message

<Module>: <Message>

<Optional Description>

TYPE: <Feature|Performance|Bugfix>
LINK: <Ticket-Number>
HINT: <Optional upgrade hint>

## Checklist

### DIFF
--- a/src/onegov/org/models/search.py
+++ b/src/onegov/org/models/search.py
@@ -6,7 +6,7 @@ from elasticsearch_dsl.query import MultiMatch
 from functools import cached_property
 from sedate import utcnow
 from sqlalchemy import func
-from typing import TYPE_CHECKING, Any, List
+from typing import TYPE_CHECKING, Any
 
 from onegov.core.collection import Pagination, _M
 from onegov.core.orm import Base
@@ -216,7 +216,7 @@ class SearchPostgres(Pagination[_M]):
             return NotImplemented
         return self.page == other.page and self.web_search == other.web_search
 
-    def subset(self) -> 'List[Searchable] | None':  # type:ignore[override]
+    def subset(self) -> 'list[Searchable] | None':  # type:ignore[override]
         return self.batch
 
     @property
@@ -227,7 +227,7 @@ class SearchPostgres(Pagination[_M]):
         return SearchPostgres(self.request, self.web_search, index)
 
     @cached_property
-    def batch(self) -> 'List[Searchable]':  # type:ignore[override]
+    def batch(self) -> 'list[Searchable]':  # type:ignore[override]
         if not self.web_search:
             return []
 
@@ -246,9 +246,9 @@ class SearchPostgres(Pagination[_M]):
         actual search results form the query.
 
         """
-        batch: List[Searchable] = self.batch
-        future_events: List[Searchable] = []
-        other: List[Searchable] = []
+        batch: list[Searchable] = self.batch
+        future_events: list[Searchable] = []
+        other: list[Searchable] = []
 
         for search_result in batch:
             if (isinstance(search_result, Event)
@@ -297,7 +297,7 @@ class SearchPostgres(Pagination[_M]):
 
     def generic_search(self) -> list['Searchable']:
         doc_count = 0
-        results: List[Any] = []
+        results: list[Any] = []
         language = locale_mapping(self.request.locale or 'de_CH')
         ts_query = func.websearch_to_tsquery(language,
                                              func.unaccent(self.web_search))
@@ -305,10 +305,12 @@ class SearchPostgres(Pagination[_M]):
 
         for base in self.request.app.session_manager.bases:
             for model in searchable_sqlalchemy_models(base):
-                if model.es_public or self.request.is_logged_in:
+                if model.es_public or self.request.is_logged_in:  # needed?
                     query = session.query(model)
                     if not self.request.is_logged_in:
-                        query = query.filter(model.es_public == True)
+                        # query = query.filter(model.es_public == True)
+                        query = query.filter(
+                            model.fts_idx_data['es_public'].astext == 'True')
 
                     if session.query(query.exists()).scalar():
                         weighted = (
--- a/src/onegov/org/models/search.py
+++ b/src/onegov/org/models/search.py
@@ -4,25 +4,29 @@ from elasticsearch_dsl.query import Match
 from elasticsearch_dsl.query import MatchPhrase
 from elasticsearch_dsl.query import MultiMatch
 from functools import cached_property
+from sedate import utcnow
+from sqlalchemy import func
+from typing import TYPE_CHECKING, Any, List
+
 from onegov.core.collection import Pagination, _M
+from onegov.core.orm import Base
 from onegov.event.models import Event
+from onegov.search.utils import searchable_sqlalchemy_models
 
-
-from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from onegov.org.request import OrgRequest
+    from onegov.search import Searchable
     from onegov.search.dsl import Hit, Response, Search as ESSearch
 
 
 class Search(Pagination[_M]):
-
     results_per_page = 10
     max_query_length = 100
 
     def __init__(self, request: 'OrgRequest', query: str, page: int) -> None:
         super().__init__(page)
         self.request = request
-        self.query = query
+        self.web_search = query
 
     @cached_property
     def available_documents(self) -> int:
@@ -35,13 +39,13 @@ class Search(Pagination[_M]):
 
     @property
     def q(self) -> str:
-        return self.query
+        return self.web_search
 
     def __eq__(self, other: object) -> bool:
         return (
             isinstance(other, self.__class__)
             and self.page == other.page
-            and self.query == other.query
+            and self.web_search == other.web_search
         )
 
     if TYPE_CHECKING:
@@ -56,11 +60,11 @@ class Search(Pagination[_M]):
         return self.page
 
     def page_by_index(self, index: int) -> 'Search[_M]':
-        return Search(self.request, self.query, index)
+        return Search(self.request, self.web_search, index)
 
     @cached_property
     def batch(self) -> 'Response | None':  # type:ignore[override]
-        if not self.query:
+        if not self.web_search:
             return None
 
         search = self.request.app.es_search_by_request(
@@ -70,7 +74,7 @@ class Search(Pagination[_M]):
 
         # queries need to be cut at some point to make sure we're not
         # pushing the elasticsearch cluster to the brink
-        query = self.query[:self.max_query_length]
+        query = self.web_search[:self.max_query_length]
 
         if query.startswith('#'):
             search = self.hashtag_search(search, query)
@@ -95,13 +99,16 @@ class Search(Pagination[_M]):
         batch = self.batch.load()
         events = []
         non_events = []
+
         for search_result in batch:
             if isinstance(search_result, Event):
                 events.append(search_result)
             else:
                 non_events.append(search_result)
+
         if not events:
             return batch
+
         sorted_events = sorted(
             events,
             key=get_sort_key
@@ -147,7 +154,7 @@ class Search(Pagination[_M]):
             first_entry = self.batch[0].load()
 
             # XXX the default view to the event should be doing the redirect
-            if first_entry.__tablename__ == 'events':
+            if first_entry.es_type_name == 'events':
                 return self.request.link(first_entry, 'latest')
             else:
                 return self.request.link(first_entry)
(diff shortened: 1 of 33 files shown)


src/onegov/org/models/search.py
```diff
@@ -159,5 +166,221 @@ def subset_count(self) -> int:
...
+        self,
+        model: Any,
+        language: str = 'simple'
+    ) -> Any:
+        # for now weight the first field with 'A', the rest with 'B'
+        weighted_vectors = [
+            func.setweight(
+                func.to_tsvector(
+                    language,
+                    getattr(model.fts_idx_data, field, '')),
```
Tschuppi81, line 281: the weighted vector bases on the static data from column `fts_idx_data` generated upon `update` or `reindex` events.

src/onegov/org/models/search.py
```diff
@@ -159,5 +166,221 @@ def subset_count(self) -> int:
...
+                             .add_columns(rank_expression))
+                    res = list(query.all())
+                    doc_count += len(res)
+                    results.extend(res)
+
+        # exclude non-public documents if user is not logged in
+        # we do this on python level as using hybrid properties and sql
+        # expressions did not work as expected
+        if not self.request.is_logged_in:
+            results = [r for r in results if r[0].es_public == True]
```
Tschuppi81, line 328: Test for `public` documents done in python as the publication needs to be tested according the current time and therefore cannot base on static data in column `fts_idx_data`

Tschuppi81:
With this approach no additional `hybrid_properties` are needed.

src/onegov/org/models/search.py
Tschuppi81: Or as you, @Daverball, suggested earlier we run re-indexing from recently published documents as a cron job...

Daverball: This is already happening for elasticsearch, so it would be more or less free to extend this with a postgres reindex: https://github.com/OneGov/onegov-cloud/blob/master/src/onegov/org/cronjobs.py#L91

Tschuppi81:
@Daverball Final review for postgres searching on separate views `/search-postgres?q=test` (not yet productive)

src/onegov/org/models/search.py
```diff
@@ -159,5 +166,229 @@ def subset_count(self) -> int:
...
+        return [r[0] for r in results]
+
+    def hashtag_search(self) -> list['Searchable']:
+        q = self.web_search.lstrip('#')
+
+        # Skip certain tables for hashtag search for better performance
+        results = [
+            doc for model in searchable_sqlalchemy_models(Base)
+            if model.es_type_name not in
+               ['attendees', 'files', 'people', 'tickets', 'users']
```
Daverball, line 356: This seems fragile to me, it seems better to me to change the mixin so you can distinguish between a model that defines es_tags and one that doesn't.

src/onegov/org/models/search.py
```diff
@@ -159,5 +166,229 @@ def subset_count(self) -> int:
...
+            if model.es_public or self.request.is_logged_in  # type:ignore
+            for doc in self.request.session.query(model).all()
```
Daverball, line 358:
```suggestion
            for doc in self.request.session.query(model)
```
You don't need to call `all` to iterate through a query, `all` will create a list for you, which you don't need here since you're iterating through the results.

src/onegov/org/models/search.py
```diff
@@ -159,5 +166,229 @@ def subset_count(self) -> int:
...
+            if doc.es_tags and q in doc.es_tags
+        ]
```
Daverball, line 360: This once again seems completely backwards to me, you're doing too much work during search. Store `es_tags` on the model in the fts data or a new column and filter the query based on that, then you also don't run into performance issues.

src/onegov/org/models/search.py
```diff
@@ -159,5 +166,229 @@ def subset_count(self) -> int:
...
+
+        # remove duplicates
+        results = list(set(results))
```
Daverball, line 363: Same problem as with `generic_search`.

src/onegov/search/indexer.py
```diff
@@ -421,18 +426,25 @@ def index(
...
             )
+
             stmt = (
                 sqlalchemy.update(table)
                 .where(id_col == sqlalchemy.bindparam('_id'))
-                .values({self.TEXT_SEARCH_COLUMN_NAME: tsvector_expr})
+                .values({
+                    self.TEXT_SEARCH_COLUMN_NAME: tsvector_expr,
+                    self.TEXT_SEARCH_DATA_COLUMN_NAME:
+                    sqlalchemy.bindparam('data', type_=sqlalchemy.JSON)
```
Daverball, line 444: It might be better to use `JSONB` here and above, otherwise you introduce additional conversion overhead on Postgres' end since it has to convert from `::json` to `::jsonb`.

src/onegov/search/integration.py
```diff
@@ -471,12 +473,13 @@ def reindex_model(model: type['Base']) -> None:
...
-        models = self.get_searchable_models()
-        index_log.info(f'Number of models to be indexed: {len(models)}')
-
         with ThreadPoolExecutor() as executor:
             results = executor.map(
-                reindex_model, (model for model in models)
+                reindex_model, (
+                    model
+                    for base in self.session_manager.bases
+                    for model in searchable_sqlalchemy_models(base)
```
Daverball, line 481: Based on the fact that we had duplicates in the search results when using this function we also probably compute the index several times for some models. So same thing here, retrieve the polymorphic base class and remove the duplicate base classes.

You will need to take additional care filtering out the non-searchable instances from the searchable ones. E.g. with `SearchableFile` the polymorphic base class is `File`, but it's not searchable.

Another way to deduplicate without introducing an additional filtering step would be to skip all the searchable models that are a subclass of another searchable model and they both have the same `__tablename__`.

Daverball (changes requested):
It looks fairly close to a first version we can deploy, there are however some engineering decisions that don't make sense to me and harm performance significantly, so I would like you to revisit those problem areas.

src/onegov/org/models/search.py
```diff
@@ -4,25 +4,29 @@
...
 class Search(Pagination[_M]):
-
     results_per_page = 10
     max_query_length = 100
 
     def __init__(self, request: 'OrgRequest', query: str, page: int) -> None:
         super().__init__(page)
         self.request = request
-        self.query = query
+        self.web_search = query
```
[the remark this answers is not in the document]
Tschuppi81: Done

src/onegov/org/models/search.py
```diff
@@ -147,7 +154,7 @@ def feeling_lucky(self) -> str | None:
             first_entry = self.batch[0].load()
 
             # XXX the default view to the event should be doing the redirect
-            if first_entry.__tablename__ == 'events':
+            if first_entry.es_type_name == 'events':
```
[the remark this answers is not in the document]
Tschuppi81: I like your proposal `isinstance(first_entry, Event)`

src/onegov/org/models/search.py
```diff
@@ -159,5 +166,229 @@ def subset_count(self) -> int:
...
+    results_per_page = 10
+    max_query_length = 100
+
+    def __init__(self, request: 'OrgRequest', query: str, page: int):
+        self.request = request
+        self.web_search = query
+        self.page = page  # page index
+
+        self.nbr_of_docs = 0
+        self.nbr_of_results = 0
```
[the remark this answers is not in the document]
Tschuppi81: Done

src/onegov/org/models/search.py
Tschuppi81: I just reworked the hashtag search including suggestions

src/onegov/search/indexer.py
Tschuppi81: done

src/onegov/org/models/search.py
Tschuppi81: done

src/onegov/search/integration.py
Tschuppi81: Duplicates disappeared once I only search in models that are not base classes of other models

src/onegov/search/utils.py
```diff
@@ -43,7 +42,9 @@ def searchable_sqlalchemy_models(
     )
 
 
-def filter_non_base_models(models: 'set[type[T]]') -> 'set[type[T]]':
+def filter_non_base_models(
+    models: 'set[type[T]]'
+) -> 'set[type[T]]':
```
Daverball, line 47: This is the opposite of what we want. This will now potentially miss some models rather than lead to overlapping queries. We want to get rid of models that are already covered by their base class, not the other way around.

Tschuppi81: So instead of indexing `Topic` and `News` we only index `Page`. Same with `XYTicket` just `Ticket`

Daverball: With your examples in the tests, yes. In reality `Page` is not searchable, so it will still include both `Topic` and `News` and not `Page` (since that was never included to begin with).

I've seen some cases where polymorphic queries weren't working properly and the query always included the base class. If that's what's happening here we always need to use the polymorphic base class, regardless of whether it's searchable or not and instead apply the searchable filter on the results.

Or we manually create a filter expression on the `type` column for all the searchable polymorphic identities. There may be some performance benefit to this approach, since we avoid generating too many individual queries this way.

Tschuppi81: With the following statement in `reindex_model` no rows will be indexed twice
```
                if i.polymorphic_on is not None:
                    q = q.filter(i.polymorphic_on == i.polymorphic_identity)
```

Daverball: I looked into it briefly:

```python
i = inspect(Topic)
i.base_mapper.class_   # Page
{
    m.polymorphic_identity
    for m in mapper.base_mapper.self_and_descendants
    if issubclass(m.class_, Searchable)
}  # {'topic', 'news'}
```

This should give you everything you need. Note that you only want to do this for models that have `i.polymorphic_on is not None`. So you want a `get_base_models` function which de-duplicates and gives you the base class for polymorphic models and just the model itself for everything else. And then you change the filter from a single polymorphic identity, to all the searchable ones.

Tschuppi81: I see, the `base_mapper` does the trick!! I did a very similar thing which lead me to identities like `['topic', 'news', 'generic', 'extended', 'image', 'general', 'landsgemeinde', 'builtin', 'custom', None, 'FRM', 'RSV', 'EVN', 'DIR', 'CHT', 'FER', 'vacation', 'daypass', 'room', 'daily-item']`

```
+            for model in models:
+                i = inspect(model)
+                idents = set(e.polymorphic_identity
+                             for e in list(i.self_and_descendants)[1:]
+                             if e.polymorphic_identity is not None)
```

src/onegov/search/integration.py
```diff
@@ -444,28 +443,24 @@ def reindex_model(model: type['Base']) -> None:
...
             session = self.session()
             try:
-                q = session.query(model).options(undefer('*'))
-                i = inspect(model)
-
-                if i.polymorphic_on is not None:
-                    q = q.filter(i.polymorphic_on == i.polymorphic_identity)
-
-                for obj in q:
+                for obj in session.query(model).options(undefer('*')):
```
Daverball, line 446:
```suggestion
                q = session.query(model).options(undefer('*'))
                                i = inspect(model)

                if i.polymorphic_on is not None:
                    q = q.filter(i.polymorphic_on.in({
                        m.polymorphic_identity for m in i.self_and_descendants})

                for obj in q:
```

src/onegov/search/utils.py
```diff
@@ -42,29 +45,34 @@ def searchable_sqlalchemy_models(
...
+            i.base_mapper.self_and_descendants
+            if issubclass(e.base_mapper.class_, Searchable)
+            if e.polymorphic_identity is not None
+            if e.base_mapper.class_ != model
+        }
+        if base_classes:
+            for base in base_classes:
+                new_models.add(base)
+        else:
+            new_models.add(model)
```
Daverball, line 73:
```suggestion
        i = inspect(model)
        new_models.add(i.base_mapper.class_ if i.polymorphic_on else model)
```

Daverball (changes requested):
This is what I was hinting at, you need to do this in two steps in separate locations, you can't do it in one function.

src/onegov/search/utils.py
Tschuppi81: Oh, my comment did not make it. 
I tried the above code and it comes back to what I mentioned in my inexistent comment(:laughing:) regarding `File` is not properly handled. `File` is not a subclass of `Searchable` but `GeneralFile` is...

```
(venv) reto@reto-pc:~/workspace/onegov-cloud$ onegov-search --select /onegov_town6/ebikon reindex 
Reindexing onegov_town6/ebikon
2024-12-13 15:44:32 - INFO: Indexing schema onegov_town6-ebikon..
Error psql indexing model 'File': 'EventFile' object has no attribute 'es_skip'
Total objects re-indexed: 12614
took 0:01:07.309937
```
That is why I came up with the latest commit

Daverball: That's what the other suggestion is for. It was missing one line, but it's supposed to filter out non-searchable polymorphic identities.

src/onegov/search/integration.py
Daverball: @Tschuppi81 This part of the suggestion should be correct now

src/onegov/search/utils.py
Tschuppi81: It now indexes `File`,`FileSet` and `FormDefinition` which both are not `Searchable`s. On the other hand I miss `GeneralFile`, `DaypassResource, `RoomResource` and `ItemResource`...

```
(venv) reto@reto-pc:~/workspace/onegov-cloud$ onegov-search --select /onegov_town6/ebikon reindex 
Reindexing onegov_town6/ebikon
2024-12-16 08:15:56 - INFO: Indexing schema onegov_town6-ebikon..
Reindexing model 'CourseEvent'..
Reindexing model 'Event'..
Reindexing model 'Ticket'..
Reindexing model 'ExternalLink'..
Reindexing model 'Directory'..
Reindexing model 'DirectoryEntry'..
Reindexing model 'SettlementRun'..
Reindexing model 'User'..
Reindexing model 'CourseAttendee'..
Reindexing model 'FormDefinition'..
Reindexing model 'Course'..
Reindexing model 'Activity'..
Reindexing model 'Resource'..
Reindexing model 'AgendaItem'..
Reindexing model 'File'..
Reindexing model 'Assembly'..
Reindexing model 'Party'..
Reindexing model 'Newsletter'..
Reindexing model 'ParliamentaryGroup'..
Reindexing model 'Person'..
Reindexing model 'Commission'..
Reindexing model 'Parliamentarian'..
Reindexing model 'Attendee'..
Reindexing model 'AgencyMembership'..
Reindexing model 'Agency'..
Reindexing model 'Page'..
Reindexing model 'Votum'..
Reindexing model 'LegislativePeriod'..
Reindexing model 'FileSet'..
took 0:01:19.406114
```
The filter now adds `base_mapper.class_`es that are not searchable and does not go through all its descendants. For example `class LandsgemeindeFile(File, SearchableFile)` `File` is picked instead of `SearchableFile`. 

I need to revert or skip this optimization of indexing `Page` only instead of `News` and `Topic`

Daverball: No, this is correct and working as intended. When you index, you index all the submodels too and only the ones with a searchable polymorphic identity.

Daverball: The logging message is confusing at this point though, I agree. Maybe switch to `f"Reindexing table  '{mode.__tablename__}'.."`?

Or you generate a log message for each searchable subclass.

Tschuppi81: Yes, the search is not working as the `fts_idx_data` columns are missing for e.g. `Page` model. I added the `fts_idx_data` columns to `Topic` and `News`.

Tschuppi81: > The logging message is confusing at this point though, I agree. Maybe switch to `f"Reindexing table '{model.__tablename__}'.."`?
> 
> Or you generate a log message for each searchable subclass.

Right, I confused myself

Tschuppi81:
@Daverball Could you please check my latest changes?

src/onegov/org/models/search.py
```diff
@@ -163,3 +169,241 @@ def suggestions(self) -> tuple[str, ...]:
...
+        self.request = request
+        self.query = query
+        self.page = page  # page index
+
+        self.number_of_docs = 0
+        self.number_of_results = 0
+
+        self.search_models = {
+            model for base in self.request.app.session_manager.bases
+            for model in searchable_sqlalchemy_models(base)}
```
Daverball, line 197: This is still not really correct for performing as few non-overlapping queries as possible. But if we go with a separate table for the search index, this should simplify away to some degree.

Daverball (changes requested):
The indexer looks a lot better now. But I think we still have fundamental problems with the actual search. I think it's time to consider alternative architectures, such as a single shared table for the search metadata, so we can do all the counting/filtering/slicing of the entries in the database on a single query. So we only actually have to go out and fetch the models we're actually displaying results for on the current page.

This should mean we now only have one potentially expensive fts query, with the rest turning into simple "Fetch these primary keys from table X and these others from table Y" queries that should be very fast.

src/onegov/org/models/search.py
```diff
@@ -163,3 +169,241 @@ def suggestions(self) -> tuple[str, ...]:
...
+    def available_documents(self) -> int:
+        if not self.number_of_docs:
+            _ = self.load_batch_results
+        return self.number_of_docs
+
+    @cached_property
+    def available_results(self) -> int:
+        if not self.number_of_results:
+            _ = self.load_batch_results
+        return self.number_of_results
```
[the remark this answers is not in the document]
Tschuppi81: On the search results page we say `number of documents` is the total number of documents in the schema where as `available results` is the number of documents we found for the current query.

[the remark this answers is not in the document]
Daverball: I see, but then you aren't really correctly counting `number of documents` correctly. Unless I'm missing something your number of documents is just the number of search results without removing the duplicates, which are only there because you're emitting too many queries that can partially overlap.

src/onegov/search/upgrade.py
```diff
@@ -3,28 +3,24 @@
...
+    if not context.has_table('search_index'):
+        context.operations.create_table(
+            'search_index',
+            Column('id', UUID, primary_key=True, default=uuid4),
+            Column('owner_id_int', Integer, nullable=True),
+            Column('owner_id_uuid', UUID, nullable=True),
+            Column('owner_type', String, nullable=False),
+            Column('fts_idx_data', JSONB, default={}),
+            Column('fts_idx', TSVECTOR)
+        )
```
Daverball, line 26: This migration is not necessary. The ORM automatically adds missing tables. You only need migrations for changes to the table columns/indexes/constraints.

tests/shared/fixtures.py
```diff
@@ -228,9 +232,68 @@ def session_manager(postgres_dsn):
...
+    mgr.bases.append(ORMBase)
+
+    engine = create_engine(postgres_dsn)
+
+    # create schema
+    if not engine.dialect.has_schema(engine, schema_name):
+        engine.execute(f'CREATE SCHEMA {schema_name}')
+
+    # verify search index table exists
+    if not engine.has_table('search_index', schema=schema_name):
```
Tschuppi81, line 280: Some tests failed saying the `search_index` relation is missing, so I created this fixture. But somehow the tests don't setup ORM properly. Do you see why @Daverball?

Daverball:
> Since I introduced table `search_index` (class SearchIndex) the tests fail with the following error which I am not able to resolve
> 
> ```
>    File "/__w/onegov-cloud/onegov-cloud/src/onegov/search/indexer.py", line 25, in <module>
>     from onegov.search.search_index import SearchIndex
> ImportError: Error importing plugin "tests.shared": No module named 'onegov.search.search_index'
> ```

You forgot to add this file to your commits, it's not currently part of the pull request.

tests/shared/fixtures.py
Daverball: Probably the same problem, the file is missing from the pull request.

src/onegov/search/indexer.py
```diff
@@ -410,45 +412,139 @@ def index(
...
+                elif isinstance(_owner_id, str):
+                    stmt = (
+                        sqlalchemy.delete(SearchIndex.__table__)
+                        .where(SearchIndex.owner_id_str == _owner_id)
+                    )
+
+                if session is None:
+                    connection = self.engine.connect()
+                    with connection.begin():
+                        connection.execute(stmt)
```
Daverball, line 540: Don't you need to set the schema here as well?

src/onegov/search/indexer.py
```diff
@@ -410,45 +412,139 @@ def index(
...
-                'to_tsvector(:language, :data)').bindparams(
-                sqlalchemy.bindparam('language', type_=sqlalchemy.String),
-                sqlalchemy.bindparam('data', type_=sqlalchemy.JSON)
-            )
+
             stmt = (
-                sqlalchemy.update(table)
-                .where(id_col == sqlalchemy.bindparam('_id'))
-                .values({self.TEXT_SEARCH_COLUMN_NAME: tsvector_expr})
+                sqlalchemy.insert(SearchIndex.__table__)
```
Daverball, line 461: Postgres supports upserting, so rather than `DELETE` +`INSERT` we should run a single upsert query (`ON CONFLICT "unique_constraint_name" DO UPDATE SET fts_index = EXCLUDED.fts_index ...`). You will need to add a combined unique index on the four owner columns though. So Postgres detects that you're trying to insert a row that's already in the index.

Daverball (commented):
This is looking a lot closer to what we want. However there are still quite a few things we can improve performance-wise here.

They don't all need to happen in this PR, but we should at least attempt to get the ones done that would change the structure and the indexes of `SearchIndex`. The other optimizations can be done in a follow-up PR.

src/onegov/core/framework.py
```diff
@@ -1500,10 +1505,37 @@ def unsign(self, text: str, salt: str = 'generic-signer') -> str | None:
...
+                Column('owner_id_int', Integer, nullable=True),
+                Column('owner_id_uuid', String, nullable=True),
+                Column('owner_id_str', String, nullable=True),
+                Column('fts_idx_data', JSONB, default={}),
+                Column('fts_idx', TSVECTOR),
+            )
+
+            metadata.create_all(engine)
+            assert engine.has_table('search_index', schema=schema_name)
+
```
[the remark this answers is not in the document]
Tschuppi81: Right, I had issues while developing the `search_index` table. Now I cannot reproduce it.
Going to remove

src/onegov/org/models/search.py
```diff
@@ -163,3 +173,275 @@ def suggestions(self) -> tuple[str, ...]:
...
+        # as a member we only want to see public and member content
+        if self.request.is_member:
+            query = query.filter(
+                SearchIndex.fts_idx_data['access'].
+                astext.in_(('public', 'member')))
+
+        # as non-manager we only want to see public content
+        elif not self.request.is_manager:
+            query = query.filter(
+                SearchIndex.fts_idx_data.contains({'es_public': True}))
```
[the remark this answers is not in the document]
Tschuppi81: Used the `NewsCollection` approach to filter for access level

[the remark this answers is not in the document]
Tschuppi81: I now added `UTCPublicationMixin` to `SearchIndex` and copy the publication dates from the source model if available

src/onegov/org/models/search.py
```diff
@@ -163,3 +173,275 @@ def suggestions(self) -> tuple[str, ...]:
...
+                          filter(model.id == owner_id).first())
+
+                if result:
+                    results.append((result, rank))
+
+        results.sort(key=lambda x: x[1], reverse=True)
+        self.number_of_results = len(results)
+
+        # only return the model instances, not the rank
+        return [r[0] for r in results]
```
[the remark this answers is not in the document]
Tschuppi81: Ticket opened https://linear.app/onegovcloud/issue/OGC-2260

src/onegov/search/indexer.py
```diff
@@ -410,45 +412,139 @@ def index(
...
                 connection = self.engine.connect()
                 with connection.begin():
-                    connection.execute(stmt, content)
+                    connection.execute(f'SET search_path TO "{schema}"')
+                    connection.execute(stmt, params)
             else:
                 # use a savepoint instead
                 with session.begin_nested():
-                    session.execute(stmt, content)
+                    session.execute(f'SET search_path TO "{schema}"')
```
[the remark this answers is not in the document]
Tschuppi81: Indeed, it caused my `hstore does not exist` issue

src/onegov/search/search_index.py
```diff
@@ -0,0 +1,44 @@
...
+    owner_id_int = Column(Integer, nullable=True)
+
+    #: The uuid id of the original model if it is a uuid
+    owner_id_uuid = Column(UUID, nullable=True)
+
+    #: The string id of the original model if it is a string
+    owner_id_str = Column(String, nullable=True)
+
+    #: The full text search index data (fts properties)
+    fts_idx_data = Column(JSONB, default={})
```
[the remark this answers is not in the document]
Tschuppi81: `access`/`tags`/`suggestions`/`last_change` moved to separate columns

src/onegov/search/search_index.py
```diff
@@ -21,22 +22,38 @@ class SearchIndex(Base, UTCPublicationMixin):
...
+    public = Column(Boolean, nullable=False, default=False)
+
+    #: Access level of entry (AccessExtension::access)
+    access = Column(String, nullable=False, default='public')
+
+    #: Timestamp of the last change to the entry (Searchable::es_last_change)
+    last_change = Column(UTCDateTime, nullable=False)
+
+    #: Tags associated with the entry (Searchable::es_tags)
+    tags = Column(JSONB, default=None)
```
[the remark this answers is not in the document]
Tschuppi81: changed to hstore

src/onegov/search/search_index.py
```diff
@@ -0,0 +1,94 @@
...
+        Index('ix_search_index_owner_id_int', 'owner_id_int'),
+        Index('ix_search_index_owner_id_uuid', 'owner_id_uuid'),
+        Index('ix_search_index_owner_id_str', 'owner_id_str'),
+        Index('ix_search_index_public', 'public'),
+        Index('ix_search_index_access', 'access'),
+        Index('ix_search_index_last_change', 'last_change'),
+        Index('ix_search_index_tags', 'tags'),
+        Index('ix_search_index_suggestion', 'suggestion'),
+        Index('ix_search_index_fts_idx_data', 'fts_idx_data'),
+        Index('ix_search_index_fts_idx', 'fts_idx'),
```
Daverball, line 84: Some of these should probably be special indexes, like the `TSVECTOR`/`HSTORE`/`ARRAY` columns, the default index is a B-Tree index that only supports basic equality/ordering operations. But for these columns we use other operators, have a look at https://www.postgresql.org/docs/current/indexes-types.html to choose the most appropriate index type for the kinds of queries we execute on this table.

PR #1562, opened 2024-10-29
Remove invalid state filter

Tschuppi81:
Fsi: Remove invalid state filter for survey export

TYPE: Bugfix
LINK: ogc-1889


Tschuppi81:
@BreathingFlesh I picked-up this issue from sentry and removed the invalid `state` filter but cannot really judge if this has an unwanted impact on the results. What is your opinion?

BreathingFlesh (approved):

PR #1574, opened 2024-11-07
Prevent deleting next file info

Tschuppi81:
Files: Fix ajax request removing to many elements for action delete

TYPE: Bugfix
LINK: ogc-1851

### DIFF
--- a/src/onegov/town6/templates/macros.pt
+++ b/src/onegov/town6/templates/macros.pt
@@ -2355,7 +2355,6 @@
             ic-on-success="
                 var info = $(elt).closest('tr');
                 info.prev().remove();
-                info.next().remove();
                 info.remove();
             "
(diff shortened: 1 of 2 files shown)


src/onegov/town6/templates/macros.pt
```diff
@@ -2353,10 +2353,21 @@
         <a class="button secondary confirm delete" role="button"
             ic-delete-from="${request.link(file)}?csrf-token=${layout.csrf_token}"
             ic-on-success="
-                var info = $(elt).closest('tr');
-                info.prev().remove();
-                info.next().remove();
```
Tschuppi81, line 2358: this line was removing the following file -> bad

BreathingFlesh (approved):

PR #1587, opened 2024-11-21
Remove ruff pin and make use of RUF035 rule for markupsafe

Tschuppi81:
Remove ruff pin lower 0.7.1 and replace the MS001 with RUF035 for markupsafe


Tschuppi81:
Any idea why the ruff executable is missing @Daverball 
```
Run /app/bin/ruff check src/ tests/ stubs/ --output-format github
  /app/bin/ruff check src/ tests/ stubs/ --output-format github
  /app/bin/flake8 \
  --format '::error file=%(path)s,line=%(row)s,col=%(col)s,title=Flake8 (%(code)s)::%(text)s' \
  src/
  shell: bash --noprofile --norc -e -o pipefail {0}
/__w/_temp/91917a24-9662-4225-b319-b7003235f8f6.sh: line 1: /app/bin/ruff: No such file or directory
Error: Process completed with exit code 127.
```

setup.cfg
```diff
@@ -186,9 +186,7 @@ lint =
     # TODO: consider hosting our own plugin on pypi
     #       or get an equivalent rule added to bandit
     #       for now we pin to a specific commit for safety
-    flake8-markupsafe@git+https://github.com/vmagamedov/flake8-markupsafe#egg=b391bd13df9330e01666d304b7f4403d67e5ceba
     flake8-type-checking
-    ruff<=0.7.1
```
Daverball, line 191: This is why you're getting CI errors, ruff is still a dependency, it just shouldn't be pinned to `<=0.7.1`. I'd revert to the previous `!=0.7.2`, since we know that one is broken.

Also make sure to update the pin in `.pre-commit-config.toml` (`make update` or `pre-commit autoupdate` should do that for you)

src/onegov/feriennet/forms/attendee.py
```diff
@@ -318,7 +318,7 @@ def populate_tos(self) -> None:
 
         layout = DefaultLayout(self.model, self.request)
 
-        self.accept_tos.description = Markup(render_macro(  # noqa: MS001
+        self.accept_tos.description = Markup(render_macro(
```
Daverball, line 321: This one should still trigger, so if it doesn't, that's a bug in the ruff implementation I'll have to look into.

setup.cfg
Tschuppi81: Oh my!!!

src/onegov/feriennet/forms/attendee.py
Tschuppi81: It didn't trigger!

Daverball: I see the problem, the Ruff implementation is correct, this is using `chameleon.utils.Markup`, we should be using `markupsafe.Markup` here too. Technically they're interchangeable as far as template rendering is concerned, but it's better to be consistent.

We can also add the following section to `pyproject.toml`, so this other `Markup` class is detected as well:

```toml
[ruff.lint.ruff]
extend-markup-names = ["chameleon.utils.Markup"]
```

src/onegov/feriennet/forms/attendee.py
```diff
@@ -318,7 +318,7 @@ def populate_tos(self) -> None:
 
         layout = DefaultLayout(self.model, self.request)
 
-        self.accept_tos.description = Markup(render_macro(
+        self.accept_tos.description = Markup('{}').format(render_macro(
```
Daverball, line 321: Yeah, that doesn't work with `chameleon.utils.Markup`, that particular class only has a `__html__` method, nothing else, so you can't use `format` with it.

Actually you should just be able to use `render_macro` directly, since it return `markupsafe.Markup` now, no need to wrap it.

src/onegov/feriennet/forms/attendee.py
Tschuppi81: Added it if somebody ever uses it

PR #1593, opened 2024-11-26
Organization mapping if only office assigned

Tschuppi81:
Agency: Import Staka LU: Organisation mapping if only office assigned

TYPE: Feature
LINK: ogc-1891


Tschuppi81:
moved to branch feature/ogc-1927-organisation-mapping-if-only-office-assigned

PR #1601, opened 2024-11-28
Ogc 1912 course subscriptions

BreathingFlesh:

## Commit message

FSI: Course Subscribtions

Ensure selected course is at least 6 years in the future from the last course the attendee subscribed to

TYPE: Feature
LINK: OGC-1912

## Checklist


Daverball (approved):
LGTM

You could perhaps speed things up slightly by moving the `last_subscribed_event` filter in `possible_course_events` into a subquery, like we do with `excl`, so we only send one query to the database, but it's probably not worth the extra effort in this case.

We can always optimize this later if it ends up being a problem.

PR #1606, opened 2024-12-03
Feature/person function width

Tschuppi81:
Town6: Sidebar make person function as wide as document names and map

TYPE: Feature
LINK: None


Tschuppi81:
Just saw that a persons function does not use the full width
image

original:
image

Also reduced the margin between the sidebar sections

BreathingFlesh (approved):

PR #1642, opened 2025-01-03
Sends email notifications for directory entries with publication date…

Tschuppi81:
Directory: Delay sending update notifications to subscribers if publication starts in future

TYPE: Feature
LINK: ogc-1825

## Checklist

    - [x] I have tested sending mails


src/onegov/org/cronjobs.py
```diff
@@ -84,6 +90,34 @@ def send_scheduled_newsletter(request: 'OrgRequest') -> None:
...
+    now = utcnow()
+
+    for directory in DirectoryCollection(request.session).query().filter(
+            ExtendedDirectory.enable_update_notifications == True):
+
+        directory_entries = ExtendedDirectoryEntryCollection(
+            directory).query().filter(
+                now >= ExtendedDirectoryEntry.publication_start,
+                now < ExtendedDirectoryEntry.publication_end,
+                # ExtendedDirectoryEntry.published == True,
```
Tschuppi81, line 110: I got confused as `ExtendedDirectoryEntry.published == True` did not pass the tests. Any idea why?

Daverball: Good question, not sure.

Daverball (changes requested):
I think this is currently a bit too complicated, I would use the exact same logic as in `reindex_published_models` and just send notifications for everything that was published in the last hour, rather than rely on additional metadata.

In fact you can just add another `isinstance(obj, ExtendedDirectoryEntry)` check to the existing function and trigger the notification in that same loop in order to avoid a redundant query.

Since cronjobs always run at specific intervals there's no chance of triggering too many notifications (there is however a small chance of skipping one during an upgrade)

While storing whether or not the notification was sent seems more robust at first, it also introduces potential confusion for entries that have their publication dates set a second time for republishing and then not triggering a second notification.

So if we wanted to handle that case properly we would need to introduce additional logic elsewhere. It doesn't seem worth the hassle right now.

Daverball:
Small caveat to my suggestion: Since the existing loop also includes recently unpublished objects you will need to check `obj.published` in addition to `isinstance(obj, ExtendedDirectory)`.

Tschuppi81:
> Since cronjobs always run at specific intervals there's no chance of triggering too many notifications (there is however a small chance of skipping one during an upgrade)
Yes. Also we could miss sending out notifications in cases where an instance is down for > 1h

I will work on the approach with no metadata

Daverball:
> Yes. Also we could miss sending out notifications in cases where an instance is down for > 1h

If we wanted to address this I think my preference would be to store on `Organisation` when this cronjob was last run. That's a lot less book-keeping for similar robustness.

So here:
https://github.com/OneGov/onegov-cloud/blob/f67e97d4ed357dccaee3008bc58544504b22b02d/src/onegov/org/cronjobs.py#L113-L114

Instead of `then` being static you would change it to

```python
    now = utcnow()
    then = request.app.org.meta.get('hourly_maintenance_tasks_last_run', now - timedelta(hours=1))
```

and at the end of the function you do

```python
    request.app.org.meta['hourly_maintenance_tasks_last_run'] = now
```

This also avoids potentially dropping anything between the small second gaps between cronjob runs due to jitter.

Tschuppi81:
I got it integrated

PR #1655, opened 2025-01-17
Adds cronjob to poll bounce stats daily

Tschuppi81:
Org: Collect inactive email addresses daily and indicate delivery failures for newsletter recipients

TYPE: Feature
LINK: ogc-1896

## Checklist


Tschuppi81:
@Daverball Can you please have an eye on how I configure/use the postmark token? there is a related pr in puppet

pyproject.toml
```diff
@@ -72,13 +72,10 @@ exclude_dirs = [
 skips = [
     # ignore asserts
     "B101",
-    # ignore lxml.etree.fromstring (we may want to revisit this in the future)
```
Tschuppi81, line 75: The two rules got removed in bandit version 1.8.1, https://github.com/PyCQA/bandit/releases

Daverball:
> @Daverball Can you please have an eye on how I configure/use the postmark token? there is a related pr in puppet

I would use the existing postmark mail queue configuration, rather than define a new configuration key. We already have the token for instances that can send their mail through postmark, so no need to define it a second time.

Daverball:
@Tschuppi81 Basically all you need is probably:

```python
if self.mail['transactional']['mailer'] == 'postmark':
    postmark_token = self.mail['transactional']['token']
```

But you would also need to check the `marketing` configuration, since they can use two different mailer configurations.

tests/onegov/org/test_cronjobs.py
```diff
@@ -1285,160 +1286,54 @@ def count_occurrences():
         assert count_occurrences() == 0
 
 
-def test_send_email_notification_for_recent_directory_entry_publications(
```
Daverball, line 1288: Why did you remove this test? Is this a merge conflict resolution mistake?

Tschuppi81: Yes, accidentially removed

src/onegov/org/app.py
```diff
@@ -416,6 +416,15 @@ def redirect_after_login(
...
         return URL(request.link(dashboard)).path()
 
+    def get_postmark_token(self) -> str:
+        if self.mail:
+            mailer = self.mail.get('transactional', {}).get('mailer', None)
+            if mailer == 'postmark':
+                token = self.mail.get('transactional', {}).get('token', '')
+                return token
+
+        return ''
```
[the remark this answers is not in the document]
Tschuppi81: ok, will move the logic into the cronjob

onegov.yml.example
```diff
@@ -112,6 +112,9 @@ configuration: &global-config
       sender: newsletters@onegovcloud.ch
       <<: *local_smtp
 
+    # Postmark token for email bounce statistics cronjob
+    postmark_token: abcd
+
```
Daverball, line 117:
```suggestion
```

onegov.yml.example
```diff
@@ -29,6 +29,12 @@ mail_queues:
     username:
     password:
 
+  # postmark: &postmark
+    # mailer: postmark
+    # token: abcd
+    # NOTE: Do not define folder otherwise emails will be sent to customers and
+    # subscribers
+
```
Daverball, line 37: @Tschuppi81 We already have this example, see line 15. Also without a directory you will cause exceptions when you try to send mails.

Mails will only be sent out if you have a productive token in your config and call `onegov-core sendmail postgres`.

Tschuppi81: Correct, gonna remove it

PR #1663, opened 2025-01-22
Ogc 2003 new document form

BreathingFlesh:

## Commit message

Org: Document forms

Upload a PDF to be displayed in the forms view.

TYPE: Feature
LINK: OGC-2003

## Checklist

- [x] I have updated the election_day API docs


src/onegov/org/forms/document_form.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import annotations
+from onegov.core.utils import normalize_for_url
```
Daverball, line 2: nit. future import spacing.
```suggestion
from __future__ import annotations

from onegov.core.utils import normalize_for_url
```

src/onegov/org/models/document_form.py
```diff
@@ -0,0 +1,146 @@
+from __future__ import annotations
+from uuid import uuid4
```
Daverball, line 2:
```suggestion
from __future__ import annotations

from uuid import uuid4
```

src/onegov/org/models/document_form.py
```diff
@@ -0,0 +1,146 @@
...
+from sqlalchemy import Column, Text
+
+
+from typing import Any, TYPE_CHECKING
+if TYPE_CHECKING:
+    import uuid
+    from sqlalchemy.orm import Query, Session
+
+
+class DocumentFormFile(File, SearchableFile):
```
Daverball, line 27: It might be better to include the file contents in the search data for `FormDocument` so you don't end up with separate search results for the PDF and the Form.

Add something like:
```python
    pdf_extract: dict_property[str | None] = content_property()

    @observes('pdf')
    def pdf_observer(self, pdf: DocumentFormFile | None):
        self.pdf_extract = pdf.extract if pdf is not None else None
```

to `FormDocument` and add `pdf_extract` to the `es_properties`.

You could probably also get away without duplicating the extract from the file to the document, but it would be slower during indexing, so it's probably worth duplicating. Just add a test to make sure the extract remains synced if the file gets replaced.

Daverball (changes requested):
Looks good, but I think we can optimize the search results for this kind of form.

src/onegov/org/models/document_form.py
```diff
@@ -0,0 +1,146 @@
...
+        default=uuid4
+    )
+
+    #: A nice id for the url, readable by humans
+    name: Column[str] = Column(Text, nullable=False, unique=True)
+
+    title: Column[str] = Column(Text, nullable=False)
+
+    #: Describes the document briefly
+    lead: dict_property[str | None] = content_property()
```
Daverball, line 61: This one should probably be a meta property
```suggestion
    lead: dict_property[str | None] = meta_property()
```

For pages we care less, since we cache them in a different format for things like navigation. But for other things we want everything we'd use in a list view to be a `meta_property`, rather than a `content_property`.

PR #1665, opened 2025-01-24
Ensure function in case newsletter categories are not enabled

Tschuppi81:
Newsletter: Adds option to notify admins if a user unsubscribes from the newsletter subscription list

TYPE: Feature
LINK: ogc-2037

## Checklist

- [x] I made changes/features for org and town6
    - [x] I have tested sending mails

### DIFF
--- a/src/onegov/org/forms/settings.py
+++ b/src/onegov/org/forms/settings.py
@@ -1165,8 +1165,8 @@ class NewsletterSettingsForm(Form):
 
     notify_on_unsubscription = ChosenSelectMultipleEmailField(
         label=_('Notify on newsletter unsubscription'),
-        description=_('Send an email notification when a user unsubscribes '
-                      'from the newsletter'),
+        description=_('Send an email notification to the following users '
+                      'when a recipient unsubscribes from the newsletter'),
         validators=[StrictOptional()],
         choices=[]
     )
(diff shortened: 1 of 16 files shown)


src/onegov/org/forms/settings.py
```diff
@@ -1161,6 +1163,14 @@ class NewsletterSettingsForm(Form):
         },
     )
 
+    notify_on_unsubscription = ChosenSelectMultipleEmailField(
+        label=_('Notify on newsletter unsubscription'),
+        description=_('Send an email notification when a user unsubscribes '
```
BreathingFlesh, line 1168: I would write "send an email notification to the following users when a recipient unsubscribes"

src/onegov/org/views/subscription.py
```diff
@@ -56,7 +59,27 @@ def view_unsubscribe(
...
             mapping={'address': address}
         ))
+
         log.debug(f'Unsubscribed: {address}')
+
+        # check if admin wants an email for each unsubscription
+        receivers = request.app.org.notify_on_unsubscription or None
+        if receivers:
+            subject = _(
+                'Unsubscription from the newsletter at ${address}',
```
BreathingFlesh, line 69: This isn't translated yet.

src/onegov/town6/templates/mail_notify_unsubscribe.pt
```diff
@@ -0,0 +1,16 @@
+<div metal:use-macro="layout.base" i18n:domain="onegov.town6">
+    <tal:b metal:fill-slot="title">
+        ${title}
+    </tal:b>
+    <tal:b metal:fill-slot="body">
+        <p i18n:translate>Hi Admin,</p>
```
BreathingFlesh, line 6: The user doesn't have to be an admin, right? Maybe use the username or just "hello".

Tschuppi81: I only allowed admins at first. Now, admins and editors can be choosen

src/onegov/org/views/subscription.py
Tschuppi81: Done

PR #1680, opened 2025-02-06
Fix grouping by letter in order to show all entry notification subscr…

Tschuppi81:
Directory: Ensure all Directory Entry Subscribers are listed

TYPE: Bugfix
LINK: ogc-2063


cyrillkuettel (approved):
It seems that the `recipients` were not sorted before being passed into `groupby`, which `groupby` assumes.

PR #1690, opened 2025-02-14
Make the `--clean` option significantly faster.

cyrillkuettel:
No more waiting 5-10min till the script finishes :+1:


src/onegov/agency/cli.py
```diff
@@ -233,21 +236,62 @@ def import_lu_data_files(
...
+                        f'Elasticsearch index {index_pattern} removed',
+                        fg='green',
+                    )
+                except SearchOfflineError:
+                    click.secho(
+                        f'Warning: Elasticsearch is offline, '
+                        f'could not delete index {index_pattern}', fg='yellow',
+                    )
+                except Exception:
+                    click.secho('Unknown Error', fg='red')
```
Daverball, line 292:
```suggestion
            if app.es_client is not None:
                es_client = app.es_client
                hostname = app.es_indexer.hostname

                for type_name in es_type_names:
                    index_pattern = f'{hostname}-{schema}-*-{type_name}'
                    try:
                        es_client.indices.delete(
                            index=index_pattern, ignore_unavailable=True
                        )
                        click.secho(
                            f'Elasticsearch index {index_pattern} removed',
                            fg='green',
                        )
                    except SearchOfflineError:
                        click.secho(
                            f'Warning: Elasticsearch is offline, '
                            f'could not delete index {index_pattern}', fg='yellow',
                        )
                    except Exception:
                        click.secho('Unknown Error', fg='red')
```

Daverball (approved):
Just a small nit, guard against missing `es_client`, since that depends on the configuration. Otherwise looks good.

cyrillkuettel:
Just tested locally, modified it so only the cleaning part is run. No data is removed if i run it:
```
onegov-agency --select /onegov_agency/lu import-lu-data /home/cyrill/Downloads/februar/export_2025-02-10-korrigiert.csv --clean
```

Even if I run it as `postgres` user

```
sudo -u postgres /home/cyrill/onegov-cloud/venv/bin/python -m onegov.agency.cli --select "/onegov_agency/lu" import-lu-data /home/cyrill/Downloads/februar/export_2025-02-10-korrigiert.csv --clean 2> error.log
```

Running with `psql` directly works every time:
```
sudo -u postgres psql onegov-cloud
set search_path to 'onegov_agency-lu';
TRUNCATE TABLE agencies CASCADE;
TRUNCATE TABLE people CASCADE;
exit;
```

I'm not quite sure what it is.
It might be a problem with my local postgres configuration.

Daverball:
I just noticed that you're manually setting the search path in the CLI, that should not be necessary, the session is already scoped to the schema. Other than that you might need to manually call `session.flush`, since `session.execute` doesn't necessarily hook into the transaction machinery.

If that doesn't work you may need to add a `COMMIT;` line a the end of the delete query.

Daverball:
Another option may be to do this at a slightly higher level by disabling the indexer and then generate a regular bulk `DELETE` query using the ORM.

cyrillkuettel:
Okay nice, now it worked after adding `COMMIT;`.

> that should not be necessary, the session is already scoped to the schema.

Yes, I'm just a little paranoid.


> Another option may be to do this at a slightly higher level by disabling the indexer and then generate a regular bulk DELETE query using the ORM.

That sounds like a better, simpler way.

Side note: That I had to use `COMMIT` also leaves me to wonder, how the [feriennet/cli.py](https://github.com/OneGov/onegov-cloud/blob/6b2a21b1d0ab1e81a4492c7d72ef94fb7d0b03e1/src/onegov/feriennet/cli.py#L45-L68) which also usess SQL directly, does not seem to require `COMMIT` in the text block.

Daverball:
> Side note: That I had to use `COMMIT` also leaves me to wonder, how the [feriennet/cli.py](https://github.com/OneGov/onegov-cloud/blob/6b2a21b1d0ab1e81a4492c7d72ef94fb7d0b03e1/src/onegov/feriennet/cli.py#L45-L68) which also usess SQL directly, does not seem to require `COMMIT` in the text block.

The final `session.delete` ensures the regular transaction machinery will perform a `COMMIT` at the end of the transaction.

cyrillkuettel:
> The final `session.delete` ensures the regular transaction machinery will perform a `COMMIT` at the end of the transaction.

Excellent point, that's how it works!
So now the linting is the last thing, bandit has a new `Markupsafe` rule which impacts us in a lot of places.

Daverball:
> Excellent point, that's how it works! So now the linting is the last thing, bandit has a new `Markupsafe` rule which impacts us in a lot of places.

That's already fixed upstream, so you just need to merge the current master.

PR #1711, opened 2025-02-28
Fix/Allow to copy news article into a topic

Tschuppi81:
Org: Don't allow to paste topic into news and vice versa

TYPE: Feature
LINK: ogc-2105


Tschuppi81:
As the test fails I am unsure if my fix broke something. Can you comment on it please?

Tschuppi81:
@cyrillkuettel The implemented behavior shall not be possible, see comment in ticket. I will rework it to display an info/error instead.

Tschuppi81:
Pasting a topic into a news is not possible anymore, the paste link is disabled.
Pasting a news into a topic in also not possible anymore.

BreathingFlesh (approved):

Tschuppi81:
Also added a tooltip when pasting is not allowed

PR #1717, opened 2025-03-04
Fix typo in remove lu external id db upgrade

Tschuppi81:
Ticket: Make db upgrade for ticket closed_on column more performant

TYPE: Feature
LINK: None

### DIFF
--- a/src/onegov/ticket/upgrade.py
+++ b/src/onegov/ticket/upgrade.py
@@ -4,13 +4,10 @@ upgraded on the server. See :class:`onegov.core.upgrade.upgrade_task`.
 """
 from __future__ import annotations
 
-from datetime import timedelta
-
 from onegov.core.orm.types import JSON, UTCDateTime
 from onegov.core.upgrade import upgrade_task
 from onegov.ticket import Ticket
-from sqlalchemy import Boolean, Column, Integer, Text, Enum
-
+from sqlalchemy import Boolean, Column, Integer, Text, Enum, update, func, and_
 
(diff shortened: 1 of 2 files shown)


Daverball (changes requested):
Looks good apart from the fixed task requiring a new name

src/onegov/ticket/upgrade.py
```diff
@@ -157,13 +154,16 @@ def add_closed_on_column_to_ticket(context: UpgradeContext) -> None:
...
+        and_(
+            Ticket.state.in_(('closed', 'archived')),
+            Ticket.reaction_time.isnot(None),
+            Ticket.process_time.isnot(None)
+        )
+    ).values(
+        closed_on=Ticket.created +
+                  func.make_interval(secs=Ticket.reaction_time) +
+                  func.make_interval(secs=Ticket.process_time)
+    )
```
Daverball, line 167: This might actually only work in SQLAlchemy 2.0, I think the `select` API only worked for table objects in SQLAlchemy 1.3/1.4

So you would need to replace all instances of `Ticket` with `Ticket.__table__` and for the columns you would need to insert an additional `.c.` in between, so e.g. `Ticket.__table__.c.state`

cyrillkuettel: Locally it ran through without issues though, e.g. `onegov-core --select '/onegov_org/*' upgrade`

Daverball (approved):

cyrillkuettel (approved):

PR #1719, opened 2025-03-05
Firebase bugfixes

cyrillkuettel:


src/onegov/org/upgrade.py
```diff
@@ -391,3 +392,41 @@ def remove_stored_contact_html_and_opening_hours_html(
...
+        AND constraint_type = 'FOREIGN KEY'
+        AND constraint_name = :constraint
+    """)
+
+    result = context.session.execute(query, {
+        'schema': schema_name,
+        'constraint': constraint_name
+    })
+
+    if result.scalar():
```
Daverball, line 419: Consider refactoring this into a utility function `has_constraint`, analogous to `has_enum` on `UpgradeContext`. You can change `has_enum` to use query interpolation, rather than an f-string, so it's no longer vulnerable against SQL injection.

src/onegov/core/upgrade.py
```diff
@@ -482,17 +482,45 @@ def has_column(self, table: str, column: str) -> bool:
...
+            """
+            SELECT constraint_name
+            FROM information_schema.table_constraints
+            WHERE table_schema = :schema
+            AND table_name = :table_name
+            AND constraint_type = 'FOREIGN KEY'
+            AND constraint_name = :constraint_name
+            """
+        ).bindparams(
+            bindparam('schema', schema_name),
```
Daverball, line 501:
```suggestion
        query = text(
            """
            SELECT constraint_name
            FROM information_schema.table_constraints
            WHERE table_schema = :schema
            AND table_name = :table_name
            AND constraint_type = 'FOREIGN KEY'
            AND constraint_name = :constraint_name
            """
        ).bindparams(
            bindparam('schema', self.schema),
```

src/onegov/core/upgrade.py
```diff
@@ -482,17 +482,45 @@ def has_column(self, table: str, column: str) -> bool:
...
+                WHERE typname = :enum_name
+                  AND typnamespace = (
                     SELECT oid FROM pg_namespace
-                    WHERE nspname = '{self.schema}'
+                    WHERE nspname = :schema
                   )
             )
-        """).scalar()
+        """).bindparams(
+            bindparam('schema', self.app.schema),
```
Daverball, line 519:
```suggestion
            bindparam('schema', self.schema),
```

src/onegov/core/upgrade.py
```diff
@@ -482,17 +482,45 @@ def has_column(self, table: str, column: str) -> bool:
...
+            bindparam('enum_name', enum_name)
+        )
+        result = self.session.execute(query)
+        return result.scalar() is not None
```
Daverball, line 523:
```suggestion
        return result.scalar()
```
`EXISTS` subqueries always return a boolean, so this check is wrong. You might want to consider using an `EXISTS` subquery for the other helper function as well, so you can just directly return `result.scalar()`

Daverball (commented):
Just some nits. `UpgradeContext` has direct access to the schema and a small regression in `has_enum`.

src/onegov/core/upgrade.py
```diff
@@ -482,17 +482,35 @@ def has_column(self, table: str, column: str) -> bool:
...
 
-    def has_enum(self, enum: str) -> bool:
-        return self.session.execute(f"""
+    def has_constraint(self, table_name: str, constraint_name: str) -> bool:
+        return self.session.execute(text("""
+            SELECT EXISTS (
+                SELECT 1 FROM information_schema.table_constraints
+                WHERE table_schema = :schema
+                  AND table_name = :table_name
+                  AND constraint_type = 'FOREIGN KEY'
```
Daverball, line 491:
```suggestion
                  AND constraint_type = 'FOREIGN KEY'
```
This condition I'd either take out or make it an argument to the function, since we may want to check other kinds of constraints, like `UNIQUE` constraints.

cyrillkuettel:
image
Testing it. Ran through.


Another note, there might also be some way to deducde the name: 'push_notification_news_id_fkey` 
e.g. `{table_name}_{column_name}_{variable}`, but I guess this should suffice for now.

cyrillkuettel:
I have also written a browser test (almost done) which will test things whole thing with all the edge cases. Also then we finally have some way to directly test the `many-js`link interaction.

cyrillkuettel:
Ready for merge.

src/onegov/org/cronjobs.py
```diff
@@ -826,29 +827,34 @@ def delete_unconfirmed_newsletter_subscriptions(request: OrgRequest) -> None:
...
-    ))
     only_public_with_send_push_notification = only_public_news.filter(
         News.meta['send_push_notifications_to_app'].astext == 'true'
     )
-
     return only_public_with_send_push_notification
 
 
-@OrgApp.cronjob(hour='*', minute='*/10', timezone='UTC')
+@OrgApp.cronjob(hour='*', minute='*/10', timezone='Europe/Zurich')
```
Daverball, line 857: I would stick to `UTC` here, since it runs every 10 minutes anyways. Less likely to cause issues on days where we switch to or from daylight savings time.

src/onegov/org/cronjobs.py
```diff
@@ -826,29 +827,34 @@ def delete_unconfirmed_newsletter_subscriptions(request: OrgRequest) -> None:
...
+    # in UTC
     now = utcnow()
-    ten_minutes_ago = now - timedelta(minutes=10)
 
     # Get all news items that should trigger push notifications
     query = session.query(News)
-    query = query.filter(News.published == True)
+    query = query.filter(News.published.is_(True))
     query = query.filter(News.publication_start <= now)
+    query = query.filter(News.publication_start.isnot(None))
```
Daverball, line 838:
```suggestion
```

I'm pretty sure this condition is redundant, since the comparison above already gets rid of rows with a NULL value.

src/onegov/org/cronjobs.py
```diff
@@ -826,29 +827,34 @@ def delete_unconfirmed_newsletter_subscriptions(request: OrgRequest) -> None:
...
+
+    news_with_sent_notifications = session.query(
+        PushNotification.news_id
+    ).subquery()
+    query = query.filter(~News.id.in_(news_with_sent_notifications))
+    only_public_news = query.filter(
+        or_(
+            News.meta['access'].astext == 'public',
+            News.meta['access'].astext == None,
```
Daverball, line 847:
```suggestion
            News.meta['access'].is_(None),
```

Should be equivalent and more succinct

src/onegov/org/forms/extensions.py
```diff
@@ -19,7 +19,8 @@
 from wtforms.fields import TextAreaField
 from wtforms.validators import DataRequired, InputRequired, ValidationError
 
-from typing import TypeVar, TYPE_CHECKING
+from typing import TypeVar, TYPE_CHECKING, Any
+
```
Daverball, line 23:
```suggestion
from typing import Any, TypeVar, TYPE_CHECKING
```
Minor nit

src/onegov/org/forms/extensions.py
```diff
@@ -298,25 +303,29 @@ class PushNotificationFormExtension(FormExtension[FormT], name='publish'):
...
+                    render_kw={'disabled': 'disabled'},
+                )
+                self.push_notifications = MultiCheckboxField(
+                    label=('Topics'),
+                    choices=[],
+                    fieldset=_('Publication'),
+                    depends_on=('send_push_notifications_to_app', 'y'),
+                    validators=[StrictOptional()],
+                    render_kw={'class_': 'indent-form-field'},
+                )
```
Daverball, line 325: Generally we go the other way, we always add the fields, and then in `on_request` we delete them, if we don't need them. I would suggest doing it that way for consistency.

Or better yet, in this case you can probably just do this at the top of the function:

```python
if not self.request.app.org.firebase_adminsdk_credential:
    return self.form_class
```

Then you don't need any branches below at all.

src/onegov/org/models/push_notification.py
```diff
@@ -34,20 +40,20 @@ class PushNotification(Base):
...
         Integer,
-        ForeignKey('pages.id'),
+        ForeignKey('pages.id', ondelete='CASCADE'),
         nullable=False
     )
-
     news: relationship[News] = relationship(
-        'News', backref='sent_notifications',
+        'News',
+        backref=backref('sent_notifications', cascade='all, delete-orphan'),
```
Daverball, line 48:
```suggestion
        backref=backref('sent_notifications', passive_deletes=True),
```

This should be more efficient,  and work just as well with `ondelete='CASCADE'` on the foreign key.

Daverball (approved):
Looks good I mostly have some nits and small enhancement suggestions

src/onegov/org/models/push_notification.py
cyrillkuettel: Ok, would this necessitate an upgrade task, or just changing here?

Daverball: This is pure configuration, so no upgrade task necessary, you already added the cascade to the foreign key in an upgrade task.

src/onegov/org/forms/extensions.py
cyrillkuettel: I think that is what I tried before, this will not work

```python
    def create(self, timezone: str = 'Europe/Zurich') -> type[FormT]:

>       if not self.request.app.org.firebase_adminsdk_credential:
E       AttributeError: 'PushNotificationFormExtension' object has no attribute 'request'
```

Daverball: Oh right, `self` refers to the the form in that  context, not the extension.

src/onegov/org/forms/extensions.py
```diff
@@ -298,25 +303,33 @@ class PushNotificationFormExtension(FormExtension[FormT], name='publish'):
...
             def on_request(self) -> None:
-                if not self.request.app.org.meta.get(
-                        'firebase_adminsdk_credential'
-                ):
+
+                if not self.request.app.org.firebase_adminsdk_credential:
+                    if 'send_push_notifications_to_app' in self._fields:
+                        self._fields.pop('send_push_notifications_to_app')
+                    if 'push_notifications' in self._fields:
+                        self._fields.pop('push_notifications')
```
Daverball, line 329:
```suggestion
                    self.delete_field('send_push_notifications_to_app')
                    self.delete_field('push_notifications')
```

cyrillkuettel:
```
FAILED tests/onegov/org/test_extensions.py::test_people_shown_on_main_page_extension - AssertionError: You can only follow redirect responses (not 200 OK)
FAILED tests/onegov/org/test_views_map.py::test_map_set_marker - AssertionError: You can only follow redirect responses (not 200 OK)
FAILED tests/onegov/org/test_views_news.py::test_news - AssertionError: You can only follow redirect responses (not 200 OK)
```
I have seen this sort of thing before, I think replacing `follow()` with maybe_follow` resolves it. This however seems to be only symptomatic treatment, not the underlying root cause. It seems that somehow in the page edit form the user is no longer redirected but stays on the form. I will fix this then it should be good

PR #1724, opened 2025-03-07
Fix margin right for plus and minus button

Tschuppi81:
Town6: Fix margin for plus and minus button

TYPE: bugfix
LINK: None


Tschuppi81:
The only location I found these buttons in use is when editing Topics

Tschuppi81:
Thanks for the commit

PR #1733, opened 2025-03-13
Feature/ogc 2128 newsletter categories remove org name

Tschuppi81:
Newsletter: Remove `org_name` from Newsletter Category Definition

TYPE: Feature
LINK: ogc-2128


src/onegov/org/forms/settings.py
```diff
@@ -1189,61 +1187,57 @@ def ensure_categories(self) -> bool | None:
                 )
                 return False
 
+            print('*** tschupre data', data)
```
BreathingFlesh, line 1190: Remove the print

PR #1747, opened 2025-03-21
Render  fields as markdown

Tschuppi81:
Directory: Render markdown in accordion layout

TYPE: Feature
LINK: ogc-2131

### DIFF
--- a/src/onegov/org/theme/styles/org.scss
+++ b/src/onegov/org/theme/styles/org.scss
@@ -122,6 +122,7 @@ h6,
 /*
     Disable hyphenation in e-mail links and make them breakable
 */
+
 a[href^="mailto:"] {
     @include prefix(hyphens, none);
     @include prefix(word-break, break-all);
@@ -3566,10 +3567,10 @@ button {
 
 }
 
-.accordion .accordion-navigation .field-display {
+.accordion .accordion-navigation dl {
     margin-bottom: 1rem;
 }
-.accordion .accordion-navigation .field-display dd a {
+.accordion .accordion-navigation dd a {
     padding: 0;
     display: inline;
 }
--- a/src/onegov/org/theme/styles/org.scss
+++ b/src/onegov/org/theme/styles/org.scss
@@ -3566,6 +3566,14 @@ button {
 
 }
 
+.accordion .accordion-navigation .field-display {
+    margin-bottom: 1rem;
+}
+.accordion .accordion-navigation .field-display dd a {
+    padding: 0;
+    display: inline;
+}
+
 /*
     Popup Menu
 */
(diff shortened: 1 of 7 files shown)


Daverball:
@Tschuppi81 Did you check whether the other view type uses markdown. It might use `to_html_ul`, which should be sufficient for this use-case (it'll also add the correct css class for bulleted lists). Full markdown support seems like overkill to me.

Daverball:
> @Tschuppi81 Did you check whether the other view type uses markdown? It might use `to_html_ul`, which should be sufficient for this use-case (it'll also add the correct css class for bulleted lists). Full markdown support seems like overkill to me.

Actually nevermind, this is default form display rendering. So it is probably a markdown field. I would suggest relying on the field display renderer for rendering values, rather than trying to roll your own custom display. Have a look at the `display_fields` macro. You will need some special handling for `UploadField`/`UploadMultipleField` but for the rest you should be able to rely on `form.render_display(field)`.

Tschuppi81:
Yes, I will try to reuse the `display_fields` macro

src/onegov/org/theme/styles/org.scss
```diff
@@ -3566,6 +3566,14 @@ button {
...
 }
 
+.accordion .accordion-navigation .field-display {
+    margin-bottom: 1rem;
+}
+.accordion .accordion-navigation .field-display dd a {
+    padding: 0;
+    display: inline;
+}
+
```
Daverball, line 3576: Why is there only a CSS change for Org, does Town6 not need one?

Daverball (approved):
LGTM.

I just have a small question regarding the CSS change.

src/onegov/org/theme/styles/org.scss
Tschuppi81: Fort town6 it was introduced with ogc-1634, https://github.com/OneGov/onegov-cloud/commit/ed93c7d4844cb522603fa8fd7a704904372f3ee0

Tschuppi81: I now found a way to make it independent of `.field-display`

Daverball: My problem is not with the rules you added or what elements they address.

I'm asking why you have custom CSS for the header and content of the accordion in Town6 and not Org, and why you have custom styling for the display of the fields in Org but not Town6.

If you're satisfied with how it looks in both styles I'm satisfied too.

I was merely puzzled by the difference between Org and Town6, since you've decided to manually style completely different elements in both cases. So I was asking whether or not that was intentional or if you accidentally forgot to sync the rules between Org and Town6.

PR #1758, opened 2025-03-28
Core: Make AdjacencyList use midpoint for insertion of new items.

cyrillkuettel:
## Commit message

Core: Make AdjacencyList use midpoint for insertion of new items.

It's essentially implementing a sparse ordering system where new 
items can be inserted between existing ones. This means we can
now insert news items without having to reorder everything.

TYPE: Performance
LINK: OGC-2134

## Before / After: Submitting a News
| Before | After |
|--------|-------|
| image | image |

## TODO

- [ ] clean up: there's technically no need for `sort_siblings` function anymore. Shall we remove it

## Checklist

### DIFF
--- a/src/onegov/org/cronjobs.py
+++ b/src/onegov/org/cronjobs.py
@@ -12,6 +12,7 @@ from itertools import groupby
 from onegov.chat.collections import ChatCollection
 from onegov.chat.models import Chat
 from onegov.core.orm import find_models, Base
+from onegov.core.orm.abstract import AdjacencyList
 from onegov.core.orm.mixins.publication import UTCPublicationMixin
 from onegov.core.templates import render_template
 from onegov.directory.collections.directory import EntryRecipientCollection
@@ -49,8 +50,10 @@ from onegov.org.models import TicketMessage, ExtendedDirectoryEntry
 from onegov.user import User, UserCollection
 from onegov.user.models import TAN
 from sedate import to_timezone, utcnow, align_date_to_day
-from sqlalchemy import and_, or_, func
+from sqlalchemy import and_, or_, func, text
 from sqlalchemy.orm import undefer
+from sqlalchemy.schema import Table
+from sqlalchemy.inspection import inspect
 from uuid import UUID
 from onegov.org.notification_service import (
     get_notification_service,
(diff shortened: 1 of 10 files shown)


src/onegov/org/cronjobs.py
```diff
@@ -1011,3 +1014,98 @@ def send_push_notifications_for_news(request: OrgRequest) -> None:
...
+                from
+                    "{table_name}"
+                where
+                    "parent_id" is not null
+            )
+            update "{table_name}"
+            set "order" = numbered_siblings.new_order
+            from numbered_siblings
+            where "{table_name}"."{pk_column_name}" =
+                numbered_siblings."{pk_column_name}";
```
Daverball, line 1094:
```suggestion
            WITH numbered_siblings AS (
                SELECT
                    "id",
                    ROW_NUMBER() OVER (
                        PARTITION BY "parent_id"
                        ORDER BY "order" ASC NULLS LAST
                    ) AS new_order
                FROM
                    "{table_name}"
                WHERE
                    "parent_id" IS NOT NULL
            )
            UPDATE "{table_name}"
            SET "order" = numbered_siblings.new_order
            FROM numbered_siblings
            WHERE "{table_name}"."id" =
                numbered_siblings."id";
```

Technically this can break if `table_name` contains a quote. But it should be fine. We can switch to `sqlalchemy.update` on the table/column object if it becomes an issue. There's probably also an escape function we could use to escape the table name (there definitely is in psycopg, but we might not want to use their API, since we may switch backends, so we should be backend agnostic where possible).

src/onegov/org/cronjobs.py
```diff
@@ -1011,3 +1014,98 @@ def send_push_notifications_for_news(request: OrgRequest) -> None:
...
+        mapper = inspect(model)
+        table = mapper.local_table  # Get the mapped table
+
+        # Basic sanity checks for the model's mapping
+        if (table is None or not mapper.primary_key or
+            len(mapper.primary_key) != 1):
+            continue
+
+        table_name = table.name
+        pk_column_name = mapper.primary_key[0].name
```
Daverball, line 1061:
```suggestion
```

This is always `id`, since this column is part of the abstract base class.

src/onegov/org/cronjobs.py
```diff
@@ -1011,3 +1014,98 @@ def send_push_notifications_for_news(request: OrgRequest) -> None:
...
+
+        try:
+            session.execute(update_sql)
+            processed_tables.add(table_name)
+            log.info(f"Successfully normalized 'order' in '{table_name}'.")
+        except Exception as e:
+            log.error(
+                f"Error normalizing 'order' in '{table_name}': {e}",
+                exc_info=True  # Log traceback for better debugging
+            )
```
Daverball, line 1106:
```suggestion
        except Exception:
            log.exception(f"Error normalizing 'order' in '{table_name}'")
```

src/onegov/org/cronjobs.py
```diff
@@ -1011,3 +1014,98 @@ def send_push_notifications_for_news(request: OrgRequest) -> None:
...
+            try:
+                session.rollback()
+            except Exception:
+                log.error(f"Error during rollback for table '{table_name}': "
+               "{rb_exc}")
```
Daverball, line 1111:
```suggestion
            except Exception:
                log.exception(f"Error during rollback for table '{table_name}'")
```

Daverball (approved):
Looks good, I just have some suggestions for simplifications/improvements.

PR #1764, opened 2025-04-01
Feature/ogc 2108 support phone number search 2

Tschuppi81:
Search: Support phone number search (Agency, Org, Town6)

TYPE: Feature
LINK: ogc-2108


src/onegov/agency/models/person.py
```diff
@@ -63,15 +63,17 @@ def phone_internal(self) -> str:
...
         result = [self.phone_internal]
         for number in (self.phone, self.phone_direct):
             if number:
                 number = number.replace(' ', '')
                 result.append(number)
+                result.append(number[-4:])
                 result.append(number[-7:])
                 result.append(number[-9:])
                 result.append('0' + number[-9:])
+                result.append('+41' + number[-9:])
```
Daverball, line 76: I would suggest using `phonenumbers` to normalize the phone number to the international and local format, rather than rely on something hacky and unreliable like prefixing `+41`/`0`.

Since the local format will have standardized number spacing you can then also split the number into its groups and choose to do some of what you are doing here, like including the last number group, or the last two groups in addition to the international and local format without spaces (although I would make sure that their combined length exceeds 3, there are numbers where the final group is a single digit).

I.e. something like the following could make sense
```python
try:
    number_obj = phonenumbers.parse(number, 'CH')
except NumberParseException:
    result.append(number.replace(' ', ''))
    continue
result.append(format_number(number_obj, PhoneNumberFormat.E164))
national = format_number(number_obj, PhoneNumberFormat.NATIONAL)
groups = national.split()
for idx in range(len(groups)):
    partial_number = ''.join(groups[idx:])
    if len(partial_number) > 3:
        result.append()
```

src/onegov/core/utils.py
```diff
@@ -1226,3 +1228,35 @@ def batched(
             return
 
         yield batch
+
+
+def generate_fts_phonenumbers(numbers: Iterable) -> list[str]:
```
Daverball, line 1233:
```suggestion
def generate_fts_phonenumbers(numbers: Iterable[str | None]) -> list[str]:
```

src/onegov/core/utils.py
```diff
@@ -1226,3 +1228,35 @@ def batched(
...
+            result.append(format_number(
+                parsed, PhoneNumberFormat.E164))
+
+            national = format_number(
+                parsed, PhoneNumberFormat.NATIONAL)
+            groups = national.split()
+            for idx in range(len(groups)):
+                partial = ''.join(groups[idx:])
+                if len(partial) > 3:
+                    result.append(partial)
```
Daverball, line 1260:
```suggestion
        if not number:
            continue

        try:
            parsed = parse(number, 'CH')
        except NumberParseException:
            # allow invalid phone number
            result.append(number.replace(' ', ''))
            continue

        result.append(format_number(
            parsed, PhoneNumberFormat.E164))

        national = format_number(
            parsed, PhoneNumberFormat.NATIONAL)
        groups = national.split()
        for idx in range(len(groups)):
            partial = ''.join(groups[idx:])
            if len(partial) > 3:
                result.append(partial)
```
Let's dedent this by one level

src/onegov/people/cli.py
```diff
@@ -167,3 +167,32 @@ def _import(request: CoreRequest, app: Framework) -> None:
...
+            ('Telefon direkt', 'phone_direct'),
+            ('Telefon intern', 'phone_internal'),
+        ]
+        for p in session.query(Person):
+            click.secho(f'{p.title}', fg='green')
+            for label, *prop in properties:
+                value = ', '.join(
+                    str(getattr(p, attr))
+                    for attr in prop
+                    if hasattr(p, attr) and getattr(p, attr) != '')
```
Daverball, line 192:
```suggestion
                    value
                    for attr in prop
                    if (raw_value := getattr(p, attr, None)) is not None
                    if (value := str(value))
```
I think we should check for `None`, unless you wanted ti be included as `"None"`?

Let's also reduce the number of times we try to access the same attribute via `hasattr`/`getattr`.

Daverball (commented):
Looks good, although I have a more fundamental question about this change:

IIRC we can configure certain fields on `Person` as private. What if some of those phone numbers, that we now include in the search index are marked as private?

Should we still include them? Technically we're not leaking the number directly, but you can still figure out some of the digits by what the search returns.

In the API we seem to have made the decision to ignore this setting at the moment, but we're excluding `phone_internal` from the API endpoint results, that might be a mistake as well.

Both of these things should probably depend on `hidden_people_fields` and exclude whichever fields are selected there from the public data.

Although for the search you could make the argument, that it should be included, since otherwise it would be impossible for logged in users, that can see this information, to use the hidden fields in their search term.

src/onegov/people/cli.py
Tschuppi81: Nice!

PR #1766, opened 2025-04-03
Returns the laste change date of news items using the  attribute to l…

Tschuppi81:
Search: Returns the last change date of news items using the `published_or_created` attribute to lower priority of older news items (in favor for topics)

TYPE: Feature
LINK: ogc-2180

Validated
- [x] org
- [x] town6, verified on two big customers


Tschuppi81:
News are `one time` information, important at the time of publishing it. We sort them after time of publishing.
Topics provide information valid for longer periods, may be changed/updated over time. We sort items after `last_change` property

Tschuppi81:
@Daverball how can i fix the linting error?

src/onegov/page/model.py
```diff
@@ -40,12 +41,12 @@ def root(self) -> Page: ...
...
+        published_or_created: Column[datetime]
     else:
         @hybrid_property
-        def published_or_created(self):
+        def published_or_created(self) -> datetime:
             return self.publication_start or self.created
 
         @published_or_created.expression  # type:ignore[no-redef]
-        def published_or_created(cls):
+        def published_or_created(cls) -> datetime:
```
Daverball, line 51:
```suggestion
        def published_or_created(cls) -> ColumnElement[datetime]:
```

Would be the correct annotation here, but it's in a block that's not evaluated by the type checker, so there isn't much of a point to adding the annotations here.

Daverball (approved):

PR #1771, opened 2025-04-04
Suppress vat settings for agencies

Tschuppi81:
Agency: Suppress VAT settings

TYPE: Feature
LINK: None


BreathingFlesh (commented):
In Feriennet we just hide the settings via css. I think both ways are fine but maybe we should do it the same way everywhere so it doesn't get confusing.

PR #1835, opened 2025-06-02
FIXME: Move format methods from Layout to utils

Tschuppi81:
Layout: Make formatter methods available in utils

With the formatters as utility functions it is not needed to initiate a `Layout` in order to use the formatters.

TYPE: Feature
LINK: None

### DIFF
--- a/src/onegov/core/layout.py
+++ b/src/onegov/core/layout.py
@@ -1,9 +1,6 @@
 from __future__ import annotations
 
 
-import arrow
-import babel.dates
-import babel.numbers
 import isodate
 import numbers
 import sedate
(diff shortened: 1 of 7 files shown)


src/onegov/core/layout.py
```diff
@@ -154,40 +151,8 @@ def csrf_protected_url(self, url: str) -> str:
...
-                fmt.replace('skeleton:', ''),
-                datetime=dt,
-                fuzzy=False,
-                locale=locale
-            )
-        elif hasattr(dt, 'hour'):
-            return babel.dates.format_datetime(dt, format=fmt, locale=locale)
-        else:
-            return babel.dates.format_date(dt, format=fmt, locale=locale)
+        fmt = getattr(self, f'{format}_format', format)
```
Daverball, line 154: I would move this part to `CoreRequest` as well, we should never manually specify a `fmt` string, it should always be one of our pre-configured format options. If one of the applications overwrites the formats in their specific layout class, you would need to instead overwrite them in their specific request class, but I'm not sure if we do overwrite them anywhere.

PR #1843, opened 2025-06-10
RIS light initial commit

Tschuppi81:
RIS: Models for RIS light

TYPE: Feature
LINK: ogc-2245

The drawing shows the new models that are not adopted from `PAS` to get a quick overview.

ris-light-class-diagram drawio


Tschuppi81:
I created the namespace `parliament`, as a tablename prefix I use  `par`

Daverball:
Make sure to add `onegov.parlament` to `setup.cfg`. I would probably also add an empty `upgrade.py` to the module and add it to the `onegov_upgrades` entry point in `setup.cfg`.

You could put the migration for renaming the tables in there, instead of putting it into PAS.

setup.cfg
```diff
@@ -296,7 +297,7 @@ console_scripts =
     onegov-gazette = onegov.gazette.cli:cli
     onegov-landsgemeinde = onegov.org.cli:cli
     onegov-org = onegov.org.cli:cli
-    onegov-people = onegov.people.cli:cli
+    onegov-parliament = onegov.parliament.cli:cli
```
Daverball, line 300:
```suggestion
    onegov-people = onegov.people.cli:cli
```
parliament doesn't have any CLI commands (yet), and you removed the people specific CLI commands.

src/onegov/parliament/upgrade.py
```diff
@@ -0,0 +1,60 @@
...
+        ('par_commission_memberships', 'membership_type'),
+        ('par_commissions', 'commission_type'),
+        ('par_legislative_periods', 'legislative_period_type'),
+        ('par_parliamentarian_roles', 'role_type'),
+        ('par_parliamentarians', 'parliamentarian_type'),
+        ('par_parliamentary_groups', 'group_type'),
+        ('par_parties', 'party_type'),
+    ):
+        if not context.has_column(table, type_name):
+            context.operations.add_column(
```
Daverball, line 57: Don't forget to pre-populate the polymorphic identity of existing rows with the PAS identity, otherwise the mapper will not be able to load the correct version of the class when you query those models.

Daverball (approved):
Looks good as a first version.

Although I'd check which models are truly `es_public = True`. I kind of would've assumed these are all internal views both for RIS and PAS.

Tschuppi81:
> I kind of would've assumed these are all internal views both for RIS and PAS.

As I understood and how the website https://www.stadtwil.ch/stadtparlament is, RIS will be fully `public`, while PAS is fully internal.

cyrillkuettel:
Everything is private in pas

```bash
cyrill@cyrills-thinkpad:~/onegov-cloud/src/onegov/pas/models$ rg es_public
legislative_period.py
22:    es_public = False

parliamentary_group.py
27:    es_public = False

settlement_run.py
25:    es_public = False

parliamentarian.py
65:    es_public = False

party.py
26:    es_public = False

commission.py
46:    es_public = False
` `

Tschuppi81:
@daverball `test_hierarchy` failes as `parliament/models/change.py` uses `TownRequest` in the `add` method to get the `current_user`. Options:
- move `current_user` to `CoreRequest`
- move `parliament` module to applications level
- other/better?

Daverball:
> @Daverball `test_hierarchy` failes as `parliament/models/change.py` uses `TownRequest` in the `add` method to get the `current_user`. Options:
> 
> * move `current_user` to `CoreRequest`
> * move `parliament` module to applications level
> * other/better?

It's unclear to me whether we even need this model for the RIS, if we don't, we can simply move it to PAS. Usually `add` is a method on the collection and you would manually have to pass in the user linked to the change.

But if we do want this model in RIS, the easiest short term solution would be to just move the method to the PAS subclass and deal with whether or not we need a method like that for the RIS later (it could easily be a utility function instead).

Tschuppi81:
> `chy` failes as `parliament/models/change.py` uses `TownRequest` in the `add` method to get the `current_user`. Options:
> 
>     * move `current_user` to `CoreRequest`
> 
>     * move `parliament` module to applications level
> 
>     * other/better?

I am also not sure if it will be used in RIS, as it's for informational purpose only. I will move it to the `PAS` Subclass as you suggested.

PR #1844, opened 2025-06-10
Feature/ogc 2315 formcode relax for multiple spaces in labels

Tschuppi81:
Formcode: Allow multiple spaces in formcode labels

TYPE: Feature
LINK: ogc-2315


src/onegov/form/parser/grammar.py
```diff
@@ -603,7 +603,7 @@ def marker_box(characters: str) -> ParserElement:
     #       If pricing wasn't optional this would be easy...
     label = Regex(
         # a sequence of words (that can't start with brackets)
-        rf'(?P<label>{text_without_re(characters + "()")}(?: ?{word_re})*?) *'
+        rf'(?P<label>{text_without_re(characters + "()")}(?: *{word_re})*?) *'
```
Daverball, line 606: In hindsight I'm not convinced this is right call, since all the other labels still require single spaces, so I think my other suggested change is more appropriate, it mirrors what would've happened previously, i.e. it cuts off anything it doesn't recognize as part of the label.

src/onegov/form/parser/grammar.py
```diff
@@ -603,7 +603,7 @@ def marker_box(characters: str) -> ParserElement:
...
         # followed by optional pricing or discount followed by end of line
         r'(?=$|[(] *-?[0-9]+(?:\.[0-9]+)? *(?:%|[A-Za-z]{3}!?) *[)] *$)'
```
Daverball, line 608:
```suggestion
        r'(?=$| *[(] *-?[0-9]+(?:\.[0-9]+)? *(?:%|[A-Za-z]{3}!?) *[)]|  )'
```
Let's try this instead. It will make one of the test cases fail where there's a label with a price and then something after the price, but that's fine, we'll just live with that behavior (i.e. it discards everything after the price information)

Daverball (changes requested):
While allowing more than one space in labels seems fine, it would introduce an inconsistency with the rest of formcode.

If we wanted to relax this restriction, we would need to make it for all labels, but I don't think we should do that until we have a more robust storage format, so we can allow syntax modification, without risking changing existing forms.

src/onegov/form/parser/grammar.py
Daverball: Hang on the first part will need to change slightly as well, I will make a new comment with the full suggested change

Daverball, line 608:
```suggestion
        rf'(?P<label>{text_without_re(characters + "()")}(?: *{word_re})*?)'
        # followed by optional pricing or discount followed by end of line or multiple spaces
        r'(?= *$|[(] *-?[0-9]+(?:\.[0-9]+)? *(?:%|[A-Za-z]{3}!?) *[)]|  )'
```

Daverball: This should be closer to what we had before, without being as disruptive.

Daverball:
> `test_discount` fails but I don't see why. It seems to cut off all non-discount text in parentheses..
> 
> ```
>         f = field.parseString("[ ] Discount (50%) (For Kids)")
>         assert f.type == 'checkbox'
> >       assert f.label == 'Discount (50%) (For Kids)'
> E       AssertionError: assert 'Discount' == 'Discount (50%) (For Kids)'
> E         
> E         - Discount (50%) (For Kids)
> E         + Discount
> ```

Yes, this is expected with the new Regex, just change the test to expect `Discount` for the label and a discount of 50%, rather than a long label with no discount.

PR #1910, opened 2025-07-17
Menu to add, edit and remove meetings

Tschuppi81:
RIS: Menu to add, edit and remove meetings

TYPE: Feature
LINK: ogc-2453

### DIFF
--- a/src/onegov/org/forms/meeting.py
+++ b/src/onegov/org/forms/meeting.py
@@ -1,24 +1,99 @@
 from __future__ import annotations
 
-from wtforms import DateTimeField
+from wtforms.validators import Optional, InputRequired
+from wtforms import StringField
 
 from onegov.form import Form
+from onegov.form.fields import TimezoneDateTimeField, ChosenSelectMultipleField
 from onegov.org.forms.fields import HtmlField
 from onegov.org import _
-from wtforms.validators import Optional
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from onegov.org.models import Meeting
 
 
 class MeetingForm(Form):
 
-    start_datetime = DateTimeField(
+    title = StringField(
+        label=_('Title'),
+        validators=[InputRequired()],
+    )
+
+    start_datetime = TimezoneDateTimeField(
+        timezone='Europe/Zurich',
         label=_('Start'),
         validators=[Optional()],
     )
 
+    end_datetime = TimezoneDateTimeField(
+        timezone='Europe/Zurich',
+        label=_('End'),
+        validators=[Optional()],
+    )
+
     address = HtmlField(
-        label=_('Portrait'),
+        label=_('Address'),
+        validators=[InputRequired()],
+        render_kw={'rows': 3}
     )
 
     description = HtmlField(
         label=_('Description'),
+        validators=[Optional()],
+        render_kw={'rows': 5}
     )
+
+    meeting_items = ChosenSelectMultipleField(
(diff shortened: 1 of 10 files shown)


src/onegov/org/forms/meeting.py
```diff
@@ -1,24 +1,99 @@
...
+            .order_by(MeetingItem.number, MeetingItem.title)
+            .all()
+        )
+        self.meeting_items.choices = [
+            (item.id.hex, item.display_name)
+            for item in meetings
+        ]
+
+    def populate_obj(self, obj: Meeting, *args, **kwargs) -> None:
+        from onegov.org.models import MeetingItem
```
Tschuppi81, line 67: @Daverball When I put this import to the top of the file I get a `circular import error` as the form imports the partially initialized model class. How can I deal with this?

Daverball: You can try importing from the submodule instead, i.e. `from onegov.org.models.meeting_item import MeetingItem`, since fewer modules have to be fully initialized that way.

src/onegov/org/forms/meeting.py
```diff
@@ -1,24 +1,102 @@
...
+            (item.id.hex, item.display_name)
+            for item in meetings
+        ]
+
+    def populate_obj(self, obj: object, *args: Any, **kwargs: Any) -> None:
+        from onegov.org.models import MeetingItem
+
+        super().populate_obj(obj, exclude={'meeting_items'})
+
+        meeting: Meeting = obj  # type:ignore[assignment]
```
Daverball, line 72:
```suggestion
    def populate_obj(  # type: ignore[override]
        self,
        obj: Meeting,  # type: ignore[override]
        exclude: Collection[str] | None = None,
        include: Collection[str] | None = None
    ) -> None:
        from onegov.org.models import MeetingItem

        super().populate_obj(obj, exclude={'meeting_items', *(exclude or ())})

        meeting = obj
```
It's slightly more robust to add the model to signature and use ignore the override errors, since you will still get an error if you pass in something other than a `Meeting` into that function this way.

src/onegov/org/forms/meeting.py
```diff
@@ -1,24 +1,102 @@
...
+                .filter(MeetingItem.id == current_id)
+                .one()
+            )
+            if item is not None:
+                meeting.meeting_items.remove(item)
+
+    def process_obj(self, obj: object) -> None:
+        super().process_obj(obj)
+
+        meeting: Meeting = obj  # type:ignore[assignment]
```
Daverball, line 99:
```suggestion
    def process_obj(self, obj: Meeting) -> None:   # type: ignore[override]
        super().process_obj(obj)

        meeting = obj
```
same thing here

src/onegov/town6/views/meeting.py
```diff
@@ -86,9 +141,67 @@ def view_meeting(
...
+
+    layout = MeetingLayout(self, request)
+
+    if form.submitted(request):
+        form.populate_obj(self)
+        request.success(_('Your changes were saved'))
+        return request.redirect(request.link(self))
+
+    form.process(obj=self)
+
```
Daverball, line 169:
```suggestion

```
While you did use `pass_model=True`, you forgot to remove the `form.process()` call, this can mess up some things that rely on `on_request()`. We should ideally get rid of every manual `form.process` call in edit views, but we can improve this iteratively.

I think that this should fix the weird behavior with file uploads you're experiencing.

PR #1929, opened 2025-07-25
Remove role check

Tschuppi81:
Directories: Publication seetings for new and edit directory entries independent of role

Same as for submittted directory entries

TYPE: Feature
LINK: xxx

### DIFF
--- a/src/onegov/org/views/directory.py
+++ b/src/onegov/org/views/directory.py
@@ -96,8 +96,7 @@ def get_directory_entry_form_class(
             if model.directory.enable_map == 'no':
                 self.delete_field('coordinates')
 
-            if (not model.directory.enable_publication and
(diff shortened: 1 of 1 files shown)


Tschuppi81:
Currently the behavior is not the same when submitting a directory entry vs creating/editing on as manager.

The submission does not depend on the role
```
    @property
    def extensions(self) -> tuple[str, ...]:
        extensions = ['coordinates', 'submitter', 'comment', 'publication']
        if self.enable_map == 'no':
            extensions.remove('coordinates')
        if not self.enable_publication:
            extensions.remove('publication')
        return tuple(extensions)
```

Tschuppi81:
I understand your arguments. 

Background of my change was a sentry issue if I remember correctly, but I didn't note it down...

PR #1933, opened 2025-07-25
Bugfix/swi 63 handle empty policy area

Tschuppi81:
SWI: Ignore empty policy area

TYPE: Bugfix
LINK: swi-63

## Checklist


Tschuppi81:
Is there a way to influence the converters to handle this case?

Daverball:
> Is there a way to influence the converters to handle this case?

Yes, you can write a custom converter for the policy area format, so any values that don't conform to the format get dropped.

Ideally we change the collection to accept `list[PolicyArea]` instead and write a converter for `PolicyArea` that verifies the regex `r'[0-9](\.[0-9])*'.

src/onegov/core/filters.py
Daverball, line 1: I think you should merge master, this looks like changes that happened a while ago on master, so they should not be included in the diff.

src/onegov/swissvotes/collections/votes.py
```diff
@@ -71,6 +71,10 @@ def __init__(
         sort_by: str | None = None,
         sort_order: str | None = None
     ) -> None:
+        # ignore empty policy areas swi-63
+        if policy_area and '' in policy_area:
+            policy_area = [i for i in policy_area if i]
+
```
Daverball, line 77: This check should go away if the converter is doing its job

src/onegov/swissvotes/converters.py
Daverball, line 1: This seems a little bit too complex for a converter, I would choose two simple converter functions like we do for things like `datetime_converter` and use the base class. Converter decode functions are allowed to raise exceptions IIRC, but you could also add simple `try catch` that returns `None` if splitting the components and turning them into integers fails.

The other issue is that you're converting from `str` to `str`. My idea was to convert to `PolicyArea` and move some of the validation into the `PolicyArea` constructor, this makes the API less error-prone to use in general, since the collection will now expect a `PolicyArea` instance rather than `str`, which has already been validated, with `str` it's still possible to pass in bogus data from somewhere else.

Daverball: I'm also not sure if this will actually work, since `Converter.__init__` will replace your `decode/encode` methods with whatever you passed in when constructing the instance, which is `str`.

Daverball (changes requested):
This is closer to what I suggested, but it's not quite there yet.

src/onegov/swissvotes/path.py
```diff
@@ -80,14 +82,17 @@ def get_votes(
     sort_by: str | None = None,
     sort_order: str | None = None
 ) -> SwissVoteCollection:
+    filtered_policy_area = [
+        p for p in policy_area if p
+    ] if policy_area else None
```
Daverball: I don't think this step is necessary. I'm pretty sure morepath will discard any values that convert to `None` for list converters. Please verify. If it doesn't do that it might be necessary for the converter to raise `ValueError` when it doesn't see a valid `PolicyArea`.

Daverball: @Tschuppi81 This is the only feedback I still have, other than this it looks good now.

Tschuppi81: Thanks, I overlooked your feedback

Tschuppi81: I now raise `ValueError` for invalid or empty policy areas. This simplified a few type annotations too

src/onegov/core/orm/mixins/content.py
```diff
@@ -410,7 +410,7 @@ def __init__(
             attribute,
             key,
             default,  # type:ignore[arg-type]
-            Markup  # type:ignore[arg-type]
+            Markup
```
Daverball, line 413: This should already be fixed on master, so best to merge master.

Tschuppi81: done

PR #1941, opened 2025-07-31
Bugfix/ogc ris political business filter

Tschuppi81:
RIS: Improve political business filtering

Sentry: https://seantis-gmbh.sentry.io/issues/6778531266/

TYPE: Bugfix
LINK: None

### DIFF
--- a/src/onegov/town6/views/political_business.py
+++ b/src/onegov/town6/views/political_business.py
@@ -108,7 +108,7 @@ def view_political_businesses(
             text=request.translate(text) +
                 f' ({count_per_status[status]})',
             active=status in self.status,
-            url=request.link(self.for_filter(s=status)),
+            url=request.link(self.for_filter(status=status)),
         )
         for status, text in POLITICAL_BUSINESS_STATUS.items()
         if (status in count_per_status and count_per_status[status] > 0)
--- a/src/onegov/org/models/political_business.py
+++ b/src/onegov/org/models/political_business.py
@@ -264,8 +264,8 @@ class PoliticalBusinessCollection(
         self,
         session: Session,
         page: int = 0,
-        status: PoliticalBusinessStatus | Sequence[str] | None = None,
-        types: PoliticalBusinessType | Sequence[str] | None = None,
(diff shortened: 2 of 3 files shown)


src/onegov/town6/views/political_business.py
```diff
@@ -108,7 +108,7 @@ def view_political_businesses(
             text=request.translate(text) +
                 f' ({count_per_status[status]})',
             active=status in self.status,
-            url=request.link(self.for_filter(s=status)),
+            url=request.link(self.for_filter(status=status)),
```
Daverball, line 111:
```suggestion
            url=request.link(self.for_filter(status=[status])),
```
I think this should be like this now.

Daverball (approved):
Looks good, there's just one small mistake.

And some general unrelated feedback for `view_political_businesses`, you create empty lists you don't use, and you call `sorted` on list comprehensions.

Both of those things are memory-inefficient, since you're immediately discarding those lists after creating them, so you're giving the garbage collector more work. Either call `sorted` on a generator expression or create the list and then call its `sort` method.

Also one nice thing about generators is that you can use as many `if` expressions as you like, so you don't need to use `and`, so you can write the first generator like this:

```python
    types = sorted((
        Link(
            text=request.translate(text) +
                 f' ({count_per_business_type[type]})',
            active=type in self.types,
            url=request.link(self.for_filter(type=type)),
        )
        for type, text in POLITICAL_BUSINESS_TYPE.items()
        if type in count_per_business_type
        if count_per_business_type[type] > 0
    ), key=lambda x: x.text.lower() if x.text else '')
```

Of course in this case it would probably would be more succinct to just do this instead:

```python
    types = sorted((
        Link(
            text=request.translate(text) +
                 f' ({count_per_business_type[type]})',
            active=type in self.types,
            url=request.link(self.for_filter(type=type)),
        )
        for type, text in POLITICAL_BUSINESS_TYPE.items()
        if count_per_business_type.get(type, 0) > 0
    ), key=lambda x: (x.text or '').lower())
```

With a membership test + retrieval you're usually doing more work than with a simple `get` call, since the hash table lookup happens twice.

PR #1949, opened 2025-08-06
Include commission membership in active filter and improve collection…

Tschuppi81:
RIS: Include commission memberships in Parliamentarians active property

TYPE: Feature
LINK: ogc-2245


Tschuppi81:
@cyrillkuettel Consider commission memberships to determine if a parliamentarian is active or not makes also sense for PAS, right

PR #2018, opened 2025-08-28
Feature/ris do no show access field

Tschuppi81:
RIS: Don't show access field for meeting and political business

All RIS information is public

TYPE: Feature
LINK: ogc-2245


Tschuppi81:
branch renamed https://github.com/OneGov/onegov-cloud/pull/2019

PR #2037, opened 2025-09-02
Ogc 2269 agenda item files

BreathingFlesh:

## Commit message

Town6: Agenda item files

File upload for additional files

TYPE: Feature
LINK: OGC-2269

## Checklist


src/onegov/landsgemeinde/locale/de_CH/LC_MESSAGES/onegov.landsgemeinde.po
```diff
@@ -351,6 +360,9 @@ msgstr ""
 msgid "Copied to Clipboard!"
 msgstr "In die Zwischenablage kopiert!"
 
+msgid "<dynamic element>"
+msgstr ""
+
```
Daverball, line 365: It seems like `macros.pt` contains a lot of unnecessary/incorrect `i18n:translate` attributes, so this bogus translation entry is probably a result from one of those. The attribute only really makes sense if the tag contains a static string or a static string with one or more tags marked with a `i18n:name` for dynamic substitutions.

This doesn't have anything to do with your change. But I'd get rid of this superfluous translation entry in this PR. Please make sure to fix the incorrect translation attributes in the macros in a follow-up PR.

src/onegov/landsgemeinde/models/agenda.py
```diff
@@ -149,3 +149,21 @@ def title_parts(self) -> list[str]:
...
         ]
+
+    @property
+    def more_files(self) -> list[LandsgemeindeFile]:
+        files = self.files
+        if self.memorial_pdf:
+            return [
+                file for file in files
+                if file != self.memorial_pdf and isinstance(file,
+                                                            LandsgemeindeFile)
```
Daverball, line 160:
```suggestion
                if file != self.memorial_pdf
                if isinstance(file, LandsgemeindeFile)
```

minor style nit. Comprehension may contain an arbitrary number of conditions, so using `and` is not necessary.

Although I'm not sure if the second condition is even necessary. If it is, you should probably check it in both branches. It doesn't really make sense to only check it if `memorial_pdf` exists.

It seems easier however to just change all the annotations from `LandsgemeindeFile` to `File` though. If you really wanted "type safety" (it could still be circumvented in the database itself), ensuring that only `LandsgmeindeFile` can be linked to `Assembly`, you should change the annotation of `Assembly.files` to `relationship[list[LandsgemeindeFile]]` and ignore the type error there.

It also seems more efficient to collapse the two branches by doing this:

```python
return [file for file in files if file.name != 'memorial_pdf']
```

Daverball (commented):
Looks good. Just an unrelated problem I noticed and a typing nit/suggestion.

PR #2048, opened 2025-09-08
Meeting items are no longer searchable

Tschuppi81:
RIS: Meeting items are no longer searchable

Prevents duplicates in search results

TYPE: Feature
LINK: ogc-2559


Tschuppi81:
if a meeting item is not a political business I can't be found anymore

PR #2097, opened 2025-09-25
Feature/ogc 2628 directory api

BreathingFlesh:

## Commit message

API: Add Directories to API endpoints

TYPE: Feature
LINK: OGC-2628

## Checklist


src/onegov/api/path.py
```diff
@@ -37,10 +39,12 @@ def get_api_endpoint(
...
         return AuthEndpoint(app)
 
-    cls = ApiEndpointCollection(app).endpoints.get(endpoint)
-    if not cls:
+    item = ApiEndpointCollection(request, app).endpoints.get(endpoint)
+    if not item:
         raise ApiException('Not found', status_code=404)
-    return cls(request, extra_parameters=extra_parameters, page=page)
+    item.page = page
+    item.extra_parameters = extra_parameters or {}
```
Daverball, line 46: I'm also not super happy about this, although the way we use it currently it seems harmless.

My preference would be something along the lines of:
- Keep the `endpoints` `cached_property`
- Add a new method `get_endpoint(self, name: str, page: int = 0, extra_parameters: dict[str, Any] | None = None) -> ApiEndpoint[Any] | None`
- Extend `config.setting_registry.api.endpoints` with the same optional parameters
- Use `get_endpoint` here instead of `endpoints.get`

This seems a little less messy and error-prone to me. Changing attributes on cached properties is always a little bit dangerous as soon as there is some concurrency or parallelism with a shared collection.

src/onegov/town6/api.py
```diff
@@ -215,31 +217,51 @@ def for_filter(self, **filters: Any) -> Self:
...
+                data = {f.name: f.object_data for f in item.content_fields}
+
+            for field in item.directory.fields:
+                if any(field_type in field.type for field_type in [
+                    'fileinput', 'url']):
+                    data.pop(field.id, None)
+                if any(field_type in field.type for field_type in [
+                    'date', 'datetime']):
+                    if data.get(field.id):
+                        data[field.id] = data[field.id].isoformat()
```
Daverball, line 247: Can we roll this into a single loop? You have access to the type of field in the comprehension above, filtering out the file fields can be the `if` part of the comprehension and for the value you can write a little helper function that formats `date`/`datetime` values.

src/onegov/town6/api.py
```diff
@@ -215,31 +217,51 @@ def for_filter(self, **filters: Any) -> Self:
...
 
     def item_links(self, item: ExtendedDirectoryEntry) -> dict[str, Any]:
-        data = {(f.note or 'file'): f for f in item.files.copy()}
+        data = {}
+        if item.access == 'public':
+            content_field_names = []
+            if item.content_fields:
+                content_field_names = [i.name for i in item.content_fields]
+            data = {(f.note or 'file'): f for f in item.files
+            if f.note and f.note.split(':')[0] in content_field_names}
```
Daverball, line 263:
```suggestion
            data = {
                (file.note or 'file'): file
                for file in item.files
                if file.note
                if file.note.split(':', 1)[0] in content_field_names
            }
```
The formatting made this a little hard to read, i like putting each part of the comprehension on a separate line when the conditions get complex enough. I also don't recall what kind of upload field we're using in formcode, there's a possibility the file is accessible from the field itself, which would simplify the code a bit.

I think you also forgot to include the url field values in these items.

PR #2347, opened 2026-02-17
Town6: Add API endpoints for forms, resources, people, meetings, political businesses and parliament

somehowchris:
## Commit message

Town6: Add API endpoints for forms, resources, people, meetings, political businesses and parliament

Registers 8 new API endpoints. Enriches form endpoint with text field for internal forms. Aligns PersonApiEndpoint with the agency model: 20 data fields, respects hidden_people_fields.

TYPE: Feature
LINK: -

## Checklist

### DIFF
--- a/src/onegov/town6/api.py
+++ b/src/onegov/town6/api.py
@@ -2,12 +2,27 @@ from __future__ import annotations
 
 from functools import cached_property
 from onegov.api.models import ApiEndpoint, ApiEndpointItem
+from onegov.core.collection import GenericCollection, Pagination
 from onegov.event.collections import OccurrenceCollection
+from onegov.form.models import FormDefinition
 from onegov.gis import Coordinates
 from onegov.org.models.directory import (
     ExtendedDirectory, ExtendedDirectoryEntry,
     ExtendedDirectoryEntryCollection)
+from onegov.org.models.external_link import (
+    ExternalFormLink, ExternalResourceLink)
+from onegov.org.models.meeting import Meeting, MeetingCollection
 from onegov.org.models.page import News, NewsCollection, Topic, TopicCollection
+from onegov.org.models.parliament import (
+    RISCommission, RISCommissionCollection,
+    RISParliamentarian, RISParliamentarianCollection,
+    RISParliamentaryGroup, RISParliamentaryGroupCollection)
+from onegov.org.models.political_business import (
+    PoliticalBusiness, PoliticalBusinessCollection)
+from onegov.people import Person
+from onegov.people.collections import PersonCollection
+from onegov.reservation.collection import ResourceCollection
+from onegov.reservation.models import Resource
 from onegov.town6 import _
 from sqlalchemy.exc import SQLAlchemyError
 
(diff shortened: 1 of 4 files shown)


src/onegov/town6/api.py
```diff
@@ -40,6 +56,218 @@ def get_modified_iso_format(item: TimestampMixin) -> str:
...
+    def previous(self) -> Self | None:
+        if self.page > 0:
+            return self.page_by_index(self.page - 1)
+        return None
+
+    @property
+    def next(self) -> Self | None:
+        if self.page < self.pages_count - 1:
+            return self.page_by_index(self.page + 1)
+        return None
```
Daverball: I think you should be able to get rid of most of this code by using the `Pagination` mixin and the original non-paginated collection class. You will run into some mypy errors, because `libres`/`onegov.reservation` models use their own base class. But that should go away once the SQLAlchemy 2.0 changes have been merged.

src/onegov/town6/api.py
```diff
@@ -300,3 +528,337 @@ def item_links(self, item: ExtendedDirectoryEntry) -> dict[str, Any]:
         data['html'] = item  # type: ignore
 
         return data
+
+
+FormOrExternalLink = FormDefinition | ExternalFormLink
+ResourceOrExternalLink = Resource | ExternalResourceLink
```
Daverball:
```suggestion
FormOrExternalLink: TypeAlias = FormDefinition | ExternalFormLink
ResourceOrExternalLink: TypeAlias = Resource | ExternalResourceLink
```
This way these type aliases will be automatically rewritten to `type` statements by Ruff, once we migrate to Python 3.12.

Daverball (changes requested):
Looks like a good start. I'm not happy with the handling of `ExternalLink` however, they should be properly mixed into the pagination, so we don't lie about our items per page and/or how many pages there are.

There is a more general problem with models that make use of `AccessExtension` and `UTCPublicationMixin` (for some models you will only find those on the `onegov.org` specific subclass, but you still need to handle them in that case), since some of these collections can contain non-public entries and there is a `mtan` access which is technically public, but you shouldn't be able to see anything beyond what is displayed on the listing view. Take a look at `NewsCollection`/`TopicCollection` for how to handle access/publication restrictions.

Daverball (changes requested):
This looks pretty good.

While I like the generic approach, there's a little bit too much wrapping and re-wrapping of collections going on, I think we can peel back at least one layer without too much code duplication.

You also wrapped at least one collection that could've been used as-is (apart from that collection missing a filter it should've had). I didn't check all of them, but if the source collection already is a pagination, you should be using that directly and extend it if necessary.

somehowchris:
@Daverball FYI^^

Daverball:
Since the merge was going to be quite messy with the changes that happened to the API in the meantime I took care of that for you.

I also did a small simplification pass over everything, to get rid of some of the complexity we don't need (yet).

I will finish up the PR in my branch, good work!

PR #2428, opened 2026-03-27
Bugfix/ogc 3003 fix migrate links

Tschuppi81:
Org: Fix migrate links tool
                                                                                                   
Use `re.subn` to count URL occurrences per field (not just whether a field changed), and wrap the
substituted value in `Markup` when the original was `Markup` to prevent double-escaping HTML 
entities. 

TYPE: Bugfix
LINK: ogc-3003


tests/onegov/town6/test_views_settings.py
```diff
@@ -165,3 +167,69 @@ def test_resource_settings(client: Client) -> None:
...
+    session = client.app.session()
+    request: Any = Bunch(**{
+        'session': session,
+        'identity.role': 'admin'
+    })
+    old_domain = 'foo.ch'
+
+    # create topic
+    topic = Topic(title='Foo Topic', name='foo-topic')
+    topic.text = '<p>Wow, https://foo.ch/abc is a great page!</p>'
```
Daverball:
```suggestion
    topic.text = Markup('<p>Wow, https://foo.ch/abc is a great page!</p>')
```
Make sure to use `Markup` here, otherwise `<p>` will get escaped to `&lt;p&gt;`

tests/onegov/town6/test_views_settings.py
```diff
@@ -165,3 +167,69 @@ def test_resource_settings(client: Client) -> None:
...
+
+    def get_topic_text() -> str:
+        t = TopicCollection(session).by_title('Foo Topic')
+        assert t is not None and t.text is not None
+        return str(t.text)
+
+    def get_news_text() -> str:
+        n = NewsCollection(request).by_title('Big News')
+        assert n is not None and n.text is not None
+        return str(n.text)
```
Daverball:
```suggestion
    def get_topic_text() -> Markup:
        t = TopicCollection(session).by_title('Foo Topic')
        assert t is not None and t.text is not None
        return t.text

    def get_news_text() -> Markup:
        n = NewsCollection(request).by_title('Big News')
        assert n is not None and n.text is not None
        return n.text
```

Daverball (changes requested):
I initially thought your fix was incorrect, but I misread parts of it. So the fix is fine, but the tests could be improved a little.

src/onegov/org/models/page.py
```diff
@@ -557,6 +560,9 @@ def __link_alias__(self) -> str:
             }
         )
 
+    def by_title(self, title: str) -> News | None:
+        return self.subset().filter(News.title == title).first()
+
```
Daverball: Since these are only used for tests in two places. I think you should just manually do the equivalent query in the tests, or rely on `by_path` instead, which is more reliable.

If you do want to keep these for convenience, then for real-world use I would probably not use `subset()` here, but `query(ordered=False)` instead, so it behaves the same as `by_id` and `by_path`. I would also add them to `AdjacencyListCollection`, since `AdjacencyList` always has a `title`, so it doesn't make sense that only these two specific collections would have this method, just remember that `title` is not unique, even `name` is only unique if the pages share the same parent, so it would also beg the question if there's a `by_title`, why not a `by_name` too?

Daverball (commented):
Looks good, but you should reconsider or clean up the addition of the `by_title` method.

PR #2441, opened 2026-04-13
Ogc management and settings

BreathingFlesh:

## Commit message

Org: Rearrange settings and management bar

- Added a link for User manual
- Added a new category "modules" to the global tools
- Rearranged the global tools and settings
- Added categories to settings

TYPE: Feature

## Checklist


src/onegov/feriennet/templates/layout.pt
```diff
@@ -94,14 +94,15 @@
         <main>
             <header id="header" tal:define="header_options request.app.org.header_options or {}">
                 <div id="sticky-header-area">
-                    <div id="globals" class="globals" tal:define="toolbox global_tools|False" tal:condition="request.is_logged_in">
+                    <div id="globals" class="globals" tal:define="toolbox global_tools|False;" tal:condition="request.is_logged_in">
                         <div class="grid-container">
                         <div class="grid-x grid-padding-x">
                             <div class="cell small-12">
                                 <div class="global-header-flex ${'' if toolbox else 'centered'}">
+                                    <div></div>
```
Daverball, line 102: Was this empty div added intentionally? I know there are some weird browser layouting quirks, that sometimes require workarounds like this, but I still wanted to point it out, just in case it was an accident.

src/onegov/feriennet/templates/layout.pt
```diff
@@ -111,7 +112,7 @@
                                         </tal:b>
                                         <tal:b condition="tool.url|tool.id|nothing" define="link tool">
                                         <tal:b replace="link(layout)" />
-                                        </tal:b>
+                                    </tal:b>
```
Daverball: The indentation here got a little messed up, I assume it's some auto-indenting thing in your editor. While I'm not a fan of leaving some nested blocks flat, since it makes it harder to see what's going on, it does make sense for deeply nested structures like this that would use up way too much horizontal space in the editor, although in this case it may me prudent to just extract this into a macro or even multiple macros, so you can work at a lower indentation level and properly indent everything.

src/onegov/org/forms/__init__.py
```diff
@@ -99,7 +99,7 @@
     'ManualInvoiceItemForm',
     'MapSettingsForm',
     'MeetingForm',
-    'ModuleSettingsForm',
+    'AccessSettingsForm',
```
Daverball: Try to keep these sorted (It would also be nice with the imports themselves, since they were sorted before, but a little less important there). There is a ruff rule with an auto-fix for this, we should probably enable. I think it just was a little too noisy, when I tried it last.

src/onegov/org/templates/layout.pt
```diff
@@ -82,14 +82,32 @@
...
+                    <h4 tal:condition="header_options['left_header_name']|nothing"><a
+                            href="${header_options.get('left_header_url', '#')}"
+                            style="color:${color or '#fff'}; font-size: ${rem_size}rem">
+                            ${header_options['left_header_name']}
+                    </a></h4>
+                </div>
+
+              </div>
+              <div class="columns small-12 medium-4">
+                <div></div>
```
Daverball, line 109: Since there's another one these, they're either intentional, or it's a quirk of your editor that makes it easy for them to accidentally be inserted.

src/onegov/org/views/settings.py
```diff
@@ -82,18 +89,47 @@ def query_settings() -> Iterator[dict[str, Any]]:
...
+                    setting['name'] == 'newsletter-settings'
+                    and not request.app.org.show_newsletter
+                ):
+                    continue
+
+                if (
+                    setting['name'] == 'ris-settings'
+                    and not request.app.org.ris_enabled | False
+                ):
+                    continue
```
Daverball, line 103: It might not be quite worth it yet, but if we see more conditions sneaking in like this, we may want to extract this into an optional `setting_visible` predicate that takes a callback (`lambda request: True` by default) that needs to return `True` in order for the setting to be visible.

But if you feel like cleaning this up a bit, please go ahead.

Furthermore in terms of refactors, that might be worth it: Defining a new directive for settings views, with its own registry (or rather making the one we didn't end up using work correctly). So we don't have to use quite as much work to retrieve all of the setting views. This could be a `CompositeAction` that yields a `HtmlHandleFormAction` and a new action that registers the setting or a subclass of `HtmlHandleFormAction` that adds the extra logic for registering the setting.

We could then also be a little bit more robust and emit an exception if we encounter a setting with a setting category we don't know about.

src/onegov/org/views/settings.py
```diff
@@ -49,6 +49,13 @@ class ApiKeyWithDeleteLink(ApiKey):
         delete_link: Link
 
 
+setting_categories = [
+    _('General'),
+    _('Modules'),
+    _('Advanced Settings'),
+]
```
Daverball, line 56: Hardcoding this seems a little inflexible and fragile, we could either register this as a setting on `OrgApp`, so other apps can have their own categories with their own order or collect all the possible categories from the settings themselves, the order of the categories would then be determined according to the minimum `order` value for all the settings in that category. With the latter you may end up accidentally reordering the categories or adding a new category, you didn't intend to, so I'm not sure it's a great idea, even if it makes it a little easier to work with categories.

src/onegov/org/views/ticket.py
```diff
@@ -1791,6 +1804,7 @@ def view_archived_tickets(
     owner = next((o for o in owners if o.active), None)
     submitter = next((s for s in submitters if s.active), None)
     layout = layout or ArchivedTicketsLayout(self, request)
+    layout.breadcrumbs[-1] = Link(_('Archived Tickets'), request.link(self))
```
Daverball: Why are we doing this? Shouldn't this be covered by `ArchivedTicketsLayout`? Are we perhaps passing in the wrong layout in the Town6 version of this view or did we maybe define incorrect breadcrumbs for this layout in the first place?

src/onegov/org/views/settings.py
```diff
@@ -194,108 +311,132 @@ def handle_links_settings(
...
+    return handle_generic_settings(self, request, form, _('Analytics'), layout)
+
+
+# Module Settings
+
+
+@OrgApp.form(
+    model=Organisation, name='module-activation-settings', template='form.pt',
+    permission=Secret, form=ModuleActivationSettingsForm,
+    setting=_('Activate/deactivate modules'), order=0,
```
Daverball: I found this to be a little wordy and it also ends up not fitting into one line in the settings view and gets broken up weirdly, on the other hand it's very clear what it means and I can't think of something shorter that still explains what the setting is for. Maybe "Optional Modules" could work?

src/onegov/org/custom.py
```diff
@@ -92,7 +186,7 @@ def get_global_tools(
                     Auth.from_request_path(request), name='register'
                 ), attrs={'class': 'register'})
 
-    # Management dropdown
+    # Managementrequest.link(
```
Daverball:
```suggestion
    # Management dropdown
```

This looks like an accidental edit of the comment

src/onegov/town6/forms/settings.py
```diff
@@ -318,3 +314,11 @@ class RISSettingsForm(Form):
             'Example: Cat 1 text; Cat 2 text; Cat 3 text'),
         render_kw={'rows': 12}
     )
+
+
+class ModuleActivationSettingsForm(OrgModuleActivationSettingsForm):
+    ris_enabled = BooleanField(
+        label=_('Enable Ratsinformationssystem (RIS)'),
```
Daverball, line 321: This is missing a translation

src/onegov/town6/views/ris_settings.py
```diff
@@ -22,7 +22,7 @@
 
 @TownApp.html(
     model=Organisation,
-    name='ris-settings',
+    name='ris-overview',
```
Daverball, line 25: While I do agree that this is a better name, we have to be careful about old links no longer working. Maybe add a redirect from this old name to the new name just in case?

Daverball (approved):
Nicely done. I like this a lot better.

There are a couple of minor issues, like a missing translation.

I also have some other small nitpicks and some suggested potential refactors, but we don't necessarily need to tackle those now, since it is no worse than it was before.

src/onegov/feriennet/templates/layout.pt
BreathingFlesh: Yes, it's a workaround so the global tools are still aligned on the right without any additional css.

src/onegov/town6/views/ris_settings.py
BreathingFlesh: I can't. The new actual RIS-Settings use that name now. But there's only one place where this was linked and it's only for logged in users anyways and only for Wil, they probably won't notice the new path.

Daverball: Fair enough

src/onegov/org/views/settings.py
BreathingFlesh: I like "optional Modules". I will leave the "activate/deactivate Modules" in the modules dropdown and rename it in the settings.

src/onegov/org/layout.py
```diff
@@ -1819,7 +1775,8 @@ class ArchivedTicketsLayout(DefaultLayout):
     def breadcrumbs(self) -> list[Link]:
         return [
             Link(_('Homepage'), self.homepage_url),
-            Link(_('Tickets'), '#')
+            Link(_('Tickets'), '#'),
+            Link(_('Archived Tickets'), self.request.link(self))
```
Daverball:
```suggestion
            Link(_('Tickets'), self.request.class_link(
                TicketCollection,
                {'handler': self.model.handler_code, 'state': 'open'}
            )),
            Link(_('Archived Tickets'), '#')
```

PR #2444, opened 2026-04-14
Api: New filters.

cyrillkuettel:

## Commit message

Api: New filters.

Includes also OGC-3085 which is similar.

TYPE: Feature
LINK: OGC-3086

## Checklist


src/onegov/event/collections/occurrences.py
```diff
@@ -599,6 +621,28 @@ def query(self) -> Query[Occurrence]:
...
+        if self.syndicate is not None:
+            if self.syndicate:
+                query = query.filter(Event.meta['syndicate'].astext == 'true')
+            else:
+                query = query.filter(
+                    or_(
+                        Event.meta['syndicate'] == None,
+                        Event.meta['syndicate'].astext != 'true',
+                    )
+                )
```
Daverball:
```suggestion
            query = query.filter(
                func.coalesce(Event.syndicate, False).is_(self.syndicate)
            )
```
This seems a little better to me, casting to string for a JSONB boolean is a bit ugly and potentially worse performance-wise. The hybrid property handles the type-cast to boolean automagically for you since the default value is a `bool`. We probably could go one step further and also add the `coalesce` step for the default value automagically. Maybe try that as well, you can change `_default_expr` to the following:

```python
    def _default_expr(self, owner: type[Any]) -> ColumnElement[T]:
        column: ColumnElement[dict[str, Any]] = getattr(owner, self.attribute)
        expr = column[self.key]
        if self.value_type is None:
            pass
        elif issubclass(self.value_type, str):
            expr = expr.as_string()
        elif issubclass(self.value_type, bool):
            expr = expr.as_boolean()
        elif issubclass(self.value_type, float):
            expr = expr.as_float()
        elif issubclass(self.value_type, int):
            expr = expr.as_integer()
        default = self.default() if callable(self.default) else self.default
        if default is not None:
            expr = func.coalesce(expr, default)
        return expr
```

If that causes some regressions somewhere our test-suite hopefully will notice.

That way you can change it to the dead-simple
```suggestion
            query = query.filter(Event.syndicate.is_(self.syndicate))
```

Same thing for the highlight filter.

src/onegov/core/orm/mixins/content.py
```diff
@@ -323,16 +323,27 @@ def _default_deleter(self, instance: object) -> None:
...
+            coerced = True
         elif issubclass(self.value_type, int):
             expr = expr.as_integer()
+            coerced = True
+        if coerced:
+            default = (
+                self.default() if callable(self.default) else self.default
+            )
+            if default is not None:
+                expr = func.coalesce(expr, default)
```
Daverball:
```suggestion
        default = (
            self.default() if callable(self.default) else self.default
        )
        if default is not None:
            expr = func.coalesce(
                expr,
                default if coerced else type_coerce(default, JSON)
            )
```
You can still make it work this way when we don't coerce the expression to a simple type.

It might even be worth to always `type_coerce` here so it always literal binds with the correct type for sure, so for each branch you could record the target type, with the default being our `JSON` type.

tests/onegov/org/test_api.py
```diff
@@ -72,36 +75,40 @@ def template(item: Any) -> set[str]:
...
-        'highlight',
+        'empfehlung',
     }
     assert event_fiters['altersgruppe'] == [
         'Kind',
         'Jugend',
         'Familie',
         'Alter'
     ]
-    assert event_fiters['highlight'] == ['Ja']
```
cyrillkuettel, line 104: I renamed this to 'empfehlung' there seems to be in some cases an already existing filter that was also called highlight. Need to watch out for name conflicts

Daverball: Yeah, because that's how they were planning to implement this originally, but they weren't happy that everyone was able to set that flag when submitting an event. Naming conflicts in real use-cases should be unlikely. But we could pretty easily extend `reserved_fields` for the `ValidFilterForm` validator in the event settings (if it doesn't already automatically include these field names from the `EventForm`).

Daverball: Alright, yeah it already does that, just not for the two admin specific fields you added. So you should probably pass it the admin-specific extended form class instead of the general `EventForm`.

Daverball: Or just manually add the extra fields that would add, so you don't have to use `on_request` to set the `reserved_fields` on the validator.

Daverball:
@cyrillkuettel I took the liberty of resolving the merge conflicts, the new test failure is a result of the test not being merged quite correctly, but it should be as simple as adding the missing filter names to the list to make it pass again.

Daverball (approved):

PR #2471, opened 2026-05-08
Feature/ogc 3157 current user none crash

Tschuppi81:
Org: Replace `assert current_user is not None` with a Sentry warning and `HTTPForbidden`
                                                                                                    
Adds `OrgRequest.get_current_user()` which captures a Sentry warning and raises `HTTPForbidden` when `current_user is None` despite a valid session. Replaces all bare asserts across feriennet, org, town6 and translator_directory.                                                   
           
                                                                                                    
TYPE: Feature                                                                                     
LINK: ogc-3157   
LINK: https://seantis-gmbh.sentry.io/issues/7467148569/


Tschuppi81:
@Daverball Is this a good approach to prevent assertion errors at runtime but try to identify the root cause of `current_user` beeing `None`?

src/onegov/translator_directory/generate_docx.py
```diff
@@ -187,9 +187,14 @@ def signature_for_mail_templates(
...
-    assert request.current_user.realname is not None
-    first_name, last_name = request.current_user.realname.split(' ')
+    current_user = request.get_current_user()
+    realname = current_user.realname
+    if not realname:
+        return None
+    name_parts = realname.split(' ', 1)
+    if len(name_parts) != 2:
+        return None
+    first_name, last_name = name_parts
```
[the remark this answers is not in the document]
Tschuppi81: why not directly access `current_user.first_name` resp. `current_user.last_name`?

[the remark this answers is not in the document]
Tschuppi81: none of both is available..

Daverball:
The root cause of this exception is most likely that changing a translator's e-mail address does not log them out. So they're still logged in with their old e-mail, which no longer has an associated account, so `get_current_user` will return `None`.

So I don't think we should make any large sweeping changes to make the application not crash, when this happens. Since after all it is a real problem if we didn't log that user out.

It might be worth adding a `change_username` method to `User` which calls `self.logout_all_sessions()` and then switch to using that instead of manually updating `User.username`, everywhere it is relevant, so we can't forget to do this step.

Although we probably could change `current_user` to use the primary key instead of the `username` in order to retrieve the `User` object, since we now store both in the identity object and the primary key will never change. But we would still need to detect when the username in the identity no longer matches the username of the `User` and update the cached username in the session, since otherwise there may still be some odd behavior for code that relies on `request.current_username`.

Tschuppi81:
> The root cause of this exception is most likely that changing a translator's e-mail address does not log them out

Is that related to LDAP?

Daverball:
> Is that related to LDAP?

It shouldn't be an issue during external login, although I'm not sure what happens if they have multiple sessions and their LDAP username gets changed during external login, I'm not sure if we remembered to terminate all the other sessions, so only the one with the new name remains.

But with translators it's also possible for admins to change the name of the translator, which gets mirrored back to the user if I'm not mistaken and we could be forgetting to log the user out there as well. There's also the LDAP synchronization cronjob, although I'm not sure if we have that for translator directory or if it's only running for FSI.

Either way there are quite a few ways the username can change, and we probably didn't remember to log out the active session for some of the ways. That's why adding a `change_username` method makes sense to me, so it's not as easy to forget. Either way you will have to look at all of the places in the code that update the `username` and check if they also call `logout_all_sessions()` and either add the missing call or refactor to `change_username()`.

src/onegov/core/security/identity_policy.py
```diff
@@ -37,6 +40,8 @@ def remember(
     ) -> None:
         for key in self.required_keys:
             request.browser_session[key] = getattr(identity, key)
+        if (uid := getattr(identity, 'uid', None)) is not None:
+            request.browser_session['uid'] = uid
```
Daverball: Just add `uid` to `required_keys` instead. I think I just forgot to do that, when I added `uid` to `Identity`.

src/onegov/org/request.py
```diff
@@ -114,11 +119,27 @@ def current_user(self) -> User | None:
...
+                    # stale username in session — fix it for subsequent
+                    # requests
+                    self.browser_session['userid'] = (
+                        self._current_user.username)
+            else:
+                self._current_user = (
+                    self.session.query(User)
+                    .filter_by(username=self.identity.userid)
+                    .first()
+                )
```
Daverball: This branching should not be necessary, valid identities always should have a `uid`. The only reason it sometimes didn't before, is because we forgot to put it in the cache along with the other identity attributes.

src/onegov/org/request.py
```diff
@@ -103,6 +105,9 @@ def is_member(self) -> bool:
 
     @property
     def current_username(self) -> str | None:
+        user = self.current_user
+        if user is not None:
+            return user.username
```
Daverball: I'm not sure I'm entirely comfortable with this change. It is a little bit more robust, but it means we always have to load the entire user object on every request where we access this property, even though we have the username cached and it should always be correct, if we didn't leave an old session hanging around, that should've been terminated.

I think we should try to get away with the old implementation, but we could check if `_current_user` already exist, when it does we can return `_current_user.username` instead. But I don't think we should emit an extra database query, unless we definitely need it.

src/onegov/user/auth/core.py
```diff
@@ -286,6 +282,11 @@ def as_identity(self, user: User) -> Identity:
 
     def by_identity(self, identity: Identity | NoIdentity) -> User | None:
         """ Returns the user record of the given identity. """
+        uid = getattr(identity, 'uid', None)
+        if uid:
+            user = self.users.by_id(UUID(uid))
+            if user is not None:
+                return user
```
Daverball: Once again, this branch should be unnecessary and we should skip the query if `identity.uid is None`, I think you also don't need to convert to `UUID`, SQLAlchemy should automatically coerce the value to the correct type.

tests/onegov/user/test_models.py
```diff
@@ -176,6 +176,23 @@ def test_user_logout_all_sessions() -> None:
...
+            remembered.return_value = True
+
+            user.save_current_session(DummyRequest('sess1'))  # type: ignore[arg-type]
+            assert user.sessions is not None
+            assert 'sess1' in user.sessions
+
+            user.change_username('new@example.org', DummyRequest('zzz').app)  # type: ignore[arg-type]
+
+            assert call(None, 'sess1') in forget.mock_calls
+            assert user.username == 'new@example.org'
```
Daverball:
```suggestion
def test_user_change_username() -> None:
    user = User(username='old@example.org')

    with (
        patch('onegov.user.models.user.remembered') as remembered,
        patch('onegov.user.models.user.forget') as forget
    ):
        remembered.return_value = True

        user.save_current_session(DummyRequest('sess1'))  # type: ignore[arg-type]
        assert user.sessions is not None
        assert 'sess1' in user.sessions

        user.change_username('new@example.org', DummyRequest('zzz').app)  # type: ignore[arg-type]

        assert call(None, 'sess1') in forget.mock_calls
        assert user.username == 'new@example.org'
```

You can invoke multiple context managers in a single `with` statement, so you don't need to indent twice.

Daverball (commented):
Looks good overall, but we can simplify some of the `uid` vs `userid` logic.

Daverball (approved):

PR #2472, opened 2026-05-08
Core: Replace per-table has_table N+1 with single get_table_names on …

Tschuppi81:
Core: Replace N+1 has_table queries with single get_table_names on schema init
                                                                              
TYPE: Performance
LINK: https://seantis-gmbh.sentry.io/issues/7472765445/
LINK: https://seantis-gmbh.sentry.io/issues/7357002739/

The fix fetches all existing table names in the schema with a single Inspector.get_table_names() call, then passes only the missing tables to create_all(..., checkfirst=False), reducing from O(n) to O(1)


Daverball (commented):
This seems like the kind of optimization SQLAlchemy should be doing itself inside `Metadata.create_all()`. It might be worth opening an issue on their bug tracker.

I don't know if I'm comfortable with doing this optimization in our code, especially since it only matters the first time a schema is accessed, so we don't really gain any substantial benefits during normal operation of the application. It might slightly speed up some CLI commands, but probably not to a noticeable degree, since the biggest overhead there is podman spinning up an extra instance of the onegov container and the initial loading of all the python modules.

Tschuppi81:
https://github.com/sqlalchemy/sqlalchemy/issues/13294

Daverball:
FYI, you can link sentry issues via `LINK` just like linear issues, our release script recognizes the `ONEGOV-CLOUD-` prefix as a sentry issue and will generate an appropriate url.

I think we'll try this workaround for now and hope it doesn't cause any issues, but make sure to include a `FIXME` comment with a link to the SQLAlchemy discussion, so we can simplify our code, if and when they make an optimization on their end.

Tschuppi81:
Good point, like this we are not blocked

PR #2483, opened 2026-05-18
Bugfix/view people with no sub org

Tschuppi81:
People: Fix missing None check for sub organizations

TYPE: Bugfix
LINK: None
SENTRY: https://seantis-gmbh.sentry.io/issues/3989540578/

### DIFF
--- a/src/onegov/org/views/people.py
+++ b/src/onegov/org/views/people.py
@@ -48,12 +48,14 @@ def get_top_level_organisations(
 
 
 def get_sub_organisations(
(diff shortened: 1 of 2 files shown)


Tschuppi81:
Not needed, as i fixed the data in production `riddes/people-settings`
Screenshot from 2026-05-18 15-11-42

Tschuppi81:
I don't know how the db could get the data like this, via form it is not possible!

Sanitizing would anyway be better than changing the code.

PR #2490, opened 2026-05-22
Propagate psql error while updating search index

Tschuppi81:
Search: Propagate psql error while updating search index

Under high load (e.g. Feriennet opening days) concurrent bookings can trigger PostgreSQL serialization failures inside the search indexer's savepoint. Previously these were silently swallowed, leaving the outer transaction dead (booking lost) while the confirmation email was still dispatched.

TYPE: Bugfix
LINK: pro-1545

### DIFF
--- a/src/onegov/core/framework.py
+++ b/src/onegov/core/framework.py
@@ -67,7 +67,8 @@ from psycopg2.extensions import TransactionRollbackError
 from purl import URL
 from sqlalchemy.exc import OperationalError
 from urllib.parse import urlencode
-from webob.exc import HTTPConflict, HTTPServiceUnavailable
+from translationstring import TranslationString
+from webob.exc import HTTPConflict, HTTPSeeOther, HTTPServiceUnavailable
 
 
 from typing import overload, Any, Literal, TYPE_CHECKING
@@ -1771,6 +1772,24 @@ def http_conflict_tween_factory(
                 raise
 
             log.warning('A transaction failed because there was a conflict')
+
+            # Redis-backed browser_session is not rolled back with the postgres
+            # transaction, so stale flash messages must be cleared manually.
+            request.browser_session.pop('messages', None)
+
+            # POST form: redirect back so user sees the alert and can retry.
+            if (request.method == 'POST'
+                    and not request.is_xhr
+                    and request.content_type in (
+                        'application/x-www-form-urlencoded',
+                        'multipart/form-data',
+                    )):
+                request.alert(request.translate(TranslationString(
+                    'Your request could not be completed because of a '
+                    'simultaneous conflicting request. Please try again.',
+                    domain='onegov.org'
+                )))
+                return HTTPSeeOther(location=request.path)
 
             return HTTPConflict()
 
--- a/src/onegov/search/indexer.py
+++ b/src/onegov/search/indexer.py
@@ -12,6 +12,7 @@ from onegov.search.utils import language_from_locale
 from operator import itemgetter
 from psycopg2.extensions import TransactionRollbackError
 from sqlalchemy import and_, bindparam, delete, func, text, String
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import object_session
 from sqlalchemy.orm.exc import ObjectDeletedError
 from sqlalchemy.dialects.postgresql import insert, ARRAY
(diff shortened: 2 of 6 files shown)


Tschuppi81:
The error can be reproduced locally when editing `index()` in indexer.py.
```
            with session.begin_nested():
                print('*** tschupre force serialization failure')
                from sqlalchemy.exc import OperationalError as _OE
                raise _OE(
                    statement=None,
                    params=None,
                    orig=psycopg2.errors.SerializationFailure('simulated'),
                )
                session.execute(stmt, list(params_dict.values()))
```
Then do a feriennet booking on an occasion with creation of a child. This forces the the transaction to fail ending up as a 409 error

Tschuppi81:
The current fix will show the user a 409, which is not nice. I will try to get the user back to the `<occasion>/book` page in order to re-submit

Screenshot from 2026-05-22 03-35-53

Tschuppi81:
With my third [commit](https://github.com/OneGov/onegov-cloud/pull/2490/changes/fb36f5657fe2eb2092c59957417ee5fb8ff5bc7d) the user gets back to its POST url with flash error message. From there a resubmit is clearer and more user friendly than a 409.

Screenshot from 2026-05-22 10-47-42

Tschuppi81:
FYI: @BreathingFlesh @strfx

src/onegov/core/framework.py
```diff
@@ -1772,6 +1773,24 @@ def http_conflict_tween(request: CoreRequest) -> Response:
...
+                        'application/x-www-form-urlencoded',
+                        'multipart/form-data',
+                    )):
+                request.alert(request.translate(TranslationString(
+                    'Your request could not be completed because of a '
+                    'simultaneous conflicting request. Please try again.',
+                    domain='onegov.org'
+                )))
+                return HTTPSeeOther(location=request.path)
+
```
Daverball, line 1793: This seems pretty fragile to me. Returning a 409 seems perfectly fine to me, they can just refresh to resubmit, it's not guaranteed that the form is submitted from the same url it's POST to, so this could definitely cause issues. You could try looking at the referrer, but that's not super reliable either. I think it's better to keep this the way it was.

As for flash messages not getting rolled back: I think that's a separate issue entirely and this is not a super robust solution, we should add a `DataManager`, just like we do for emails and sms, so the pending messages get rolled back along with the rest of the transaction. I'm also using a `DataManager` for the search index.

src/onegov/search/indexer.py
Daverball, line 1: This solution doesn't really make all that much sense to me, submitting changes to the search index is part of the commit step of the transaction, we deliberately try to ignore any failures in this step, so if we're not properly isolating and rolling back failures in the indexer, without causing some issues for the rest of the transaction, then that's the real problem and we should be looking into why `begin_nested` manages to roll back changes that happened before the nested transaction, when the commit should still happen. Is it possible the session didn't commit everything before the indexer tried starting its nested transaction, so those changes ended up getting bundled into the nested transaction instead of the outer one? I thought I remembered explicitly flushing the session though before entering the indexing phase.

Either way this seems a little fragile to me, since other database errors during indexing could still cause the same problem. It's not limited to serialization failures, even if those should be the most common ones.

I guess one thing I'm not fully sure about is what happens if `session.flush` causes a serialization failure within the data manager for the search index. Perhaps that gets swallowed, without dooming the transaction and that's why it manages to move on to the next data manager. But I'm not sure how `zope.sqlalchemy` data manager would succeed after that, since the SQL transaction should remain in a failed state, even if the subtransaction was rolled back, since the failure happened before the subtransaction was started, which the data manager should detect and then doom the transaction.

Maybe it's a subtle bug in SQLAlchemy where calling `begin_nested` in a failed transaction doesn't start a subtransaction and then the context manager ends up rolling back the whole transaction, instead of only the nested one, since it never created a nested one. In which case we would probably need to check if the database transaction is still alive at the start of indexing, and skip the indexing entirely if it's already dead.

Daverball (changes requested):
Overall this is a concerning problem and I'm happy you were able to identify it. However I'm not as happy with the proposed solution, it seems like a hack at best.

I think we should try to nail down why the data manager for the indexer manages to keep the transaction alive, without the database transaction as a whole succeeding after the rollback of the nested transaction and fix that, rather than try to work around it by not silencing the kind of errors we can recover from. Since we will continue to have the same very bad issues with other database errors during indexing, we just likely will rarely see it, which could give us a false sense of security.

src/onegov/search/indexer.py
Daverball: Another thing we could try is installing a tween under the transaction tween that calls `request.session.flush()`, so it gets called at the end of every request, before we start committing the transaction, so any database failures surface early and doom the transaction before it enters the data manager negotiation phase.

Tschuppi81: Agree, this is more a workaround for specific errors.

I try to find a way to get the connection state from within I guess `tpc_vote` right after processing..

Tschuppi81:
not needed any longer

PR #2491, opened 2026-05-26
propagate index errors to abort the transaction

Tschuppi81:
Search: Propagate errors while updating or deleting search index

Under high load (e.g. Feriennet opening days) concurrent bookings can trigger PostgreSQL serialization failures inside the search indexer's savepoint. Previously these were silently swallowed, leaving the outer transaction dead (booking lost) while the confirmation email was still dispatched.

TYPE: Bugfix
LINK: pro-1545

Diff for hotfix: Remove two `begin_nested` and add two `raise` statements.

```
reto@reto:~/workspace/onegov-cloud$ git diff
diff --git a/src/onegov/search/indexer.py b/src/onegov/search/indexer.py
index a5730057a..9c5c2e138 100644
--- a/src/onegov/search/indexer.py
+++ b/src/onegov/search/indexer.py
@@ -291,14 +291,13 @@ class Indexer:
                 # will not be able to infer the matching constraint
                 index_where=owner_id_column.is_not(None)
             )
-            with session.begin_nested():
-                session.execute(stmt, list(params_dict.values()))
+            session.execute(stmt, list(params_dict.values()))
         except Exception:
             index_log.exception(
                 f'Error creating index schema {schema} of '
                 f'table {tablename}, tasks: {[t["id"] for t in tasks]}',
             )
-            return False
+            raise
 
         return True
 
@@ -368,13 +367,12 @@ class Indexer:
                     owner_id_column.in_(owner_ids)
                 ))
             )
-            with session.begin_nested():
-                session.execute(stmt)
+            session.execute(stmt)
         except Exception:
             index_log.exception(
                 f'Error deleting index schema {schema} tasks {tasks}:'
             )
-            return False
+            raise
 
         return True
```


Tschuppi81:
this pr would replace this one https://github.com/OneGov/onegov-cloud/pull/2490

Tschuppi81:
@Daverball just propagating the exception leads to a transaction abort/rollback as expected. No booking, no confirmation email

Daverball:
> @Daverball just propagating the exception leads to a transaction abort/rollback as expected. No booking, no confirmation email

That makes sense, yeah, we can do this as a quick fix, long term it would still be nice to be able to emit a rollback for just the indexing step, if the rest works fine.

Instead of logging and reraising you can also just get rid of the `try/except` block entirely. We don't need to log indexing errors if we're reraising, especially if the transaction is retried, since we would see an indexing error in the log, even if it succeeded in subsequent tries.

Tschuppi81:
The 409 appears when an error happens during indexing

image

src/onegov/search/indexer.py
```diff
@@ -105,200 +106,193 @@ def index(
...
+                'title_vector': stmt.excluded.title_vector,
+                'data_vector': stmt.excluded.data_vector,
+            },
+            # since our unique constraints are partial indeces
+            # we need this index_where clause, otherwise postgres
+            # will not be able to infer the matching constraint
+            index_where=owner_id_column.is_not(None)
+        )
+        with session.begin_nested():
+            session.execute(stmt, list(params_dict.values()))
```
Daverball:
```suggestion
       session.execute(stmt, list(params_dict.values()))
```

We also no longer should emit a nested transaction, since that may have been the source of our bug in the first place, if the transaction is doomed and we call `begin_nested`, it may have the same effect as `begin` and resetting the transaction to a healthy state.

src/onegov/search/indexer.py
```diff
@@ -357,24 +351,18 @@ def delete(
...
+        assert owner_id_column is not None
+        stmt = (
+            delete(SearchIndex.__table__)  # type: ignore[arg-type]
+            .where(and_(
+                SearchIndex.owner_tablename == tablename,
+                owner_id_column.in_(owner_ids)
+            ))
+        )
+        with session.begin_nested():
+            session.execute(stmt)
```
Daverball:
```suggestion
        session.execute(stmt)
```

Daverball:
@Tschuppi81 Can you merge master? It should fix the linting errors. Other than that, this looks fine.

We should consider hotfixing this after today's release, but let's do the `raise` thing you did initially in the hotfix, so it's a smaller change, so we only have to add two `raise` statements and remove the two `begin_nested` wrappers.

PR #2495, opened 2026-05-28
Org: Fixes event search index corrupted after fetch.

cyrillkuettel:
We have hourly cronjobs that fetch events. They look like this: `sudo /usr/sbin/container-exec onegov-cloud onegov-org --select /onegov_town6_without_yubikey/zug fetch --source huenenberg --source risch --source steinhausen --published-only --create-tickets`

We have empirically determined that this exact  command above results in:
image

**I have tested this fix (hotfix and run the cli manually) and with the fix this problem doesn't appear.**

## Commit message

Org: Fixes event search index corrupted after fetch.

The fetch CLI switches schemas multiple times in a loop (remote → local
per source key), violating the session_manager's own design contract:
"we only set the schema once per request". Any ORM flush during the
schema-switch window would send indexing signals tagged with the wrong
schema.

Wrap from_import() and the submit/publish/ticket block in
disable_change_signals(), then explicitly index each added/updated
published event after signals are restored.

TYPE: Bugfix
LINK: OGC-3101

## Checklist

### DIFF
--- a/src/onegov/org/cli.py
+++ b/src/onegov/org/cli.py
@@ -762,33 +762,42 @@ def fetch(
                     return TicketCollection(local_session).by_handler_id(
                         event_id.hex)
 
-                added, updated, purged = local_events.from_import(
-                    remote_events(),
-                    to_purge=[f'fetch-{key}'],
-                    publish_immediately=False,
-                    valid_state_transfers=valid_state_transfers,
-                    published_only=published_only
-                )
+                with app.session_manager.disable_change_signals():
+                    added, updated, purged = local_events.from_import(
+                        remote_events(),
+                        to_purge=[f'fetch-{key}'],
+                        publish_immediately=False,
+                        valid_state_transfers=valid_state_transfers,
+                        published_only=published_only,
+                    )
 
-                for event_ in added:
-                    event_.submit()
-                    if not create_tickets:
-                        event_.publish()
-                        continue
+                    for event_ in added:
+                        event_.submit()
+                        if not create_tickets:
+                            event_.publish()
+                            continue
 
-                    assert local_admin is not None
-                    with local_session.no_autoflush:
-                        tickets = TicketCollection(local_session)
-                        new_ticket = tickets.open_ticket(
-                            handler_code='EVN', handler_id=event_.id.hex,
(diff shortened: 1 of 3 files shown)


src/onegov/org/cli.py
```diff
@@ -762,33 +762,42 @@ def ticket_for_event(
...
+                                request,
+                                (f'Importiert von Instanz {key}'),
+                                owner=local_admin.username,
+                            )
 
-                        ), owner=local_admin.username)
+                if hasattr(app, 'fts_orm_events'):
+                    for event_ in (*added, *updated):
+                        if not event_.fts_skip:
+                            app.fts_orm_events.index(local_schema, event_)
```
Daverball:
```suggestion
                    if app.fts_search_enabled:
                        if added or updated:
                            self.fts_indexer.process((
                                task
                                for event_ in chained(added, updated)
                                if (
                                    task := self.fts_orm_events.index_task(local_schema, _event)  # type: ignore[arg-type]
                                ) is not None
                            ), local_session)
                        if purged:
                            self.fts_indexer.process((
                                task
                                for event_ in purged
                                if (
                                    task := self.fts_orm_events.delete_task(local_schema, _event)  # type: ignore[arg-type]
                                ) is not None
                            ), local_session)
                        # NOTE: Make sure all changes are flushed, before signals are emitted again
                        local_session.flush()
```
I think it's worth doing what we're doing in `perform_reindex` here and directly process the tasks using the indexer, instead of going through the queue, if you're already taking the extra step to manually emit the tasks.

You also forgot to emit delete tasks for the purged events.

Daverball (changes requested):
I think we can go a step further and directly emit the tasks to the indexer, instead of going the indirect route through the queue. You also forgot to handle the pruned events.

src/onegov/org/cli.py
cyrillkuettel: Ok this is updated now. The CI seems to be quite red, a lot of stuff failing on master as well. Seems unrelated tough.

Daverball: Redis 8 was just released, looks like that's the cause, pin it to <8 for now.

PR #2502, opened 2026-06-01
Adds keywords (e.g. synonyms) to Topics/Pages. Keywords will be found…

Tschuppi81:
Org: Adds keywords (e.g. synonyms) to topics/pages. Keywords are also in the search index.

TYPE: Feature
LINK: ogc-3155

### DIFF
--- a/src/onegov/org/forms/page.py
+++ b/src/onegov/org/forms/page.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from onegov.form import Form
 from onegov.form.fields import ChosenSelectField
+from onegov.form.fields import TagsField
 from onegov.form.fields import URLField
 from onegov.org import _
 from onegov.org.forms.fields import HtmlField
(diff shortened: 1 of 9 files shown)


src/onegov/org/forms/page.py
```diff
@@ -59,10 +60,21 @@ class PageForm(PageBaseForm):
...
+            'Helps visitors find this page even if they use different words.'
+        )
+    )
+
     lead_when_child = BooleanField(
         label=_('Show the lead if accessing the parent page'),
     )
 
+    def on_request(self) -> None:
+        self.request.include('tags-input')
```
cyrillkuettel, line 76: nitpick: This probably should also have `super().on_request()` right?

cyrillkuettel (approved):

src/onegov/org/forms/page.py
Tschuppi81: Usually we add the `super().on_request()` if we derive our form from a class other than `Form` which is true for this case. But the base class `PageBaseForm` does not have a `on_request` method and therefore its not needed

PR #2506, opened 2026-06-01
Onboarding: Fixes town assistant.

cyrillkuettel:
Note: The url to test locally is http://127.0.0.1:8080/onegov_onboarding

## Commit message

Onboarding: Fixes town assistant.

Submitting the town onboarding assistant failed at request commit. Three
seperate issues had to be resolved.

1. Morepath apps weren't committed at server startup, so
app.translations was empty → `assert translator is not None` in
`create_new_organisation`. (Fix: Commit all registered morepath applications)
Note: I am unsure at this point how this ever worked in the first place.

2. `perform_reindex()` bypasses the transaction machinery (raw `COMMIT`)
 for speed. That is fine for the CLI, but when called mid-request
  it tears down the request-managed transaction, so the request fails with
`ResourceClosedError: This transaction is closed.`

3. Added `perform_reindex_in_transaction()` which is only necessary because
of an otherwise `AssertionError` (`src/onegov/core/orm/cache.py:170`, inside `handle_orm_change receiver`)

TYPE: Bugfix
LINK: OGC-2727

## Checklist

### DIFF
--- a/src/onegov/onboarding/models/town_assistant.py
+++ b/src/onegov/onboarding/models/town_assistant.py
@@ -177,44 +177,46 @@ class TownAssistant(Assistant):
             schema = self.get_schema(name)
             custom_config = self.config['configuration']
 
-            self.app.session_manager.set_current_schema(schema)
-            session = self.app.session_manager.session()
+            with self.app.session_manager.disable_change_signals():
+                self.app.session_manager.set_current_schema(schema)
+                session = self.app.session_manager.session()
 
-            if session.query(Organisation).first():
-                raise AlreadyExistsError
+                if session.query(Organisation).first():
+                    raise AlreadyExistsError
 
-            with self.app.temporary_depot(schema, **custom_config):
-                create_new_organisation(self.app, name=name, reply_to=user)
+                with self.app.temporary_depot(schema, **custom_config):
+                    create_new_organisation(self.app, name=name, reply_to=user)
 
-            org = session.query(Organisation).first()
-            assert org is not None and org.theme_options is not None
-            org.theme_options['primary-color-ui'] = color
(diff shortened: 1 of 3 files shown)


Daverball:
I'm a little confused about 2/3), since all the reindexing happens on separate threads, they should be using their own sessions with their own transaction, so issuing raw `COMMIT` and invalidating sessions should have no effect on the session on the main thread.

It is however still true that this would make it possible for something to fail after indexing, which causes most of the changes to be rolled back, but not the search index changes, since they have already been committed. But that doesn't seem so bad to me, considering there will already be some things that will not be rolled back, since they always happen outside transactions, such as the creation of the schema and the creation of the empty tables etc.

So I think the only problem is, that at the very end `perform_reindex` also disposes of the session on the main thread. I think I would prefer simplifying this by adding an extra argument `dispose_session: bool = True` to the method and putting the final three lines of the method into an `if dispose_session:` block.

Then you can call the method with `dispose_session=False` here.

Daverball (approved):

cyrillkuettel:
Oh, I wasn't aware of the seperate Threads with their own transaction. Now it makes sense. This seems good now after fixing linting.

Daverball:
Yeah, I'm honestly also not sure why we needed to start disposing of the session on the main thread, the old elastic search indexer didn't need to do that. But without that the postgres connections were not released back to the connection pool, so on reindexes with a large enough number of schemas we were running out of connections, so the reindex did not go through for all the schemas. We might have a bug in our CLI runner logic, that doesn't properly clean up sessions after we finish processing a schema in some scenarios, but explicitly killing the session was the most straightforward workaround. It's also possible that the workaround is no longer necessary.

PR #2511, opened 2026-06-03
Foundation: Replaces `libsass` with dart sass binary

cyrillkuettel:

## Commit message

Foundation: Replaces libsass with dart sass binary

Dart Sass is the primary implementation of Sass,

TYPE: Feature
LINK: OGC-3071

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -182,9 +182,8 @@ jobs:
           DART_SASS_VERSION=1.100.0
           curl -sL "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz" \
             | tar xz -C /tmp
-          mv /tmp/dart-sass/sass /usr/local/bin/sass
-          mv /tmp/dart-sass/src /usr/local/bin/src
-          rm -rf /tmp/dart-sass
+          mv /tmp/dart-sass /app/dart-sass
+          ln -s /app/dart-sass/sass /app/bin/sass
         fi
 
     - name: Prepare test environment
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -176,6 +176,8 @@ jobs:
         # NOTE: Suppress uv warning
         UV_LINK_MODE: copy
 
+    # TODO: Remove this step once the Dockerfile dart-sass change is merged
+    #       to master.
     - name: Install Dart Sass
       run: |
         if ! command -v sass &>/dev/null; then
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -176,6 +176,17 @@ jobs:
         # NOTE: Suppress uv warning
         UV_LINK_MODE: copy
 
+    - name: Install Dart Sass
+      run: |
+        if ! command -v sass &>/dev/null; then
+          DART_SASS_VERSION=1.100.0
+          curl -sL "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz" \
+            | tar xz -C /tmp
+          mv /tmp/dart-sass/sass /usr/local/bin/sass
+          mv /tmp/dart-sass/src /usr/local/bin/src
+          rm -rf /tmp/dart-sass
+        fi
+
     - name: Prepare test environment
       run: |
         # reset the build asset permissions
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -186,6 +186,7 @@ jobs:
             | tar xz -C /tmp
           mv /tmp/dart-sass /app/dart-sass
           ln -s /app/dart-sass/sass /app/bin/sass
+          ln -s /app/bin/sass /usr/local/bin/sass
         fi
 
     - name: Prepare test environment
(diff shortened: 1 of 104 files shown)


.github/workflows/tests.yml
```diff
@@ -176,6 +176,17 @@ jobs:
...
         UV_LINK_MODE: copy
 
+    - name: Install Dart Sass
+      run: |
+        if ! command -v sass &>/dev/null; then
+          DART_SASS_VERSION=1.100.0
+          curl -sL "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz" \
+            | tar xz -C /tmp
+          mv /tmp/dart-sass/sass /usr/local/bin/sass
+          mv /tmp/dart-sass/src /usr/local/bin/src
```
Daverball: Does dart-sass actually depend on `src`? It seems like an awfully generic name to put into `/usr/local/bin`. If it doesn't, just only install the binary please.

src/onegov/foundation6/theme.py
```diff
@@ -285,11 +285,23 @@ def compile(self, options: Mapping[str, Any] | None = None) -> str:
...
+                f'--style={style}',
+                '--quiet-deps',
+                *(f'--load-path={p}' for p in paths),
+            ],
+            input=theme.getvalue(),
+            capture_output=True,
+            text=True,
+            check=True,
         )
+        return result.stdout
```
Daverball: I think refactoring this into `compile_sass(input, paths)` method would be a little bit cleaner, then you don't need to duplicate as much code in this subclass.

Alternatively you could refactor the part that renders the input and the part that generates the path into their own sub-methods, so you can override those instead of `render`.

src/onegov/gever/gever_client.py
Daverball, line 1: I think you can revert this file, you were probably stuck on an old version of niquests when you first created this PR. You can also remove the `FIXME` comments.

Dockerfile
```diff
@@ -82,6 +82,14 @@ RUN GOBIN=/tmp go install github.com/seantis/hivemind@v1.0.4 \
     && cp /tmp/hivemind /usr/local/bin/hivemind \
     && rm -rf /tmp/hivemind
 
+# install dart-sass (SCSS compiler)
+ARG DART_SASS_VERSION=1.100.0
+RUN curl -sL "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz" \
+    | tar xz -C /tmp \
+    && mv /tmp/dart-sass/sass /usr/local/bin/sass \
+    && mv /tmp/dart-sass/src /usr/local/bin/src \
```
Daverball: Same thing with `src` here.

README.md
```diff
@@ -127,20 +127,27 @@ To run OneGov Cloud locally, you must meet the following requirements:
...
 
 To install the requirements for Ubuntu do:
 
     sudo apt install postgresql postgresql-contrib redis-server nodejs python3 docker-compose
 
+Install [Dart Sass](https://github.com/sass/dart-sass/releases):
+
+    curl -sL https://github.com/sass/dart-sass/releases/download/1.100.0/dart-sass-1.100.0-linux-x64.tar.gz | tar xz -C /tmp
+    sudo cp /tmp/dart-sass/sass /tmp/dart-sass/src /usr/local/bin/
+    rm -rf /tmp/dart-sass
```
Daverball: I think it would be nice if we just add this setup step to the `Makefile`, like `ensure_uv`, and install it to `${VIRTUAL_ENV}/bin`. You would need to check the platform though, for MacOS we probably would want to tell people to run `brew install sass/sass/sass` instead of doing anything.

And same deal with `src` here.

Daverball (commented):
Looks good overall, I'm not sure about the `src` portion of the install though and I think we can simplify the install for local development and install it directly inside the virtual environment.

.github/workflows/tests.yml
cyrillkuettel: It does indeed seem so:

  ▎ The archive is designed to be extracted as-is and added to path. I strongly recommend following that
  ▎ pattern. If you just pick a few files out of it and put them in /usr/local/bin, you'll find that it's
  ▎ missing important components at the paths it expects.

https://github.com/sass/dart-sass/issues/617#issuecomment-471502292

I agree and am not a fan of dumping stuff in  `/usr/local/bin`. I'll try if I can get the `Makefile` approach work with `${VIRTUAL_ENV}/bin`

Dockerfile
```diff
@@ -82,6 +82,14 @@ RUN GOBIN=/tmp go install github.com/seantis/hivemind@v1.0.4 \
     && cp /tmp/hivemind /usr/local/bin/hivemind \
     && rm -rf /tmp/hivemind
 
+# install dart-sass (SCSS compiler)
+ARG DART_SASS_VERSION=1.100.0
+RUN curl -sL "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz" \
+    | tar xz -C /tmp \
+    && mv /tmp/dart-sass /app/dart-sass \
+    && mkdir -p /app/bin \
+    && ln -s /app/dart-sass/sass /app/bin/sass
```
Daverball: This should probably happen after building onegov-cloud now, so `/app/bin` already exists, I'm not sure what `venv` does if a `bin` folder already exists.

.github/workflows/tests.yml
```diff
@@ -176,6 +176,16 @@ jobs:
         # NOTE: Suppress uv warning
         UV_LINK_MODE: copy
 
+    - name: Install Dart Sass
```
Daverball, line 181: We can probably get rid of this step, once this is merged on master, since the test container should already contain it afterwards. Maybe add a `TODO` comment to get rid of this step.

Daverball (approved):

PR #2512, opened 2026-06-08
Directory: Derive populate_obj exclude set from directory structure f…

Tschuppi81:
Directory: Derive populate_obj exclude set from directory structure fields

Previously the exclude fields were derived by field type and some needed to be added explicitly

TYPE: Bugfix
LINK: ogc-3195

### DIFF
--- a/src/onegov/directory/models/directory.py
+++ b/src/onegov/directory/models/directory.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import inspect
 
 from email_validator import validate_email
 from enum import Enum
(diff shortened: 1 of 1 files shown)


src/onegov/directory/models/directory.py
```diff
@@ -563,17 +561,9 @@ def populate_obj(
...
-
-                include = (
-                    'publication_start',
-                    'publication_end',
-                    'coordinates',
-                )
-                exclude = {k for k in exclude if k not in include}
+                # skip directory structure fields, directory.update()
+                # handles those
+                exclude = {f.id for f in directory.fields}
```
Daverball, line 566: There is a small potential optimization, that would apply to the `directory_update = False` case.

You could store the output of `parse_form` above into a variable before subclassing it, so you can access all the field ids that way, instead of relying on `directory.fields` which will parse the form a second time. But since `mixed_data` also accesses `directory.fields` through `directory.file_fields` and that part gets cached, it's only a win, when we don't update the custom fields on the directory entry.

Overall it's mostly a non-issue, you can decide whether you want to optimize that code path a bit or keep the change as simple as possible.

Daverball (approved):
There is a small potential optimization, but it looks good to me either way.

PR #2513, opened 2026-06-08
Form: Update image size information in submission after potential resizing

Tschuppi81:
Form: Update image size information in submission after potential resizing

TYPE: Feature
LINK: ogc-3199
HINT: `onegov-org --select /onegov_org/* fix-submission-file-sizes`
HINT: `onegov-org --select /onegov_town6/* fix-submission-file-sizes`

### DIFF
--- a/src/onegov/agency/forms/agency.py
+++ b/src/onegov/agency/forms/agency.py
@@ -16,7 +16,7 @@ from onegov.form import Form
 from onegov.form.fields import ChosenSelectField, HtmlField
 from onegov.form.fields import MultiCheckboxField
 from onegov.form.fields import UploadField
-from onegov.form.validators import FileSizeLimit, MIME_TYPES_IMAGE
+from onegov.form.validators import ImageFileSizeLimit, MIME_TYPES_IMAGE
 from onegov.gis import CoordinatesField
 from sqlalchemy import func
 from wtforms.fields import StringField
@@ -72,7 +72,7 @@ class ExtendedAgencyForm(Form):
     organigram = UploadField(
         label=_('Organigram'),
         validators=[
-            FileSizeLimit(1 * 1024 * 1024)
+            ImageFileSizeLimit()
         ],
         allowed_mimetypes=MIME_TYPES_IMAGE,
     )
--- a/src/onegov/agency/forms/agency.py
+++ b/src/onegov/agency/forms/agency.py
@@ -12,11 +12,15 @@ from onegov.agency.models import ExtendedAgency
 from onegov.agency.utils import handle_empty_p_tags
 from onegov.core.security import Private
 from onegov.core.utils import linkify, ensure_scheme
+from onegov.file.attachments import IMAGE_MAX_SIZE
 from onegov.form import Form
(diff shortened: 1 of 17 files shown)


Tschuppi81:
There is only a file size defined which in my opinion does not make sense for images like jpg. Therefore I limit the pixel size as it was, just increased it actually.

Daverball:
I'm not sure if the new size limit is much better than the old one, it's still pretty low if you want to print A4 or larger. Which is why a file size limit might be better than a pixel based limit. Allowing arbitrarily large images (in pixel dimensions) does come with its own problems though.

It might be worth making pixel dimension limits and automatic resizing part of the field validation instead, so we can still limit image dimensions in image sets, but leave the original image intact (apart from stripping the EXIF data) everywhere else, since it's more likely that the image won't be displayed inline anywhere and is more of a attachment that will be downloaded.

src/onegov/agency/forms/agency.py
```diff
@@ -72,7 +72,7 @@ class ExtendedAgencyForm(Form):
     organigram = UploadField(
         label=_('Organigram'),
         validators=[
-            FileSizeLimit(1 * 1024 * 1024)
+            ImageFileSizeLimit()
```
Daverball: I would probably keep this limit intact, it also might make sense to keep resizing this image if it exceeds the pixel size limit. If this particular limit ever becomes an issue, the customers will tell us.

src/onegov/org/forms/event.py
```diff
@@ -128,7 +129,7 @@ class EventForm(Form):
         file_class=EventFile,
         validators=[
             Optional(),
-            FileSizeLimit(5 * 1024 * 1024)
+            ImageFileSizeLimit()
```
Daverball: I think that's even more true for this one, since it will be displayed in the listing

Tschuppi81:
Not fully sure if I got your request. I now introduced `ImageFileSizeLimit` validator for image uploads in forms and keep the automatic resizing for image collection

src/onegov/org/forms/parliamentarian.py
```diff
@@ -52,7 +53,7 @@ class ParliamentarianForm(NamedFileForm):
     picture = UploadField(
         label=_('Picture'),
         fieldset=_('Basic properties'),
-        validators=[],
+        validators=[ImageFileSizeLimit()],
```
Daverball: We realistically probably should restrict the dimensions of this as well.

I think I would probably rename this validator to `ImageSizeLimit`, add a second parameter for maximum dimensions and then automatically resample the image, when it exceeds those limits, so it never even causes a validation error, unless the resampled image somehow still exceeds the bytes limit.

Daverball (commented):
I don't think we should change the limits of special-purpose image fields we defined ourselves, I think we should also still resize them, ideally using the new validator you defined.

src/onegov/org/forms/parliamentarian.py
Tschuppi81: I like the proposed class name `ImageSizeLimit`!

Daverball (approved):

PR #2531, opened 2026-06-16
Core: Fix before_cursor_execute handlers blocking ROLLBACK TO SAVEPOINT on INERROR connections

Tschuppi81:
Core: Fix `before_cursor_execute` handlers blocking `ROLLBACK TO SAVEPOINT` on `INERROR` connections

Both `activate_schema` and `limit_session_lifetime` fire on every cursor execution, including the internal `ROLLBACK TO SAVEPOINT` issued by SQLAlchemy's `begin_nested()` recovery. On a connection already in `INERROR` state their inner `cursor.execute()` calls raise `InFailedSqlTransaction` before the `ROLLBACK` statement ever reaches PostgreSQL, leaving the connection permanently stuck in `INERROR`.

Fix: check whether the statement parameter starts with `ROLLBACK TO SAVEPOINT` at the top of both handlers and return early. 

TYPE: Bugfix
LINK: ogc-3223


Daverball:
I recently changed the order of the sentry tween, can you check whether that had something to do with this? Does the issue still occur if the sentry tween is once again installed over `excview_tween`?

It's possible that since the sentry tween catches `DB_CONNECTION_ERRORS` and returns normally, we somehow skip `manager.abort`, that would explain why this happens for statement timeouts, although I don't believe that should be possible. Maybe this happens when the session never gets added to the transaction in the first place because the DB error happens before the `after_attach` event on the session gets fired?

Another possibility is that `session.close()` does not reset the raw DB transaction.

This workaround is technically fine, but it would be better to know where the broken connection comes from and prevent it from happening in the first place, rather than recovering from it, when it does happen.

Tschuppi81:
Thanks for the inputs. Yes, my second step will be to reproduce the issue while uploading vote results.

src/onegov/core/orm/session_manager.py
```diff
@@ -409,6 +410,29 @@ def register_engine(self, engine: Engine) -> None:
 
         """
 
+        @event.listens_for(engine, 'checkout')
+        def rollback_aborted_connection(
```
Tschuppi81: `rollback_aborted_connection` is just safety net for any connection in the pool left in `INERROR`.

src/onegov/core/orm/session_manager.py
```diff
@@ -421,6 +445,12 @@ def activate_schema(
 
             """
 
+            # Skip on INERROR: SET search_path would block ROLLBACK
+            # TO SAVEPOINT.
+            if (cursor.connection.get_transaction_status()
+                    == psycopg2.extensions.TRANSACTION_STATUS_INERROR):
+                return
```
Daverball: This seems pretty heavy handed and potentially dangerous to me, since if this happens with some arbitrary query that actually wants to change something, it now might happen in the wrong schema, if the transaction error somehow gets resolved by another event handler before the actual execution of the query.

I think we should make this guard a lot more narrow and only skip this if the `statement` parameter to `activate_schema` is specifically `ROLLBACK TO SAVEPOINT`. But I'm also quite surprised that a savepoint rollback triggers this hook in the first place. It certainly doesn't appear to happen all of the time, since otherwise all of the tests that involve a rollback should fail. Also if this is our culprit, do we want the safety net above? The safety net seems harmful in the long run to me, since it will hide similar issues in the future, a warning log is not good enough, at the very least it should be an error log. I think I'd rather see the error in Sentry again, if something changes and causes a similar issue. It will also give us better peace of mind,  that the issue is actually fixed by this change. But I'm also a little worried about auto roll back causing subtle issues in complex scenarios, so I think I would prefer for it to not be there. But we can certainly log it, so we know it's happening and don't have to investigate as much if this regresses in the future.

We can also improve the type hints for all our event handlers here, the SQLAlchemy docs tell you the exact types and names of all the parameters that are passed. `Any` is a relic from SQLAlchemy 1.3, where we didn't have accurate type hints for those parameters anyways, so `Any` was fine as a fallback.

Daverball (approved):

src/onegov/core/orm/session_manager.py
```diff
@@ -412,15 +418,28 @@ def register_engine(self, engine: Engine) -> None:
...
+            parameters: _DBAPIAnyExecuteParams,
+            context: ExecutionContext | None,
+            executemany: bool,
         ) -> None:
             """ Share the 'info' dictionary of Session with Connection
             objects.
 
             """
 
+            if statement.startswith('ROLLBACK TO SAVEPOINT'):
```
Daverball: Actually, this can probably be `startswith('ROLLBACK')` since a full transaction rollback should also succeed, if there are no savepoints.

Tschuppi81: good catch!

PR #2538, opened 2026-06-19
Adds hash per directory entry

Tschuppi81:
Directory: Adds has per directory entry

Each directory entry holds a `SHA1` hash of all field values and associated file checksums. The hash is recalculated on every change/update. Also the hash is visible on the api.

TYPE: Feature
LINK: ogc-3096


src/onegov/directory/models/directory_entry.py
```diff
@@ -130,3 +140,35 @@ def values(self, values: dict[str, Any]) -> None:
...
+    def update_content_hash(self) -> None:
+        # Must be called explicitly after any mutation to values or files.
+        # Directory.update() does this; bypass it and the hash goes stale.
+        values_part = json.dumps(self.values, sort_keys=True, default=str)
+        files_part = '|'.join(sorted(
+            f.checksum or f.id for f in self.files
+        ))
+        new_hash = hashlib.sha256(
+            f'{values_part}:{files_part}'.encode()
+        ).hexdigest()
```
Daverball: This is pretty inefficient and involves a lot of copying and converting bytes, only to get eventually consumed by the hash algorithm.

`sha1` should be fine for our use-case here and it's quite a bit faster than `sha256`, so let's use that, you should also add the `usedforsecurity=False` argument, so bandit doesn't complain about the inadequate hash strength for security purposed.

You can continually feed a hash object additional bytes via its `update` method, this means you don't have to build one giant bytes object first.

JSON is also a pretty expensive serialization algorhithm for this use-case, I think msgpack makes more sense here and will give us bytes from the start.

```suggestion
        hash_obj = hashlib.sha1(usedforsecurity=False)
        hash_obj.update(ormsgpack.packb(
            self.values,
            default=str,
            options=ormsgpack.OPT_SORT_KEYS
        ))
        for file_part in sorted(f.checksum or f.id for f in self.files):
            hash_obj.update(file_part.encode())
        new_hash = hash_obj.hexdigest()
```

src/onegov/org/templates/directory_entry.pt
```diff
@@ -59,6 +59,15 @@
                         <dd i18n:translate>Found an error? Propose a change to this entry.</dd>
                     </dl>
                 </div>
+                <div class="borderless-side-panel" tal:condition="entry.content_hash">
+                    <h2 i18n:translate>Hash</h2>
```
Daverball: Is this an appropriate user-facing name? Do we want to expose this on the user-interface by default or do we want to make this a setting? Do anonymous users need to see this? If so, do they need to be able to see it on the listing as well, so they don't have to mTAN login to see if something has changed?

src/onegov/org/api.py
```diff
@@ -693,6 +693,7 @@ def item_data(self, item: ExtendedDirectoryEntry) -> dict[str, Any]:
 
             data['coordinates'] = get_geo_location(item)
             data['contact'] = item.contact
+            data['content_hash'] = item.content_hash
```
Daverball: Is this truly only part of the public data? It seems relatively harmless to me to always display the hash, even if the item has `mtan` access.

Daverball (changes requested):
Looks good overall, but I think we can do a lot better, when it comes to performance, this should also speed up the initial migration.

Since the use-case is pretty specific to building permit applications it's also questionable whether this should be visible in the UI by default, or if it should be a setting. If we want to always show it, we may need to think about a more user-friendly name and potentially some hover text that explains what this is about.

Daverball (approved):

src/onegov/directory/models/directory_entry.py
```diff
@@ -130,3 +141,37 @@ def values(self, values: dict[str, Any]) -> None:
...
+def set_content_hash_before_flush(
+    session: Session,
+    flush_context: object,
+    instances: object
+) -> None:
+    # Safety net: ensure entries always have a hash even if update_content_hash
+    # was not called by the caller (new entry added directly, or hash cleared).
+    for obj in (*session.new, *session.dirty):
+        if isinstance(obj, DirectoryEntry) and obj.content_hash is None:
+            obj.update_content_hash()
```
Daverball, line 177: I just spotted this. This is pretty heavy handed and will always run, regardless of whether we even use directories in a given application.

I think an observer on `content` and `files` makes more sense here, then you also don't need to manually call `update_content_hash` in `Directory.update`.

Daverball (changes requested):
The event handler is a little bit much, I think we should use an observer instead.

src/onegov/directory/models/directory_entry.py
Daverball: You can also avoid triggering `obj.update_content_hash()` if `self.content_hash is not None` and neither `'values' in content` nor `inspect(self).state.attrs.files.history.has_changes()` is true.

Technically you should also need to observe `files.checksum`, since the reference on an individual file could change without triggering changes on `files`, but as far as I know, replacing a file upload in a directory entry deletes the old file and adds a new one. We never update just the reference, so you can get away with a simpler observer.

Tschuppi81:
Follow-up [here](https://github.com/OneGov/onegov-cloud/pull/2545) as this got `merged`

PR #2542, opened 2026-06-25
Allow to search for people

Tschuppi81:
People: Adds capabilities to search for people

TYPE: Feature
LINK: ogc-3228

### DIFF
--- a/src/onegov/people/collections/people.py
+++ b/src/onegov/people/collections/people.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
-from sqlalchemy import func, or_, type_coerce
-from sqlalchemy_utils import escape_like
+from sqlalchemy import and_, func, type_coerce
 
 from onegov.core import utils
 from onegov.core.collection import GenericCollection
 from onegov.core.orm.types import JSON
 from onegov.people.models import Person
+from onegov.search import SearchIndex
+from onegov.search.utils import language_from_locale
 
 from typing import Any
 from typing import TYPE_CHECKING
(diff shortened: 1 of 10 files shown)


src/onegov/people/collections/people.py
```diff
@@ -109,4 +110,11 @@ def people_by_organisation(
                     type_coerce([f'-{sub_org}'], JSON)
                 )
             )
+        if search:
+            term = f'%{search}%'
```
Daverball: You need to escape this so that `_` and `%` in search terms are treated literally. See e.g. `Search.suggestions`, which also makes use of `escape_like`, make sure to pass `'*'` as a second parameter to `ilike` so it is for sure recognized as an escape character.

I also wonder whether doing a full text search using the search index would make more sense here, instead of limiting ourselves to three fields. See e.g. `NewsCollection` which has  a `term` parameter, which does a full text search by joining to the search index.

Tschuppi81: Agree, then the `fts_properties` will decide what is searchable.

Currently:
```
    fts_properties = {
        'title': {'type': 'text', 'weight': 'A'},
        'function': {'type': 'localized', 'weight': 'B'},
        'email': {'type': 'text', 'weight': 'A'},
        'phone_fts': {'type': 'text', 'weight': 'A'},
    }
```

Daverball:
You may also want to consider only displaying the search field and using its value if `request.app.fts_search_enabled`. I think I'm already doing this in the API and possibly some places where you can search in the frontend as well.

src/onegov/people/collections/people.py
```diff
@@ -82,19 +85,18 @@ class PersonCollection(BasePersonCollection[Person]):
     def model_class(self) -> type[Person]:
         return Person
 
+    def query(self) -> Query[Person]:
+        return super().query().order_by(Person.last_name, Person.first_name)
```
Daverball: I don't think we want to change the order on the base query, since it makes a lot of queries slower.

I think we should keep the old `people_by_organisation` method and just remove the `.all()` at the end, so you return the query instead of a list.

It might also be worth considering moving these two methods into helper functions in the Org view, since they're technically both Org specific, they don't necessarily belong to the base collection, since Agency will not use these.

Daverball (approved):

PR #2545, opened 2026-06-25
Directory: Add content hash per entry

Tschuppi81:
Directory: Adds hash per directory entry

Each directory entry holds a `SHA-1` hash of all field values and associated file checksums. The hash will be updated on any change to either `content` or `files` implemented through an observer. 
The hash is visible for `managers` and `mTAN` authenticated users. Also the hash is exposed on the API.

TYPE: Feature
LINK: ogc-3096


Tschuppi81:
Even with the `observer` we cannot skip `entry.update_content_hash()` in `Directory.update()` as the test `test_content_hash_changes_on_new_field` shows

src/onegov/directory/models/directory.py
```diff
@@ -464,6 +464,8 @@ def get_value_field_from_note(file_id: str) -> Any:
         if session and not session._flushing:
             session.flush()
 
+        entry.update_content_hash()
+
```
Daverball: We can get rid of this now.

src/onegov/directory/models/directory_entry.py
```diff
@@ -130,3 +141,30 @@ def values(self, values: dict[str, Any]) -> None:
...
+                )
+            self.content_hash = new_hash
+
+    @observes('content', 'files')
+    def content_hash_observer(
+        self,
+        content: dict[str, Any] | None,
+        files: set[Any]
+    ) -> None:
+        self.update_content_hash()
```
Daverball, line 170:
```suggestion
        if (
            self.content_hash is None
            or 'values' in content
            or inspect(self).state.attrs.files.history.has_changes()
        ):
            self.update_content_hash()
```
We can avoid recomputing the hash, when the hash already exists and we know nothing relevant has changed.

Daverball (commented):
Looks good, just a couple of suggested improvements.

src/onegov/directory/models/directory.py
Tschuppi81: I tried, but test `test_content_hash_changes_on_new_field` shows it cannot be removed

Daverball: That test is missing a `session.flush()` after the first add. It should work, if it doesn't, the conditions on the observer are wrong.

I'm also not quite sure about not triggering the hash, if `content is None`, since removing all of the values should change the hash. Maybe it's better to remove the guard, for some reason I thought the observer gave you a set of keys that changed for `MutableDict` columns, but I guess it gives you just the actual column value, which doesn't help with detecting changes.

Tschuppi81: Although the additional `session.flush` in the test the `update_content_hash` is needed. At least I was able to move it into migration step.

Maybe its is due to migration and observer run in  `after_flush`?

Daverball: If they're both triggered in `after_flush` it's possible they're happening in the wrong order, yes.

Daverball (approved):

PR #2548, opened 2026-06-29
Directory: Admin notification emails for building permit workflow

Tschuppi81:
Directory: Admin notification emails for building permit workflow

Admins can get notified when publication starts and ends with all the details and hash to prove publication.

TYPE: Feature
LINK: ogc-3097


src/onegov/directory/models/directory_entry.py
```diff
@@ -149,6 +153,12 @@ def update_content_hash(self) -> None:
             default=str,
             option=ormsgpack.OPT_SORT_KEYS
         ))
+        for keyword in sorted(self.keywords):
+            hash_obj.update(keyword.encode())
+        if self.publication_start is not None:
+            hash_obj.update(self.publication_start.isoformat().encode())
+        if self.publication_end is not None:
+            hash_obj.update(self.publication_end.isoformat().encode())
```
Daverball: Are we sure we want these to be part of the content hash? And if so, why stop there? Why not the access as well or other content extensions? Publication seems especially questionable, since we already send notifications for when an entry has finished publication. But keywords I'm not so sure about either, since they're more about categorization, than meaningfully contributing to the content of entry in most cases.

src/onegov/org/cronjobs.py
```diff
@@ -249,6 +254,49 @@ def handle_publication_models(request: OrgRequest, now: datetime) -> None:
             send_email_notification_for_directory_entry(
                 obj.directory, obj, request)
 
+        if (
+            isinstance(obj, ExtendedDirectoryEntry)
+            and obj.publication_end is not None
+            and now >= obj.publication_end
+        ):
+            send_admin_expiry_notification_for_directory_entry(
```
Daverball: Why only send one for expiry? If we defer publication, we may also want to know when the entry was published.

src/onegov/org/cronjobs.py
```diff
@@ -249,6 +254,49 @@ def handle_publication_models(request: OrgRequest, now: datetime) -> None:
...
+            ExtendedDirectoryEntry.content_hash
+            != ExtendedDirectoryEntry.notified_hash
+        ),
+        ExtendedDirectory.meta['notification_address'].astext.isnot(None),
+    )
+
+    for entry in entries:
+        directory = entry.directory
+        is_update = entry.notified_hash is not None
+        send_admin_notification_for_directory_entry(
```
Daverball: One E-Mail per entry seems a bit much, if it's a directory that seen frequent changes, can we group these by directory and send a summary of the changes instead if more than one entry has changed?

We could even go further and group all notifications to the same `notification_address`. Although it'll be more difficult to come up with a good subject line.

Daverball (commented):
The overall direction seems fine to me, but I'm not sure about some of the details.

src/onegov/org/cronjobs.py
Tschuppi81: Yes, finding the right subject will be difficult for grouped notifications. And for a directory with frequent changes it will only be sent one an hour...

src/onegov/directory/models/directory_entry.py
Tschuppi81: I was manually testing yesterday and stumbled over it when changing a category and or publication dates. I think the publication dates might be more important since we build it for the permit use case.

Changing publication dates might be rare, but important to me

src/onegov/org/cronjobs.py
Tschuppi81: Notification for published entries are in place and verified with `test_admin_notification_publication_date_reached`

src/onegov/directory/models/directory_entry.py
Daverball: Is it important though, if we already send a notification for publication and de-publication?

The only potential problem I see is things like: Entry gets created with publication next Monday, then next Tuesday the publication gets modified and set further into the future. So now you get two published notifications and you don't know for which period between the two notifications the entry wasn't published.

Or the entry gets manually unpublished by setting a publication end in the past.

But similar things can happen if the access gets changed to private.

My main objection is conceptual, since changing the publication dates does not really change the contents of the entry, it just changes its visibility. We may want to detect changes to visibility separately, since that's a different use-case from tracking changes to the contents of an entry.

src/onegov/org/cronjobs.py
Daverball: Yeah, but grouping by directory makes sense to me. If in an hour 20 entries get published, you don't want 20 e-mails, you want one e-mail summarizing the changes. In most cases that won't happen, but that's no reason not to guard against it.

src/onegov/org/cronjobs.py
Daverball: Indeed, I missed that you were filtering by `published` in the other notifications.

Daverball: You should probably clarify in the setting that notifications will only be sent during an entry's publication period. Otherwise people will be confused, why they're not receiving any notifications, when playing around with an unpublished entry.

Daverball:
I wonder if this use-case could be served more reliably by locking entries during publication, so in order to make changes during publication they explicitly have to temporarily unpublish the entry, to remove the lock. Because then, all we need are publication notifications, for these more strictly locked entries.

Daverball:
We could then also emit a validation error, if they try to submit a new entry without setting publication into the future, so we always get a publication and de-publication notification. (Manually unlocking the entry would still have to be its own notification, outside the publication cronjob though)

Daverball:
Although maybe it's fine, if all they have to do is yet again set a publication date into the future, in order to submit changes to the locked entry. Since that way you will get a second publication notification.

Tschuppi81:
hash calculation:
- content basic fields
- files using existing file hash
- ~publication dates~
- ~keywords~

now observing `publication_start` to reset `notified_hash` in order to trigger a republication email

added all fields, files (name, upload date, file hash), publication dates, access and overall hash to notification email

src/onegov/directory/models/directory_entry.py
```diff
@@ -83,9 +83,13 @@ class DirectoryEntry(Base, ContentMixin, CoordinatesMixin, TimestampMixin,
     #: Describes the entry briefly
     lead: Mapped[str | None]
 
-    #: SHA-1 hash of the entry values and associated file checksums
+    #: SHA-1 hash of the entry values, keywords, publication dates, and
+    #: associated file checksums
     content_hash: Mapped[str | None] = mapped_column(default=None)
 
+    #: content_hash at the time of the last admin notification
+    notified_hash: Mapped[str | None] = mapped_column(default=None)
```
Daverball: This part of the solution seems a little over-engineered to me now. I think we can get rid of this column, since if the entry is immutable, we know that all we have to do is emit a notification within `handle_publication_models`, for all `ExtendedDirectoryEntry` (when the end date is missing or in the future, we know it was just published, otherwise it was just unpublished, we no longer have to care about whether or not the content has changed, since we know it can only change outside publication windows)

src/onegov/org/cronjobs.py
Daverball: I think we can revisit this now and put all of the notifications here. We just have to split it into two if statements:

```python
        if isinstance(obj, ExtendedDirectoryEntry):
            if obj.publication_end is None or obj.publication_end < now:
                # published notification
                ...
            else:
                # expired notification
                ...
```

src/onegov/org/cronjobs.py
```diff
@@ -249,6 +254,49 @@ def handle_publication_models(request: OrgRequest, now: datetime) -> None:
...
+        if (
+            isinstance(obj, ExtendedDirectoryEntry)
+            and obj.publication_end is not None
+            and now >= obj.publication_end
+        ):
+            send_admin_expiry_notification_for_directory_entry(
+                obj.directory, obj, request)
+
+
+def handle_directory_admin_notifications(request: OrgRequest) -> None:
```
Daverball: Which means we can get rid of this function as well

src/onegov/org/cronjobs.py
```diff
@@ -849,6 +897,10 @@ def delete_content_marked_deletable(request: OrgRequest) -> None:
                     if obj.publication_end and obj.publication_end < now:
                         log.info(f'Cron: Delete expired obj for {name}: '
                                  f'{obj.title}')
+                        if isinstance(obj, ExtendedDirectoryEntry):
+                            send_admin_deletion_notification_for_directory_entry(
+                                obj.directory, obj, request,
+                                auto_deleted=True)
```
Daverball: Since this entry is already expired, we don't need a notification when it is deleted, especially when it happens automatically, it would only cause confusion.

src/onegov/org/views/directory.py
```diff
@@ -917,6 +1050,11 @@ def delete_directory_entry(
 
     request.assert_valid_csrf_token()
 
+    if isinstance(self, ExtendedDirectoryEntry):
+        send_admin_deletion_notification_for_directory_entry(
+            self.directory, self, request
+        )
+
```
Daverball: We don't need a notification here, we need to disallow it, while the entry is published (as long as the directory is configured to block changes on published entries)

Daverball (changes requested):
I think we can simplify things a lot, the solution is a little over-engineered now for what we actually need.

src/onegov/org/forms/directory.py
```diff
@@ -340,6 +340,17 @@ class DirectoryBaseForm(Form):
...
 
+    notification_address = EmailField(
+        label=_('Admin notification address'),
+        description=_(
+            'Receives an e-mail when entries are published or de-published '
+            '(sent out hourly). Useful e.g. for permit directories as proof '
+            'of publication.'
+        ),
+        fieldset=_('Notifications'),
+        validators=[Optional(), Email()]
```
Daverball: This field should probably depend on `required_publication`:

```suggestion
        depends_on=('required_publication', 'y'),
        validators=[Optional(), Email()]
```

You could also argue that `enable_publication` is enough, but then we have no protection against deleting `publication_start`.

src/onegov/org/views/directory.py
Daverball, line 1012: We need to guard against currently published entries, that have publication notifications enabled here as well and raise `HTTPForbidden`, as long as changes are locked.

You should also replace the delete link in the layout with an unconditional alert that tells them that the entry is currently locked and that they need to unpublish the entry first in order to delete it.

The check should probably take into account `hourly_maintenance_task_last_run`, so it only counts the model as unpublished if it was unpublished before the last run.

Daverball (commented):
We need a couple of additional guard rails to make sure this is as robust as possible. Deleting published entries should not be possible when the notifications are turned on.

We also need to consider if it makes sense to allow change requests when this setting is enabled. We probably should make the two settings mutually exclusive, since otherwise you could bypass the mutability restrictions via change requests.

In the same vain it doesn't make sense to show the notification setting if publication dates are neither enabled nor required. So the notification setting should depend on the other two settings.

src/onegov/org/views/directory.py
```diff
@@ -105,6 +110,19 @@ def on_request(self) -> None:
                 self.publication_start.validators[0] = InputRequired()
                 self.publication_end.validators[0] = InputRequired()
 
+        def validate_publication_start(self, field: Field) -> None:
+            if (
+                field.data is not None
+                and isinstance(model, ExtendedDirectoryEntry)
+                and model.published
```
Daverball: Actually I just noticed this. If changing the date would change `model.published` this check will be incorrect and can be bypassed.

You have to take into account both the submitted start and end date, the original start and end dates are not relevant here.

Tschuppi81: Well spotted!

src/onegov/org/views/directory.py
```diff
@@ -670,6 +769,18 @@ def handle_new_directory_entry(
...
+            # falls after its last run, so a backdated start (e.g. a permit
+            # issued earlier) would never be caught - send it directly here
+            last_run = request.app.org.last_hourly_maintenance_run
+            if (
+                entry.publication_start is None
+                or entry.publication_start <= last_run
+            ):
+                send_admin_notification_for_directory_entry(
+                    self.directory, entry, request)
+
```
Daverball: A backdated start is not allowed, so this should not be necessary

src/onegov/org/cronjobs.py
```diff
@@ -251,6 +254,15 @@ def handle_publication_models(request: OrgRequest, now: datetime) -> None:
             send_email_notification_for_directory_entry(
                 obj.directory, obj, request)
 
+        if isinstance(obj, ExtendedDirectoryEntry):
+            if obj.publication_end is None or obj.publication_end > now:
+                if obj.directory.notification_address:
+                    send_admin_notification_for_directory_entry(
+                        obj.directory, obj, request)
```
Daverball:
```suggestion
        if isinstance(obj, ExtendedDirectoryEntry) and obj.directory.notification_address:
            if obj.publication_end is None or obj.publication_end > now:
                send_admin_notification_for_directory_entry(
                    obj.directory, obj, request)
```
Let's lift this condition out, it should also apply to the `else`

src/onegov/org/cronjobs.py
Daverball, line 237: Since we may need to have access to `directory` we should add a `joinedload` option to the query, so we don't end up with a N+1 query when a lot of entries are published at the same time.

```suggestion
        )
        if issubclass(model, DirectoryEntry):
            query = query.options(joinedload(model.directory))
```

Daverball (commented):
Looks good, just a couple of small things to cleanup/improve

src/onegov/org/views/directory.py
Tschuppi81: Backdating is not disallowed, `ensure_publication_start_end` verifies end is in the future and start < end.
https://github.com/OneGov/onegov-cloud/blob/4cc06c17e460f22bff32b60e1f3b8fa73cefad11/src/onegov/org/forms/extensions.py#L271

The main case here is anyways the immediate publish when no publication start is set.

Anyways, i updated the comment to list the two cases clearly

Daverball: Well if it isn't disallowed, that's a bug, it should be disallowed when a notification address is set. Otherwise you're lying about when it was first published.

Tschuppi81: Correct, backdating for this feature is a bug, but ok for the existing use cases

Tschuppi81:
> We need a couple of additional guard rails to make sure this is as robust as possible. Deleting published entries should not be possible when the notifications are turned on.
> 
> We also need to consider if it makes sense to allow change requests when this setting is enabled. We probably should make the two settings mutually exclusive, since otherwise you could bypass the mutability restrictions via change requests.
> 
> In the same vain it doesn't make sense to show the notification setting if publication dates are neither enabled nor required. So the notification setting should depend on the other two settings.

All done now

Daverball (approved):

PR #2550, opened 2026-06-29
Adds reservation cancellation path to resources

Tschuppi81:
Reservation: Adds reservation cancellation path to resources

TYPE: Feature
LINK: ogc-3064


Daverball (commented):
Currently this seems to be limited to cancelling all reservations in a ticket, which should cover individual reservations nicely, but does that cover 90+% of customer requests or is it more common, or almost as common that for large half year reservations, they want to cancel a single date?

If so, we probably need to allow customers to select which dates they want to cancel, which would also necessarily change the UI in the ticket.

Tschuppi81:
Now cancellation requests can select single reservations to be cancelled and the manager sees the request cancellation dates, then approve it.

src/onegov/org/forms/resource.py
```diff
@@ -380,6 +380,21 @@ class ResourceBaseForm(Form):
...
+    allow_cancellation_requests = BooleanField(
+        label=_('Enable cancel reservation'),
+        fieldset=_('Cancellation'),
+        default=False,
+    )
+
+    def validate_allow_cancellation_requests(
+        self, field: BooleanField
+    ) -> None:
+        if field.data and not self.reply_to.data:
```
Daverball: The `Resource.reply_to` will fallback to `Organisation.reply_to`, so I'm not sure it's correct to prevent validation if there's no resource-specific address. And even if they don't have an Organisation `reply_to` address either, that shouldn't really change things, as long as supporters receive a notifcation if notifications are turned on and there's a corresponding message in the activity feed, that should be good enough.

I don't think we should enforce a workflow, so we should probably get rid of this validation.

src/onegov/org/models/ticket.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
```
Daverball:
```suggestion
from onegov.custom import json
```

Apart from a few exceptions, where unicode escape sequences are undesirable, we should use our own json encoder.

src/onegov/org/views/reservation.py
```diff
@@ -1483,12 +1517,18 @@ def email_iter() -> Iterator[EmailJsonDict]:
...
+                if isinstance(kaba_exc, KabaApiError):
                     log.info('Kaba API error', exc_info=True)
                 failed_to_revoke = True
 
-        if len(targeted) > 1:
+        if change == 'cancellation_accepted':
+            ticket.handler_data = {
+                k: v for k, v in ticket.handler_data.items()
+                if k != 'cancellation_requested'
+            }
```
```suggestion
            ticket.handler_data.pop('cancellation_requested', None)
```
A comprehension to remove a single-key is a bit overkill, we can just use `pop` with a default, if we want to delete in a single key without having to guard against `KeyError`.

You may need to invoke `mark_changed(ticket, 'handler_data')` this way, although I don't think it should be necessary, `handler_data` should always be a SQLALchemy `MutableDict` at this point, which should trigger changes on `pop`.

src/onegov/org/views/reservation.py
```diff
@@ -1597,6 +1645,124 @@ def reject_reservation_with_message_from_ticket(
...
+            request.alert(
+                _('A cancellation request has already been submitted.')
+            )
+        else:
+            request.alert(_('A cancellation request can only be submitted for '
+                            'accepted reservations.'))
+        return morepath.redirect(request.link(self, 'status'))
+
+    payment = self.handler.payment
+    if payment and payment.state in ('invoiced', 'paid'):
```
Daverball, line 1694: This condition may be undesirable, since refunds are a thing, but I suppose we can leave it for now, just so it's less easy to cancel a reservation without issuing a refund.

src/onegov/org/views/reservation.py
```diff
@@ -1597,6 +1645,124 @@ def reject_reservation_with_message_from_ticket(
...
+        self.last_state_change = self.timestamp()
+        self.state = 'open'
+        self.user = None
+
+        ReservationMessage.create(
+            targeted,
+            self,
+            request,
+            'cancellation_requested'
+        )
```
Daverball, line 1714: If notifications are enabled, we should probably also send an e-mail notification to the ticket's assignee, letting them know about the cancellation request. See `send_chat_message_email_if_enabled`.

We can think about whether or not this needs to be a new setting for resource recipients, so they can opt in/out specifically this type of notification.

Daverball (commented):
Looks good overall, although we should consider if and how we should notify a ticket's assignee and/or resource recipients.

There's also couple of minor improvement suggestions and a couple of questionable conditions, that make the feature less flexible.

src/onegov/org/forms/resource.py
Tschuppi81: Agree, there will be a ticket anyway who ever gets notified

src/onegov/org/views/reservation.py
Tschuppi81: i added a comment that we don't handle the refund case yet

src/onegov/org/views/reservation.py
```diff
@@ -1695,17 +1697,37 @@ def request_cancellation(
...
+                    'resource': resource,
+                    'reservations': targeted,
+                    'ticket_reference': self.reference(request),
+                },
+                force=True,
+            )
+
+        self.last_state_change = self.timestamp()
+        self.state = 'open'
+        self.user = None
```
Daverball: I think it only makes sense to unassign the ticket and reopen it, if it has already been closed. If it's pending, it should stay pending.

In either case we may want to notify resource recipients in addition to assignees, just like we do when a customer writes a message into a pending ticket. Also keep in mind that there may not be an assignee, in which case you can only send an e-mail to resource recipients. Although we may want to add a new setting on resource recipients, so they can be subscribed to this specific kind of notification, but we may want to write a migration that subscribes all recipients that currently subscribe to customer messages, to also subscribe to cancellation requests.

Tschuppi81:
Now, as I review the changes i think the `reply_to` address shall not be notified about cancellation requests, as it is meant to reply to from an email. As we now have the ability to subscribe to cancellation request we should rather keep it separate.

src/onegov/org/forms/ticket.py
```diff
@@ -180,3 +183,35 @@ def on_request(self) -> None:
...
+
+    def on_request(self) -> None:
+        reservations = self.model.handler.reservations
+        if len(reservations) <= 1:
+            self.delete_field('reservation_ids')
+            return
+
+        self.reservation_ids.choices = [
+            (r.id, self.model.handler.get_reservation_title(r))
+            for r in reservations
```
Daverball, line 221: We probably should only include reservations that have a start date in the future here.

src/onegov/org/views/reservation.py
```diff
@@ -1597,6 +1643,170 @@ def reject_reservation_with_message_from_ticket(
...
+        request.alert(_('Cancellation requests are not enabled for this '
+                        'resource.'))
+        return morepath.redirect(request.link(self, 'status'))
+
+    if self.handler_data.get('cancellation_requested'):
+        request.alert(
+            _('A cancellation request has already been submitted.')
+        )
+        return morepath.redirect(request.link(self, 'status'))
+
```
Daverball, line 1679: You should also guard against all reservations being in the past, if all of them already happened, they can no longer be cancelled.

src/onegov/org/views/reservation.py
```diff
@@ -1597,6 +1643,170 @@ def reject_reservation_with_message_from_ticket(
...
+    if self.state not in ('pending', 'closed'):
+        request.alert(_('A cancellation request can only be submitted once '
+                        'the reservation has been accepted.'))
```
Daverball: I can see the internal workflow reason for disallowing this, but I don't think customers are going to like this, so I think we should allow it anyways. We can maybe just skip the additional notification if the ticket is still open, although it's probably fine if there are two notifications before anyone has been assigned to the ticket. The same thing can happen with customer messages.

The only time we may want to disallow it, is when the ticket has been archived. I'm not really sure what we do for archived tickets in the status view. But in all likelihood, by the time the ticket has been archived, all reservations are in the past anyways, so none of them can be cancelled.

src/onegov/org/views/reservation.py
```diff
@@ -1597,6 +1643,170 @@ def reject_reservation_with_message_from_ticket(
...
+                    yield res.address
+
+        # notify the ticket assignee (if any), the resource's reply-to address
+        # and the recipients registered for cancellation requests about the
+        # cancellation request. We force the mail so it is not suppressed by
+        # the auto-accept / self-notification heuristics in send_ticket_mail,
+        # but still honour the mute settings
+        receivers = (
+            *((self.user.username,) if self.user else ()),
+            *((resource.reply_to,) if resource.reply_to else ()),
```
Daverball: Yeah I don't think we should send a mail to this address, if they want one, they can just add a resource recipient.

src/onegov/org/views/reservation.py
```diff
@@ -1597,6 +1643,170 @@ def reject_reservation_with_message_from_ticket(
...
+            *recipients_registered_for_cancellation(),
+        )
+
+        if not request.app.org.mute_all_tickets or not self.muted:
```
Daverball: I actually don't recall if muting a ticket applies to messages sent to resource recipients, if it doesn't we may not want this condition. Even if it applies to some notifications, this might be one we may want to always send. What do we do for customer messages? We should probably do the same here.

src/onegov/org/upgrade.py
```diff
@@ -963,3 +964,18 @@ def recreate_missing_reserved_slots(context: UpgradeContext) -> None:
...
+) -> None:
+    # recipients who receive customer messages should also be notified about
+    # cancellation requests, which are a customer-initiated action
+    recipients = (
+        context.session.query(ResourceRecipient)
+        .options(undefer(ResourceRecipient.content))
+    )
+    for recipient in recipients:
+        if recipient.customer_messages:
+            recipient.cancellation_requests = True
```
Daverball: This could be done in a single batch update if you filter on `content->'customer_messages'::bool IS TRUE` and use `content = jsonb_set(content, 'cancellation_requests', 'true'::jsonb)` to update the content.

Generally it's better to avoid creating model instances in upgrade tasks, where possible.

Daverball (commented):
Looks good

There's still a couple of things we should consider. E.g. what happens if a customer requests a cancellation and by the time the ticket gets processed the reservation has already happened.

Technically it's harmless to cancel past reservations. But for billing purposes, if there's a mix of past reservations that were cancelled too late and future ones that can still be cancelled. You may want only the future ones to get cancelled, when cancellation gets approved. It might be fine to teach supporters to manually cancel reservations that can still be cancelled and reject the request, but it's definitely worth a discussion whether that's good enough, especially since the customer-facing notifications are going to be confusing, so they'd probably also want to mute the ticket and then send a manual message at the end.

src/onegov/org/views/reservation.py
Tschuppi81: `send_ticket_mail` checks for mute settings except if `force==True`
```
        if org.mute_all_tickets:
            return

        if ticket.muted:
            return
```

If I am right, customer messages are sent through `send_transactional_email_batch` which does not consider the mute settings, therefore sent anyways.

I will use the same pattern for registered cancellations

src/onegov/org/upgrade.py
Tschuppi81: It was not easy, I failed a few times :smiley_cat:

src/onegov/org/upgrade.py
```diff
@@ -970,12 +973,20 @@ def recreate_missing_reserved_slots(context: UpgradeContext) -> None:
...
+            ResourceRecipient.content['customer_messages'].as_boolean()
+            .is_(True),
+            ResourceRecipient.type == 'resource',
+        )
+    ).values(
+        content=func.jsonb_set(
+            ResourceRecipient.content,
+            array(['cancellation_requests']),
+            func.to_jsonb(true()),
+        )
```
Daverball: This is one of those cases where I would just write the update statement in raw SQL, it's far easier to get it right that way, we don't want any ambiguity about what the statement looks like at the end and whether or not it will have parameter bindings or is fully baked.

In this case the query should be fully baked, but I think this will probably result in a query with parameter bindings.

See e.g. the following upgrade task: https://github.com/OneGov/onegov-cloud/blob/e94d5f19c3d1547453f9b9d68d41798068556e83/src/onegov/reservation/upgrade.py#L153-L167

src/onegov/org/upgrade.py
Daverball: I believe you, it's not easy to get the types correct if you try to do it with the SQLAlchemy API, it's a lot easier if you just write the SQL. That way it's also very easy to test locally with `psql`.

Daverball (approved):

PR #2551, opened 2026-06-29
Ticket: status view orders the activity messages the same way as the …

Tschuppi81:
Ticket: status view orders the activity messages the same way as the ticket view

TYPE: Feature
LINK: None

### DIFF
--- a/src/onegov/org/templates/ticket_status.pt
+++ b/src/onegov/org/templates/ticket_status.pt
@@ -41,7 +41,7 @@
                 <div tal:condition="feed_data"
                     class="timeline"
                         data-feed-data="${feed_data}"
-                        data-feed-order="asc"
+                        data-feed-order="desc"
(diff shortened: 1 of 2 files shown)


Tschuppi81:
@BreathingFlesh  Any reason against this?

Daverball:
I think the main reason the order is different in the status view, is that there is an inline reply field for customers, so it makes sense to see the newest messages as the top, close to where you type your reply.

So I think you would need to move the field to to the bottom of the feed, if you want to change the order, but it will be easier to miss this way, so I think it's better the way it currently is, even if is a little confusing if you look at both views side-by-side. But that shouldn't really be a common use-case anyways. The status view is mainly intended for customers and any other people who don't have enough permissions to view the ticket itself.

Tschuppi81:
leave it as is

PR #2556, opened 2026-07-01
Prevent removing event filters that are still applied to events

Tschuppi81:
Org: Prevent removing event filters that are still applied on events

TYPE: Bugfix
LINK: ogc-3237


Daverball:
This feels a little heavy-handed. Depending on the number of affected events, it doesn't feel particularly reasonable to expect users to manually remove filters from each event in order to delete a specific filter key or value.

It's probably better if the validation error can be overruled, and when done so, we clean up the filters on existing events for them (You could implement this by adding a Checkbox after the rules that is only visible when this validation error is emitted, when the checkbox is set this particular step of the validation is skipped).

Tschuppi81:
The auto cleanup option looks like
Screenshot from 2026-07-01 10-38-41

Auto cleanup works nicely

Daverball (approved):

PR #2564, opened 2026-07-06
Core: Fixes HTTP 500 on null bytes in filter query parameters

Tschuppi81:
Core: Guards query-parameter filters against null bytes

TYPE: Feature
LINK: https://seantis-gmbh.sentry.io/issues/7595041870

### DIFF
--- a/src/onegov/core/utils.py
+++ b/src/onegov/core/utils.py
@@ -175,6 +175,39 @@ def normalize_for_filename(
     return sanitized
 
 
+#: Characters PostgreSQL cannot store in ``text`` / ``jsonb`` columns - null
+#: bytes and lone surrogates - which raise a ``DataError`` when bound as a
+#: query parameter. They only appear in malformed or malicious requests.
+_unstorable_text = re.compile(r'[\x00\ud800-\udfff]')
+
+
+def sanitize_query_param(value: object) -> str | None:
+    """ Returns ``value`` as a non-empty string safe to use as a database
+    query parameter, or ``None``.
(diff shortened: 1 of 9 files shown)


Tschuppi81:
Is it worth to filter those null bytes in query parameter? No direct risk, just unhandled 500 — but that makes it a cheap noise/low-grade DoS vector, so I guess still worth handling.

Daverball:
I don't think it's worth fixing. It's not even really a DOS vector. An unhandled exception takes up just as many resources as a handled one, in some cases it probably takes even less resources.

If it's only possible to trigger a 500 error through malicious manipulation of the url, so something a regular user will never encounter, then we don't need to worry about it. There's many other ways you can trigger an unhandled exception, if you really wanted to, like putting invalid UTF-8 byte patterns into the HTTP header. It's not a bad thing as long as it doesn't fully crash the application and affects subsequent requests.

There are some different cases where power users might manually manipulate the url and encounter crashes that way by misspelling some URL parameter. Those are the cases we should fix, since that usually means we need to add a converter for that URL parameter.

If I really wanted to fix this, I also wouldn't fix it in this way, since we would need to remember to apply sanitization to every single user-settable filter parameter manually. I instead would probably replace the identity converter morepath uses for `str` params with one that returns `None` for any strings that contain characters like `NUL` from the string, since we don't really have a valid use-case for allowing these weird byte patterns in url parameters anywhere.

Tschuppi81:
I agree, closing

PR #2570, opened 2026-07-09
Org: Round invoice to 0.05 before grand total

cyrillkuettel:

## Commit message

Org: Round invoice to 0.05 before grand total

<Optional Description>

TYPE: Feature
LINK: OGC-2957

## Checklist


src/onegov/org/forms/settings.py
```diff
@@ -2439,6 +2439,17 @@ class VATSettingsForm(Form):
...
+    price_rounding = BooleanField(
+        label=_('Round invoice totals to the nearest 0.05'),
+        description=_(
+            'If enabled, newly generated invoices include a '
+            'rounding position right before the total, so the '
+            'total is always a multiple of 0.05. Existing '
+            'invoices are not affected.'
+        ),
+        default=False,
+    )
```
Daverball: I would make this a `DecimalField` that is optional but with a minimum value of `0.01`, since we don't support more precision than that. For CHF a maximum of `1.0` would make sense, but for other currencies this may not be the case, so I don't think we need to restrict this any more than that.

You can round to sane prices using a helper function like this:

```python
from decimal import ROUND_HALF_UP

def round_amount(amount: Decimal, base: Decimal) -> Decimal:
    return base * (amount / base).quantize(Decimal('1'), rounding=ROUND_HALF_UP)
```

src/onegov/org/models/ticket.py
```diff
@@ -631,11 +641,21 @@ def invoice_items(self, request: CoreRequest) -> list[InvoiceItemMeta]:
...
             discounts = []
 
-        return self.resource.invoice_items_for_reservation(
-            self.reservations,
-            extras,
-            discounts,
-            reduced_amount_label=request.translate(_('Discount'))
+        return (
+
+                apply_price_rounding(
```
Daverball: It feels pretty clumsy and error-prone the way this works currently. Having to remember to use this utility function everywhere where we generate invoice items and sometimes passing in the manual total, isn't very elegant.

I think it makes more sense to put the rounding step into `refresh_invoice_items`, so the `invoice_items` functions can stay the same.

Basically in the first loop you would remember the existing rounding item, but not add it to `new_items` and remove it from the `unused` set. Then, just before we set `invoice.items = new_items` we calculate the total on `new_items` and apply the rounding, if the new total does not require rounding and there is an existing rounding item, then you need to delete the now unused item, the same way as the other unused items are deleted, if the rounding amount has changed, then you need to update the amount on the item and append the item to `new_items`. If the rounding item does not exist yet, then you need to create it the same way as other missing items are created.

Daverball: You will also need to adapt `refreshing_invoice_is_safe`, since `total_excluding_manual_entries` will not be necessarily correct. I'm a little torn here, since we can have old invoices that are unrounded, updating them can introduce rounding, which would change the total, but if changing the total is unsafe, then we should return `False` here.

The main issue is that this will introduce a bunch of complex logic to calculate what the new rounding amount should be. So your approach of applying the rounding, before we call this function is a little bit more robust there. It's just not very nice that we have to duplicate the rounding code into every ticket handler that can have an invoice.

One possibility would be to make `Handler` itself aware of the possibility that the total amount might get rounded and introduce a `rounded_invoice_items(self, request: CoreRequest, rounding_base: Decimal | None) -> list[InvoiceItemMeta]` method, that calls `invoice_items` and potentially adds the extra invoice item based on the newly generated invoice items and the manual items on a potentially already existing invoice.

However `refresh_invoice_items` and `refreshing_invoice_is_safe` would also need to accept the `rounding_base` parameter in order to make this work and it would still mean having to add to the total based on the rounding. But overall it still seems a little less error-prone than having to copy the helper call into every implementation of `invoice_items`.

It might also be worth renaming `invoice_items` to `base_invoice_items` and name the rounded version `invoice_items`. That way the type checker will complain about all the calls to `invoice_items` that are missing the `rounding_base` parameter.

src/onegov/org/views/reservation.py
```diff
@@ -634,11 +634,14 @@ def confirm_reservation(
         if failed and failed.isdigit()
     }
 
-    invoice_items = self.invoice_items_for_reservation(
-        reservations,
-        extras,
-        discounts,
-        reduced_amount_label=request.translate(_('Discount'))
+    invoice_items = utils.apply_price_rounding(
```
Daverball: You will probably need to do the same for the equivalent form submission view, which is used for `FRM`/`DIR` tickets. You can look at references to `invoice_items_for_submission` for the other places that need to change.

It's a little unfortunate, that we have to remember to apply this step, everywhere we calculate the total, prior to generating the real invoice. Maybe we can at least make it a helper `staticmethod` on `InvoiceItemMeta`, similar to `total`. So it's a little bit more discoverable.

Another option is to introduce a new utility type `InvoiceMeta`, which contains a list of all the invoice items, an optional rounding base and an optional reference to the existing invoice. Then we can move a lot of this logic into this helper class (It's `__iter__` could automatically yield the extra invoice item for rounding and it could have a helper function `total_changed` that takes into account rounding, that can then be used by `refreshing_invoice_is_safe`), including handling of manual invoice items and change a bunch of functions to return `InvoiceMeta` instead of `list[InvoiceItemMeta]`.

Daverball (commented):
Overall this seems fine. However you forgot to apply rounding to form and directory entry submissions.

I think the design could also use some refactoring, so we're less likely to forget to apply rounding. The most robust idea I have is introducing a new helper type `InvoiceMeta` that is aware of the potential existence of a rounding base and manual invoice items on an existing invoice and return that new helper type in a lot of places instead of `list[InvoiceItemMeta]`.

That way we can standardize a lot of the logic and bake it into this helper, instead of having to remember to call a utility function in half a dozen different places.

src/onegov/org/locale/de_CH/LC_MESSAGES/onegov.org.po
```diff
@@ -186,6 +186,23 @@ msgstr "Abzüge / Zuschläge"
 msgid "Lump sum"
 msgstr "Pauschalbetrag"
 
+msgid "Rounding"
+msgstr "Rundung"
```
Daverball: I would call this invoice item the "Rounding difference" i.e. "Rundungsdifferenz"

src/onegov/org/models/ticket.py
```diff
@@ -57,25 +58,32 @@ def ticket_submitter(ticket: Ticket) -> str | None:
...
-    request: CoreRequest
+    request: CoreRequest,
+    rounding_base: Decimal | None,
 ) -> None:
     payment = self.payment
     invoice = self.ticket.invoice
-    new_item_metas = self.invoice_items(request)
-    if not new_item_metas:
+    invoice_meta = self.invoice_items(request, rounding_base)
+    if not invoice_meta.items:
```
Daverball: I just realized this part is a little controversial, since we delete the invoice as soon as there are no automatic invoice items, regardless of whether are any manual items.

On the other hand, it's not possible to add manual invoice items when there is nothing to invoice, so maybe this is actually good for consistency. Otherwise we would have to consider always allowing manually adding invoice items to a ticket. Which could be useful, but for now I think we'll just leave it like this, until someone has an actual use-case..

src/onegov/org/templates/macros.pt
```diff
@@ -2483,6 +2483,9 @@
                         <tal:b tal:case="'migration'" i18n:translate>
                             Lump sum
                         </tal:b>
+                        <tal:b tal:case="'rounding'" i18n:translate>
+                            Rounding
```
Daverball:
```suggestion
                            Rounding difference
```

src/onegov/org/views/form_submission.py
```diff
@@ -227,9 +233,9 @@ def handle_pending_submission(
         # NOTE: The VAT amount can be wrong if the fee is charged to
         #       the customer. So it's better to not show it yet.
         'hide_vat_amount': True,
-        'invoice_items': group_invoice_items(invoice_items),
+        'invoice_items': group_invoice_items(list(invoice_meta)),
```
Daverball:
```suggestion
        'invoice_items': group_invoice_items(invoice_meta),
```
`group_invoice_items` accepts arbitrary iterables, so we don't need to convert to a list first.

src/onegov/org/views/form_submission.py
```diff
@@ -371,11 +381,11 @@ def handle_complete_submission(
                 )
                 TicketMessage.create(ticket, request, 'opened', 'external')
 
-                if invoice_items:
+                if invoice_meta.items:
```
Daverball: Maybe consider adding a `__bool__` to `InvoiceMeta`, so you can do `if invoice_meta` here instead (also the other places where you used the same condition).

src/onegov/org/views/reservation.py
```diff
@@ -660,7 +664,7 @@ def confirm_reservation(
         'complete_link': request.link(self, 'finish'),
         'edit_link': request.link(self, 'form'),
         'price': price,
-        'invoice_items': group_invoice_items(invoice_items),
+        'invoice_items': group_invoice_items(list(invoice_meta)),
```
```suggestion
        'invoice_items': group_invoice_items(invoice_meta),
```

src/onegov/org/constants.py
```diff
@@ -23,6 +23,7 @@
     'form': _('Selected Options'),
     'manual': _('Discounts / Surcharges'),
     'migration': _('Lump sum'),
+    'rounding': _('Rounding'),
```
Daverball:
```suggestion
    'rounding': _('Rounding difference'),
```

src/onegov/pay/utils.py
```diff
@@ -191,6 +198,83 @@ def refresh_item(self, item: InvoiceItem) -> None:
...
+    @cached_property
+    def total(self) -> Decimal:
+        """The total amount excluding manual items on an existing
+        invoice, but including the rounding item.
+        """
+        return InvoiceItemMeta.total(self)
+
+    @cached_property
+    def total_vat(self) -> Decimal:
+        return InvoiceItemMeta.total_vat(self)
```
Daverball: Overall it's maybe a little strange that these totals exclude the manual items from the existing invoice. It would make it possible to get a total that's not rounded, if the manual items change the rounding difference.

It might be worth to just include the manual totals here and add a separate `total_excluding_manual_entries` method, so we mirror the behavior of `Invoice`. There would still be a difference in `InvoiceMeta.items` vs. `Invoice.items`, but that one seems more deliberate.

It would also make `total_changed` read a little less sus, since we can now compare `self.total` against `invoice.total`

src/onegov/ticket/handler.py
```diff
@@ -89,18 +89,45 @@ def refresh(self) -> None:
...
+        return InvoiceMeta(
+            items=self.base_invoice_items(request),
+            rounding_base=rounding_base,
+            invoice=self.ticket.invoice,
+            rounding_text=self.rounding_text(request),
+        )
+
+    def rounding_text(self, request: CoreRequest) -> str:
+        """The localized text used for the rounding invoice item."""
+        return 'Rounding'
```
Daverball: If you want to avoid having this extra handler method, just for translation, you could do what I did with `fts_type_title` for models that sit in modules that don't have their own translations.

It's a little piggy, but as long as we only use ticket invoices with rounding in OrgApp or one of its derived applications, it should be fine, whenever that changes we can add translations to the ticket module.

E.g. for `DirectoryEntry`:
```python
from translationstring import TranslationString

...

    # HACK: We don't want to set up translations in this module for this single
    #       string, we know we already have a translation in a different domain
    #       so we just manually specify it for now.
    fts_type_title = TranslationString(
        'Directory entries',
        domain='onegov.org'
    )
```

src/onegov/town6/templates/macros.pt
```diff
@@ -3427,6 +3427,9 @@
                         <tal:b tal:case="'migration'" i18n:translate>
                             Lump sum
                         </tal:b>
+                        <tal:b tal:case="'rounding'" i18n:translate>
+                            Rounding
```
Daverball:
```suggestion
                            Rounding difference
```

Daverball (commented):
Looks good, I just have a few more follow-up suggestions.

cyrillkuettel:
I think I have everything now

src/onegov/pay/utils.py
Daverball: `total_vat` kind of has the same problem as `total` did, so you would need to add in the vat of the manual items here, in order to make it correct. Technically, it doesn't matter right now, since manual invoice items never have a vat rate set. But we might need to add that in the future.

Daverball (approved):
Looks good, there's one minor footgun left with `InvoiceMeta.total_vat` we probably should address, but it won't matter until we allow setting a VAT rate on manual invoice items.

PR #2575, opened 2026-07-13
Performance/n+1 query ical export

Tschuppi81:
Event: Fixes the N+1 query by dropping eagerly loading on the unused relationships in the iCal export

TYPE: Performance
LINK: https://seantis-gmbh.sentry.io/issues/7605741267


Daverball:
Technically a false positive, since it's more like a `N/500 + 1` query, it appears that SQLAlchemy loads the related objects in batches of 500 for the `selectin` strategy.

But since we actually don't need the occurrences here, it doesn't hurt to skip loading them. But I would probably switch to `noload` or `raiseload` from `lazyload` though, so we don't get a real N+1 query in the future, if we do end up accessing the occurrences for something. We'd rather know about this performance regression early, by virtue of our tests failing.

Daverball (approved):

PR #2590, opened 2026-07-20
Feature/fix edit menu and breadcrubs

Tschuppi81:
Org: Improve edit bar for several views

TYPE: Feature
LINK: None

Changed:
- directories: switch icon
- resources: Add menu to the far right
- newsletters: Settings to the left, change order to: Edit, Send, Test, Copy, Delete
- photoalbums: Edit far left
- events: Merge config filter and settings form
- news: Settings for root news, else keep Edit; adjust title

back / cancel using url param instead of browser history back in separate pr: https://github.com/OneGov/onegov-cloud/pull/2591

### DIFF
--- a/src/onegov/org/layout.py
+++ b/src/onegov/org/layout.py
@@ -1702,13 +1702,19 @@ class PersonCollectionLayout(DefaultLayout):
     @cached_property
     def editbar_links(self) -> list[Link | LinkGroup] | None:
         if self.request.is_manager:
-            return [
-                Link(
+            links: list[Link | LinkGroup] = []
+
+            if self.request.is_admin:
+                links.append(Link(
                     text=_('Settings'),
-                    url=self.request.link(
-                        self.request.app.org, 'people-settings'),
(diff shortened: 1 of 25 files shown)


src/onegov/org/layout.py
```diff
@@ -2281,34 +2281,38 @@ def editbar_links(self) -> list[Link | LinkGroup] | None:
                 Link(
                     text=_('Recipients'),
                     url=self.request.class_link(ResourceRecipientCollection),
-                    attrs={'class': 'manage-recipients'}
+                    attrs={'class': 'manage-recipients'},
+                ),
+                Link(
+                    text=_('Export All'),
+                    url=self.request.link(self.model, name='export-all'),
```
BreathingFlesh, line 2295: Needs a class to add an icon.

BreathingFlesh (commented):
It's a lot more unified now. But clicking on "settings" and then "save" doesn't return to the collection in newsletter and people, it should have a return_to parameter as in the event settings.

PR #2591, opened 2026-07-20
Feature/abort or back via url param instead of browser history back

Tschuppi81:
Org: Replace history back with return-to for several views

TYPE: Feature
LINK: None

Changed for:
- homepage
- newsletters
- people
- events
- notifications send (feriennet)
- agenda item (landsgemeinde)
- votum (landsgemeinde)


Tschuppi81:
I realized there is no `settings`menu for `/resources`, `/events` and `/ris-overview` like for others e.g. `/newsletters`. Shall we keep as-is?

Tschuppi81:
Shall I rename `/ris-overview` to `/ris`?

BreathingFlesh (approved):

PR #2592, opened 2026-07-21
Feature/ogc 236 phone number type

Tschuppi81:
Org: Extend phone number validator

TYPE: Feature
LINK: ogc-236

### DIFF
--- a/src/onegov/form/validators.py
+++ b/src/onegov/form/validators.py
@@ -11,6 +11,7 @@ from datetime import date
 from datetime import datetime
 from decimal import Decimal
 from dateutil.relativedelta import relativedelta
+from enum import Enum
 from io import BytesIO
 from mimetypes import types_map
 from onegov.core.utils import binary_to_dictionary, dictionary_to_binary
@@ -665,6 +666,12 @@ class StrictOptional(Optional):
             raise StopValidation()
 
 
+class PhoneNumberType(Enum):
+    ANY = 'any'
+    MOBILE = 'mobile'
+    FIXED_LINE = 'fixed_line'
+
+
 class ValidPhoneNumber:
     """ Makes sure the given input is valid phone number.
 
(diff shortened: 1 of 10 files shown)


src/onegov/form/validators.py
```diff
@@ -665,46 +666,121 @@ def __call__(self, form: BaseForm, field: Field) -> None:
             raise StopValidation()
 
 
+class PhoneNumberType(Enum):
+    ANY = 'any'
+    MOBILE = 'mobile'
+    FIXED_LINE = 'fixed_line'
```
Daverball: Why are we defining our own type, if we could just reuse `phonenumbers.PhoneNumberType`? (I know they currently aren't real enums, so you have to use `int` at the annotation everywhere you accept those values, I opened an issue about that years ago: https://github.com/daviddrysdale/python-phonenumbers/issues/229, so they're not as type-safe, but you also currently allow `str` instead of `PhoneNumberType`, so you're not gaining any type-safety anyways, you only get a runtime check, which you could still do if you used `phonenumbers.PhoneNumberType`)

`ANY` also seems unnecessary, why not just use `None` as the sentinel for any phone number type is valid?

In OCQMS I do the following:

```python
            if self.number_type in supported_types_for_region(region):
                # Validate type if we can
                number_type = number_type_for_number(numobj)
                if number_type == PhoneNumberType.FIXED_LINE_OR_MOBILE:
                    if self.number_type in (
                        PhoneNumberType.FIXED_LINE,
                        PhoneNumberType.MOBILE
                    ):
                        # make aggregate types match.
                        number_type = self.number_type
                if number_type != self.number_type:
                    raise ValidationError(_(
                        'Invalid ${number_type}',
                        mapping={'number_type': self.format_number_type()}
                    ))
```

src/onegov/form/validators.py
```diff
@@ -665,46 +666,121 @@ def __call__(self, form: BaseForm, field: Field) -> None:
...
-            raise ValidationError(self.message)
+        if self.number_type == PhoneNumberType.ANY.value:
+            return
+
+        # an unknown number type matches nothing, so we reject the number
+        # instead of blowing up in the middle of a request
+        valid_types, error = self.type_errors.get(
+            self.number_type, ((), self.invalid_phone_number)
+        )
+        if phonenumbers.number_type(number) not in valid_types:
```
Daverball: Depending on the country not all number types are possible, there's a `supported_types_for_region` function which tells you which types can be distinguished, if you can't distinguish the types you're asking for, then you shouldn't raise validation errors for them.

src/onegov/form/validators.py
```diff
@@ -665,46 +666,121 @@ def __call__(self, form: BaseForm, field: Field) -> None:
...
+            fixed_line_required
+        ),
+        PhoneNumberType.MOBILE.value: (
+            (
+                phonenumbers.PhoneNumberType.MOBILE,
+                phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE
+            ),
+            mobile_required
+        ),
+    }
```
Daverball: This seems overly complex, I would just special-case `FIXED_LINE_OR_MOBILE` in the validation function where we check if the types match. Like in the example code from OCQMS.

src/onegov/form/validators.py
```diff
@@ -670,41 +670,114 @@ class ValidPhoneNumber:
...
-            raise ValidationError(self.message) from exception
+            raise ValidationError(self.invalid_phone_number) from exception
+
+        reason = phonenumbers.is_possible_number_with_reason(number)
+        if reason != phonenumbers.ValidationResult.IS_POSSIBLE:
+            raise ValidationError(
+                self.length_errors.get(reason, self.invalid_phone_number)
+            )
+
+        if not phonenumbers.is_valid_number(number):
```
Daverball, line 745: Are you sure the ordering of `is_valid_number` and `is_possible_number_with_reason` is correct? It's difficult to determine from reading the libphonenumbers code, but to me it seems like `is_possible_number` is the more strict check of the two, so if the number is possible, it should never be invalid.

We used the opposite order for these two checks before this change.

Tschuppi81: Let me check back on this

Tschuppi81: the source doc (phonenumbers/phonenumberutil.py) of `is_possible_number_with_reason` says: `This provides a more lenient check than is_valid_number`. `It only checks the length of phone numbers`.

This ordering also enables a more detailed error messaging as a too short/long phone number would end up as `invalid` with the reversed ordering.

Daverball (approved):

PR #2593, opened 2026-07-21
Org: Add possibility of adding year in holidays.

cyrillkuettel:

## Commit message

Org: Add possibility of adding year in holidays.

<Optional Description>

TYPE: Feature
LINK: OGC-2385

## Checklist


src/onegov/org/forms/settings.py
```diff
@@ -1023,7 +1023,7 @@ class HolidaySettingsForm(Form):
 
     other_holidays = TextAreaField(
         label=_('Other holidays'),
-        description=('31.10 - Halloween'),
+        description=('31.10 - Halloween / 31.10.2025 - Halloween'),
```
Daverball: I would keep the old description intact, since this it was displayed as a placeholder and you can't have two entries on the same line.

If you want a more explanatory/verbose description, you're better off doing something like `organisation_hierarchy` where you give an example with a markdown code block. Long descriptions are rendered as markdown.

src/onegov/org/models/swiss_holidays.py
```diff
@@ -164,6 +182,10 @@ def other(self, year: int) -> Iterator[tuple[date, set[str]]]:
         for month, day in self._other:
             yield date(year, month, day), self._other[(month, day)]
 
+        for dt in self._other_dated:
+            if dt.year == year:
+                yield dt, self._other_dated[dt]
+
```
Daverball:
```suggestion
        for dt, descriptions in self._other_dated.items():
            if dt.year == year:
                yield dt, descriptions

```

This is also true for the loop above, but we probably should just iterate through the items, instead of the keys

Daverball: Also this is kind of an edge-case, but if you define two holidays, one with a year and one without, that cover the same day of the year, you will now emit the same date twice, instead of one date that combines the description of the two. This is especially relevant if both variants have the same description.

In `all` we properly de-duplicate the dates, so this method should probably behave the same. It might then be also worth simplifying `all` by reusing `other` instead of duplicating those two loops, even if it is less efficient.

But we could also consider different ways to encode the same information, like changing `_other` to `dict[tuple[int, int], dict[int | None, set[str]]` (i.e. `defaultdict(lambda: defaultdict(set))`), you would then iterate through `_other` like before, but instead of directly getting descriptions you construct them for a given year by combining `descriptions.get(None, set()) | descriptions.get(year, set())`.

That way you retain the old single loop structure, you have to do a little bit of extra work to get the descriptions for a given year, but you have to do less work to de-duplicate the descriptions.

cyrillkuettel: Okay thanks I'll see what I can do.

Daverball (approved):

PR #2597, opened 2026-07-23
Org: Warn if overlaps occur in availability periods in add / edit.

cyrillkuettel:

## Commit message

Org: Warn if overlaps occur in availability periods in add / edit.

<Optional Description>

TYPE: Feature
LINK: OGC-2602

## Checklist


src/onegov/org/views/allocation.py
```diff
@@ -511,6 +532,9 @@ def handle_edit_rule(
                 },
             )
         )
+
+        warn_about_skipped_allocations(request, form, new_allocations_count)
```
Daverball, line 536: I think this should be `new_allocations_count + updated_count`

cyrillkuettel: Thanks for catching that. 

It seems also that `warn_about_skipped_allocations` would need to be remembered to be called by other views if they modify rules somehow.

PR #2600, opened 2026-07-27
Feature/ogc 3098 permits notifications with signed pdf

Tschuppi81:
Directory: Permit email notifications with signed PDF attached

TYPE: Feature
LINK: ogc-3098


Tschuppi81:
Two questions:
- if the signing fails, send the email anyway (implemented without signed pdf) and remember to sign later on?
- Should i implement a specific pdf class such/similar to `TicketBasePdf`?

Daverball:
I'm not sure what the best strategy for failed signatures is. It definitely makes sense to always send the e-mail. I'm not sure the extra complexity of retrying the signature at a later date is worth it though. It might be fine to just put a comment into the email that the attached PDF could not be signed, because the signature service was not available. It should happen so rarely, that we can deal with those cases manually, if they happen to matter, rather than introducing a bunch of potentially fragile and untested code for something that almost never happens.

A specific PDF class makes sense to me, it makes it easier to unit test the different sections of the PDF and potentially reuse for something else later.

src/onegov/org/views/directory.py
```diff
@@ -706,10 +865,27 @@ def send_admin_notification_for_directory_entry(
             'publication_start': entry.publication_start,
             'publication_end': entry.publication_end,
             'content_hash': entry.content_hash,
+            'generated_at': generated_at,
         },
+        attachments=(pdf,) if pdf else (),
     )
 
 
+def sign_pdf(pdf: Attachment, request: OrgRequest) -> None:
```
Daverball: I would refactor this to take a buffer and return a buffer and do the signing step before you create the attachment object. That way you can reuse the buffer you used to generate the PDF (You might need to call `.seek(0)` before you can pass it to `sign` though, but it's possible that AIS2.py already does that for us, I can't remember).

Daverball: Or, honestly just inline it into the other function, it's just a method call and a `try`/`except`, this doesn't need its own helper.

src/onegov/org/views/directory.py
```diff
@@ -705,7 +754,30 @@ def send_admin_expiry_notification_for_directory_entry(
...
+    if pdf is not None:
+        try:
+            signed_pdf = BytesIO()
+            request.app.signing_service.sign(
+                request.session, BytesIO(pdf.content), signed_pdf, None)
+            pdf.content = signed_pdf.getvalue()
+        except Exception:
+            # an unsigned pdf is better than no notification at all
+            log.exception('Error while signing directory publication')
+            signing_failed = True
```
Daverball: You're still creating a buffer, only to then convert it to bytes and then back to a buffer. I meant refactor the signing step into the function that creates the attachment, you sign the buffer before you create the attachment, so you can reuse the buffer you already have.

At this point the helper function isn't doing anything anymore, it's just doing a single function call, so you're better off putting the two `try/except` blocks directly in there and return `Attachment | None` instead of `Attachment`.

Daverball: Or since you need to know whether signing failed, you can return `tuple[Attachment | None, bool]` or create a `NamedTuple` with attributes for `pdf` and `signing_failed`, so it's little less opaque.

src/onegov/org/pdf/directory_entry.py
```diff
@@ -0,0 +1,213 @@
...
+
+        self.h2(_('Publication'))
+
+        def value(field: BasicParsedField) -> str:
+            data = entry.values.get(field.id)
+            if data and field.type == 'datetime':
+                data = layout.format_date(data, 'datetime_long')
+            elif data and field.type == 'date':
+                data = layout.format_date(data, 'date_long')
+            return escape(str(data or ''))
```
Daverball: This is not super robust. What about other fields that don't necessarily have obvious string representations like time fields, or checkboxes, or questionable fields like password fields?

This also doesn't handle subfields at all. Numeric fields won't render `0` correctly.

Generally we rely on rendered form display here. See e.g. `FormSubmissionHandler.get_summary`. That's a lot more reliable and avoids us having to change our rendering in a dozen different places, every time we add a new field to formcode.

This does mean you will need to use the actual form object, rather than the parsed fields, so you will need to filter out the upload fields manually, if you don't want them to be displayed in this initial list.

If you want a little bit more control over how the labels are rendered you will need to manually write the `form.render_display` loop, instead of reusing the template macro.

src/onegov/org/templates/mail_macros.pt
```diff
@@ -170,9 +170,17 @@
...
+    <p>
+        <tal:b tal:repeat="field directory.basic_fields">
+            <tal:b tal:define="value entry.values.get(field.id, '')">
+                - <strong>${field.human_id}:</strong>
+                <tal:b tal:condition="python: value and field.type == 'datetime'">${layout.format_date(value, 'datetime_long')}</tal:b>
+                <tal:b tal:condition="python: value and field.type == 'date'">${layout.format_date(value, 'date_long')}</tal:b>
+                <tal:b tal:condition="python: not (value and field.type in ('datetime', 'date'))">${value}</tal:b>
+                <br tal:condition="not: repeat.field.end"/>
+            </tal:b>
+        </tal:b>
```
Daverball, line 184: Same thing here. Form display should use the standard `Form.render_display`. Or reuse the existing form display macros for e-mail. (The same ones we use in ticket e-mails)

src/onegov/org/views/directory.py
```diff
@@ -661,9 +669,55 @@ def _send_admin_email(
...
+            pdf_bytes = signed_pdf.getvalue()
+        except Exception:
+            # an unsigned pdf is better than no notification at all
+            log.exception('Error while signing directory publication')
+            signing_failed = True
+            pdf_bytes = content.getvalue()
+
+    return Attachment(
+        # slashes would be cut off by Attachment's basename()
+        f"{entry.title.replace('/', '-')}.pdf",
```
Daverball: Depending on the filesystem the file will be stored on, allowing arbitrary characters can be a problem. The path separator is not the only reserved character.

It might be better to use `entry.name` here, which is guaranteed to be url safe. Albeit a little less human readable, but that seems fine to me.

Daverball (changes requested):
Looks good overall, but the rendering of form fields is too brittle currently, we should reuse our existing form display rendering utilities.

src/onegov/org/pdf/directory_entry.py
```diff
@@ -0,0 +1,218 @@
...
+            if field.type == 'DateField':
+                rendered = escape(layout.format_date(field.data, 'date_long'))
+            elif field.type in ('DateTimeLocalField', 'TimezoneDateTimeField'):
+                rendered = escape(
+                    layout.format_date(field.data, 'datetime_long'))
+            items.append(
+                f'<li><strong>{escape(field.label.text)}</strong>: '
+                f'{rendered}</li>'
+            )
+        self.mini_html(f'<ul>{"".join(items)}</ul>')
```
Daverball:
```suggestion
        self.mini_html(Markup('<ul>{}</ul>'.format(Markup('').join(
            Markup('<li><strong>{}</strong>: {}').format(
                field.label.text,
                form.render_display(field)
            )
            for field in form
            if not isinstance(field, (UploadField, UploadMultipleField))
        ))
```
I don't think we should special-case any fields other than upload fields. Also relying on `markupsafe.Markup` here is a little bit more robust, than manually using `html.escape`.

Daverball (approved):

PR #2614, opened 2026-08-06
Core: Add audit trail.

cyrillkuettel:

## Commit message

Core: Add audit trail.

<Optional Description>

TYPE: Feature
LINK: OGC-180

## Checklist


src/onegov/core/orm/audit.py
```diff
@@ -0,0 +1,252 @@
...
+    username: Mapped[str]
+
+    #: The time at which the operation was captured
+    created: Mapped[datetime] = mapped_column(default=utcnow)
+
+    __table_args__ = (
+        Index('audit_entries_target', 'target_table', 'target_id'),
+        Index('audit_entries_created', 'created'),
+        CheckConstraint(
+            "operation IN ('insert', 'update', 'delete')",
```
Daverball: Why not use an enum? It's far more efficient on the database end and either way you will need to update the table if you want to add additional operations. So whether you will need to update a type or a constraint, really doesn't make that much of a difference.

src/onegov/core/orm/audit.py
```diff
@@ -0,0 +1,252 @@
...
+                else entry.config.snapshot(entry.instance)
+            ),
+            'previous_snapshot': entry.previous_snapshot,
+            'username': entry.username,
+            'created': entry.created,
+        }
+        for entry in staged
+    ]
+
+    session.connection().execute(insert(AuditEntry), values)
```
Daverball: Rather than writing your own raw SQLAlchemy event handlers, it might be better to rely on the same session manager events/signals the search relies on, since those signals also properly deal with bulk updates/deletes/inserts.

That way you also don't have to split this into two separate steps.

src/onegov/core/request.py
```diff
@@ -175,7 +176,11 @@ def identity_secret(self) -> str:
 
     @cached_property
     def session(self) -> Session:
-        return self.app.session()
+        session = self.app.session()
+        session.info.pop(AUDIT_USERNAME, None)
+        if is_logged_in(self.identity):
+            session.info[AUDIT_USERNAME] = self.identity.userid
+        return session
```
Daverball: It might be better to move this into a method on `SessionManager`, similar to `set_locale`, that gets invoked by a tween. This implementation seems a little dangerous, since the username in the session could be changed by a throwaway copy of the current request.

It's also worth thinking about whether we should store `user.id` in addition or instead of `user.username`, since `user.username` can change.

src/onegov/core/upgrades.py
```diff
@@ -262,3 +260,28 @@ def drop_remaining_gazette_and_notices_tables(context: UpgradeContext) -> None:
...
+            JSON,
+            nullable=False,
+            server_default=text(r"'{}'::jsonb"),
+        ),
+    )
+    context.operations.alter_column(
+        'audit_entries',
+        'previous_snapshot',
+        server_default=None,
+    )
```
Daverball: I assume you needed this during development, since you changed your mind about whether or not this column should be nullable. We should no longer need this migration.

Daverball (commented):
The overall direction looks decent to me, but we should probably rely more on the existing paradigms for attaching additional state to the current session and responding to changes, rather than rebuilding those tools in a slightly different way with slightly different semantics.

src/onegov/core/orm/session_manager.py
```diff
@@ -712,6 +715,37 @@ def set_locale(
...
+
+    @property
+    def current_user_id(self) -> str | None:
+        return getattr(self._current_user, 'user_id', None)
+
+    @property
+    def current_username(self) -> str | None:
+        return getattr(self._current_user, 'username', None)
+
+    def _bind_current_user[T: Session](self, session: T) -> T:
```
Daverball: Rather than binding the session, it might be better to follow `TranslationHybrid`'s example and rely on `SessionManager.get_active()` for retrieving the current username and user id.

src/onegov/core/framework.py
```diff
@@ -1838,6 +1840,31 @@ def current_language_tween(request: CoreRequest) -> Response:
...
+            username = identity.userid
+        else:
+            user_id = None
+            username = None
+
+        if app.has_database_connection:
+            app.session_manager.set_current_user(
+                user_id=user_id,
+                username=username,
+            )
```
Daverball: Since an incorrect user binding is a little bit more harmful than an incorrect language binding we might want to rely on a `contextmanager` which ensures the user gets reverted to its original state after the transaction has been committed.

I.e. do something like:
```python
with app.session_manager.set_current_user(...):
    return handler(request)
```

I think the tween should already be installed at the correct location in the chain of tweens.

src/onegov/core/orm/session_manager.py
```diff
@@ -304,6 +306,7 @@ def on_delete(schema, session, obj):
         self.bases = [base]
         self.created_schemas: set[str] = set()
         self.current_schema: str | None = None
+        self._current_user = threading.local()
```
Daverball: I like the increased thread safety (although currently it shouldn't matter, since the separate cronjob thread does not run the jobs, it only submits HTTP requests to itself, so it doesn't mess with the session manager).

Although if we start worrying about that, we should make everything thread-safe and not just this one setting.

src/onegov/core/orm/cache.py
```diff
@@ -435,7 +435,9 @@ def wrapper(self: AppT) -> T:
             session.flush()
 
         if cache_key in self.request_cache:
-            return maybe_merge(self.session(), self.request_cache[cache_key])
+            value = maybe_merge(session, self.request_cache[cache_key])
+            self.request_cache[cache_key] = value
+            return value
```
cyrillkuettel, line 440: This fixes `tests/onegov/org/test_views_settings.py::test_settings_tags` failing.

Which was a result of 0eece84540047242c4b66f547f0e589530608a4b where the transaction tweens were changed in order.

cyrillkuettel: It’s an edge case that only happens when a cached ORM object becomes detached, gets merged into a new session, and is then modified while the cache still holds the old object.

Daverball: good catch

PR #2618, opened 2026-08-10
Feriennet: Keep unpaid discounts/surcharges grouped with their paid bill

Tschuppi81:
Feriennet: Keep unpaid discounts/surcharges grouped with their paid bill

TYPE: Feature
LINK: no ticket yet, just stumbled over this


Tschuppi81:
Not changing this in general but `Total` and `Open` amounts to state as on the invoice

PR #2619, opened 2026-08-10
Bugfix/3366 directories save entry with publication in the past leads to loss of attachmnts

Tschuppi81:
Directory: Preserve attachments on entry save with past publication date 

Uploading a file and saving/editing a directory entry with a publication date in the past triggered a validation error that discarded the attachment. The upload now survives the re-render (new entries) and edits keep the persisted file. 
                                                          
TYPE: Bugfix                                                                                    
LINK: OGC-3366


Tschuppi81:
close as rebased, https://github.com/OneGov/onegov-cloud/pull/2633

PR #2627, opened 2026-08-13
Bugfix/ogc 3388 reservations deleted slots still visible

Tschuppi81:
Reservation: Ensure ticket deletion does not leave slots behind, cleanup orphaned `ReservedSlot`s

TYPE: Bugfix
LINK: ogc-3388


src/onegov/org/models/ticket.py
```diff
@@ -830,7 +830,22 @@ def reply_to(self) -> str | None:
...
+        # slots too, otherwise deleting the reservations directly (instead of
+        # via the scheduler) leaves orphaned slots blocking the calendar
+        token = reservations[0].token
+        self.session.query(ReservedSlot).filter(
+            ReservedSlot.reservation_token == token,
+            ReservedSlot.source_type == 'reservation',
+        ).delete(synchronize_session=False)
+
+        for reservation in reservations:
             self.session.delete(reservation)
```
Daverball, line 854: I'm not super thrilled about re-implementing the libres logic here.

Can you verify whether or not we always access `handler.resource` when we call `prepare_delete_ticket`? If we always access that anyway, it might be better to just do `self.resource.remove_reservation(self.id)` here, since we're not saving a query.

src/onegov/reservation/upgrade.py
```diff
@@ -443,65 +444,58 @@ def resources_switch_to_parsed_form(context: UpgradeContext) -> None:
...
-    """))
+    if context.has_column('reserved_slots', 'source_id'):
+        return
+
+    context.operations.add_column(
+        'reserved_slots', Column('source_id', Integer, nullable=True)
+    )
+
+    backfill_reserved_slot_source_ids(context.session)
+
```
Daverball, line 498: Don't forget to update the column to `nullable=False` afterwards. `ReservedSlots` without a source object should not exist.

src/onegov/org/models/ticket.py
Tschuppi81: We do not always access `handler.resource` in the delete flow. `delete_tickets_and_related_date` (cronjob) as well as `ticket_deletable` don't access it.

Daverball: I see, that's a little unfortunate. Although to be fair, we're already very inefficient, since we have to fetch the reservations for each ticket, so fetching the resource as well, probably wouldn't make things much worse. So it might still be worth to reuse the `libres` function, instead of manually needing to keep this in sync.

If we ever wanted to speed this up we would need to batch tickets by handler code and do a prefetch of all the related objects when deleting tickets, per handler. So we don't end up with a N+1 query. So whether the prefetch would include just the reservations or both the reservations and resources, would not really significantly change things at that point.

I'll leave it up to you whether to change it now, or just add a `FIXME` to replace this with `scheduler.remove_reservation` in the future.

Daverball:
You can pin libres to `>= 1.2.0` now, to make sure the libres schema matches the upgrade task

src/onegov/org/models/ticket.py
Tschuppi81: Added a fixme

Daverball (approved):

PR #2633, opened 2026-08-14
Directory: Preserve attachments on entry save with past publication date

Tschuppi81:
Directory: Preserve attachments on entry save with past publication date for single and multi upload fields

Uploading a file and saving/editing a directory entry with a publication date in the past triggered a validation error that discarded the attachments. The upload now survives the re-render (new entries) and edits keep the persisted files.

TYPE: Bugfix
LINK: OGC-3366


src/onegov/form/fields.py
```diff
@@ -379,12 +379,24 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
...
+                # (size, mimetype) so display and validation still work
+                original = self.object_data
+                if isinstance(original, dict) and \
+                        original.get('data') == raw_data:
+                    self.data = cast('StrictFileDict', original)
+                else:
+                    self.data = {
+                        'data': raw_data,
+                        'filename': str(valuelist[2]),
+                    }
```
Daverball: I'm not super happy that we have moved an implementation detail of how directory entries store their uploaded files into the generic `UploadField`. If you can make this work without adding any special-casing here, that would be a preferred solution.

src/onegov/directory/models/directory.py
```diff
@@ -335,7 +341,36 @@ def get_value_field_from_note(file_id: str) -> Any:
...
+                                filename=data['filename']
+                            )
+                        )
+                        entry.files.append(new_file)
+                        updated[field.id] = {
+                            'data': '@' + new_file.id,
+                            'filename': data['filename'],
+                            'mimetype': new_file.reference.file.content_type,
+                            'size': new_file.reference.file.content_length
+                        }
```
Daverball: It's a little unfortunate how things here got even more messy and repetitive, I think we're at the point where a refactor might be necessary to keep things readable/understandable.

Daverball (commented):
Overall it seems like a reasonable solution, but it also makes an already messy situation even messier. I'm also not happy that some of the implementation details of directories have now leaked into the generic form module.

If we can clean this up and improve the situation, I'll be a lot happier.

src/onegov/form/fields.py
Tschuppi81: Right, in the generic `UploadField` makes only sense if we face the same issue for other form too.

Tschuppi81: It turns out plain form submissions rely on this the same way directories do: saved uploads are referenced as @<id> and resent on edit — without the @ branch they get decoded as inline data and lost. Added `test_pending_submission_file_survives_edit` to cover it. So `UploadField` is the right location and fixes Directories and Form Submissions.

Daverball: If that's the case it makes even more sense to define formcode specific subclasses, since we still use `UploadField` in other places that aren't form submissions. And we probably want to refactor things so that file handling is reusable between `FormSubmissionCollection.update` and `Directory.update`, otherwise we will always have to fix bugs in two locations and be very careful to keep things in sync, which they may already not be.

src/onegov/form/fields.py
```diff
@@ -626,6 +635,67 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
                 self.append_entry_from_field_storage(value)
 
 
+class FormcodeUploadField(UploadField):
```
Tschuppi81: Newly added classes `FormcodeUploadField` and `FormcodeUploadMultipleField`

src/onegov/form/fields.py
```diff
@@ -379,11 +379,8 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
             # resend_upload
             action = valuelist[0]
             fieldstorage = valuelist[1]
-            # NOTE: I'm not sure why mypy complains here, a total version
-            #       of a TypedDict should be assignable to a non-total version
-            self.data = binary_to_dictionary(  # type: ignore[assignment]
-                dictionary_to_binary({'data': str(valuelist[3])}),
-                str(valuelist[2])
+            self.data = self.process_resend(
```
Daverball: Looking at this more closely, I think the entire bug here was that we ignore the value of `action` when we resend the upload. And we probably still have a subtle bug here. When you get a validation error and then switch from `keep` to `replace` or `delete` then you still keep the old file.

We should only process the resend if the action is `keep` and `if not self.object_data`, in all other cases `self.data` should already be correct.

We might not need any of the other changes.

Daverball: Really we probably want to improve resending for editing in general, since we currently can't distinguish between resending a `replace` (or an entirely new file) and resending a `keep`, a `keep` of an existing file should realistically never trigger a resend, since the user never sent a file in the first place.

Tschuppi81: Let me see, how this works out

Tschuppi81: the ideal case  `keep` never triggers a resend, is not achievable here because on a directory POST object_data is empty, so the resent @<id> is the only thing carrying the file across the re-render. I think we would need a `persistency on upload` than we could reference it by id. Not sure if this refactoring is worth doing

Tschuppi81: For now I kept `FormcodeUploadField` and `FormcodeUploadMultipleField` to keep it separate for Submission and Directory usage. It is a little simpler but not as expected or ideally..

Daverball: The simplification should be everywhere else, the goal here is to get rid of the workaround you added in `Directory.update`.

But if we get this right we also shouldn't need these new fields, since resending only happens if we're going to store a new file, it should never happen for a file we already stored.

src/onegov/form/fields.py
```diff
@@ -401,9 +400,31 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
...
+        filename: str,
+        raw_data: str
+    ) -> StrictFileDict | FileDict:
+        """ Restores a file resent across a re-render. """
+        if raw_data.startswith('@'):
+            # already stored: keep the reference, reuse loaded metadata
+            original = self.object_data
+            if isinstance(original, dict) and original.get('data') == raw_data:
+                return cast('StrictFileDict', original)
+            return {'data': raw_data, 'filename': filename}
```
Daverball: This code should go away if we do things correctly, we only need to resend data `if field.raw_data`, that way we never resend the stored manipulated format with the `@` references.

Currently `UploadWidget.template_data` incorrectly only checks `field.data`, but it needs to check both `field.data` and `field.raw_data`, there's no point in resending something we already stored.

src/onegov/directory/models/directory.py
```diff
@@ -380,23 +383,22 @@ def get_value_field_from_note(file_id: str) -> Any:
...
-                        for file in entry.files:
-                            if file.id == file_id:
-                                new_key = f'{field.id}:{new_idx}'
-                                if file.note != new_key:
-                                    file.note = new_key
-                                break
-                        new_idx += 1
-                        continue
+                        # new entry / added file: no stored file yet, fall
+                        # through to create it from the resent upload below
```
Daverball: I'm still pretty unhappy about how different this is from `FormSubmissionCollection.update` and that there's no code-reuse between the two at all. But I suppose we don't have to fix that right now.

src/onegov/directory/models/directory.py
```diff
@@ -606,6 +594,39 @@ def process_obj(self, obj: DirectoryEntry) -> None:
...
+                # POST rebuilds from formdata only; restore kept stored files
+                # before validation (new uploads carry their own resent data)
+                if not self.request.POST:
+                    return
+
+                values = getattr(self.model, 'values', None) or {}
+                for field in directory.file_fields:
+                    form_field = getattr(self, field.id, None)
+                    if form_field is None:
+                        continue
```
Daverball:
```suggestion
                    if field.id not in self:
                        continue
                    form_field = self[field.id]
```
Always use the item access API for retrieving form fields, since `getattr`/`hasattr` can include other non-field attributes

src/onegov/directory/models/directory.py
```diff
@@ -606,6 +594,39 @@ def process_obj(self, obj: DirectoryEntry) -> None:
                     else:
                         form_field.data = data
 
+            def on_request(self) -> None:
```
Daverball: All of this seems completely unnecessary. If `UploadField` works correctly, then with an action of `keep` it should leave its own `data` untouched, so it automatically should match the `field.object_data` we originally passed in via `values` when constructing the form object.

If the `UploadField` messes with it's `data` despite the action being `keep`, then that's a bug and you need to fix it there instead of trying to work around it here.

src/onegov/form/widgets.py
```diff
@@ -205,7 +206,14 @@ def template_data(
...
         previous = ''
-        if field.data and resend_upload:
+        # only resend a fresh, not-yet-stored upload (field.raw_data); never
+        # resend an already-stored '@<id>' reference
+        if (
+            field.data
+            and field.raw_data
+            and not is_stored_file_reference(field.data)
+            and resend_upload
+        ):
```
Daverball, line 218:
```suggestion
        if (
            resend_upload
            and field.data
            and field.raw_data
            and not is_stored_file_reference(field.data)
        ):
```
Let's reorder this so the least expensive check happens first

src/onegov/form/fields.py
```diff
@@ -393,17 +392,41 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
...
             fieldstorage = valuelist[0]
 
+        # a resend only ever carries a fresh upload (stored files are re-bound,
+        # never resent), so a resend with no new file picked is that upload
+        # coming back across a re-render: store it (a replace)
         if action == 'replace':
             self.action = 'replace'
             self.data = self.process_fieldstorage(fieldstorage)
+            if not self.file and resend is not None:
+                self.data = self.process_resend(resend_filename, resend)
```
Daverball: This is incorrect, if someone tries to upload a file and a validation error happens it shows up as if the upload succeeded, so the re-sent file only applies to `keep`, if someone then explicitly switches to the `replace` option and attaches a new file, that file takes priority over the resent file.

src/onegov/directory/models/directory.py
Tschuppi81: Replaced with a second `process` as long is we are not aligned to the `FormSubmission` way doing it...

src/onegov/directory/models/directory.py
Daverball: This part of the code still makes me a little uneasy, we may want to put in a guard that the subfield actually has a `file` and `filename` object we can create the file from. Otherwise you can make some pretty strange things happen if you submit outdated formdata against a directory entry that has already been modified in the meantime.

And while you increased the code-reuse the overall logic is still very different. Can we try to unify the `files_to_keep`/`files_to_add` approach? The overall logic is the most likely place to contain bugs, so ideally we use the same logic in both places, so we only have to fix bugs once and can't forget to port new features/improvements from one place to the other.

This part of the code is a little hairy because it is also used for directory migrations, so it might be difficult to unify, but we probably should still try to make it work consistently. If migrations become a sticking point we should probably change how we pass in data for executing a migration, rather than try to adapt the internal logic to support both use-cases.

tests/shared/utils.py
```diff
@@ -139,13 +139,35 @@ def create_image(
...
+        buffer = BytesIO()
+        c = canvas.Canvas(buffer)
+        c.drawString(100, 750, content)
+        c.save()
+        buffer.seek(0)
+        return buffer
+
     c = canvas.Canvas(filename)
     c.drawString(100, 750, content)
     c.save()
```
Daverball:
```suggestion
    c = canvas.Canvas((buffer := BytesIO()) if filename is None else filename)
    c.drawString(100, 750, content)
    c.save()

    if filename is None:
        buffer.seek(0)
        return buffer
```
While splitting across two if statements feels a bit awkward in its own way, it seems better than duplicating the generation of the PDF's contents, even if that part is really simple now, it might change at some point.

Daverball (commented):
Overall this is looking pretty good. I'm a little uncomfortable with the large scope of the change, but it does seem like an overall improvement.

Daverball:
Unifying the logic between single and multiple file fields is nice too, but I'm more concerned about unifying the logic between directory entries and form submissions. I would leave unifying single and multiple file fields as a future exercise to keep the changes small, the code duplication there is actually not as bad and in some ways better for performance reasons, the main goal is to make directory entries use the same logic for determining which files need to be deleted/inserted/updated as form submissions and without duplicating that code.

So rather than keep working on code that we would like to delete, I'd prefer working on making the code for form submissions reusable with directory entries and that means make it reusable as one big procedure, rather than a set of common helper functions re-arranged in two completely different ways.

src/onegov/form/utils.py
```diff
@@ -188,3 +191,110 @@ def remove_empty_links(
...
+
+def reconcile_uploaded_files[FileT: File](
+    *,
+    file_cls: type[FileT],
+    fields: Mapping[str, Any],
+    multiple: Collection[str],
+    files: MutableSequence[FileT],
+    delete: Callable[[FileT], None],
+    flush: Callable[[], None] = lambda: None,  # noop
+) -> dict[str, Any]:
```
Daverball, line 218: Can we keep this much closer to what the original function for form submissions looked like please?

This is difficult to review and I don't trust our regression tests enough to catch all of the edge cases, so the less we change, the better. We can always try to reduce repetition between multiple and single file upload fields later.

PR #2634, opened 2026-08-14
No longer remove new emojis from messages

Tschuppi81:
Core: Extend emoji ranges in `html_to_text` plaintext filter

TYPE: Feature
LINK: pro-1411

### DIFF
--- a/src/onegov/core/html.py
+++ b/src/onegov/core/html.py
@@ -59,19 +59,9 @@ SANE_HTML_ATTRS = {
     'ol': ['class']
 }
 
-# lines without these plaintext characters are excluded in html_to_text
-VALID_PLAINTEXT_CHARACTERS = re.compile(r"""
-    [
-        \d  # decimals
-        \w  # words
-        \n  # new lines
-
-        # emojis
-        \U00002600-\U000027BF
-        \U0001f300-\U0001f64F
-        \U0001f680-\U0001f6FF
-    ]+
-""", re.VERBOSE)
+# lines consisting solely of these characters are excluded in html_to_text;
+# any other character (words, digits, emoji of any kind) keeps the line
+DECORATION_ONLY_CHARACTERS = re.compile(r'[^\s\-*_|=#>.]')
 
 # match empty link expressions
 EMPTY_LINK = re.compile(r'\[\]\([^)]+\)')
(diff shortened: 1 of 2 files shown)


Tschuppi81:
Regex reverted to what shall not survive the `html_to_text` conversion

src/onegov/core/html.py
```diff
@@ -59,19 +59,9 @@
...
-
-        # emojis
-        \U00002600-\U000027BF
-        \U0001f300-\U0001f64F
-        \U0001f680-\U0001f6FF
-    ]+
-""", re.VERBOSE)
+# lines consisting solely of these characters are excluded in html_to_text;
+# any other character (words, digits, emoji of any kind) keeps the line
+DECORATION_ONLY_CHARACTERS = re.compile(r'[^\s\-*_|=#>.]')
```
Daverball: I realize that this is less maintenance work for us, but it also no longer produces quite the same results, since our exclusion set is much more narrow now. So we don't exclude things like em/en dashes or other punctuation.

Any reason we're not just adding the three or so additional Unicode ranges containing emojis that have been added since we first wrote this regex?

Tschuppi81: The number of ranges and the verbose mode on the decoration-only test case...

src/onegov/core/html.py
```diff
@@ -60,18 +60,36 @@
...
+    r'\U0001F700-\U0001F77F'
+    r'\U0001F780-\U0001F7FF'
+    r'\U0001F800-\U0001F8FF'
+    r'\U0001F900-\U0001F9FF'
+    r'\U0001FA00-\U0001FA6F'
+    r'\U0001FA70-\U0001FAFF'
+    r'\U0001F3FB-\U0001F3FF'
+    r'\U0000200D'
+    r'\U0000FE0F'
+    r']+'
```
Daverball, line 78: This list seems to include a lot of things other than emojis like e.g. mathematical symbols.

I personally wouldn't worry about those, I would focus on the few emoji specific ranges, rather than including a lot of the other symbols, I think there should only be around six or so relevant ranges, based on a quick search. We also don't need to worry about variant modifiers, since those should always be accompanied by another emoji code point, so we don't need to detect them.

Daverball (approved):

PR #2650, opened 2026-08-24
Bugfix/ogc 3406 surcharges or discounts zeros invoice

Tschuppi81:
Reservation: Fix pricing snapshot migration that zeroed invoices

The `Store pricing settings on reservations` migration zeroed the stored price of many reservations, which made `refresh_invoice_items` (triggered e.g. by adding a surcharge/discount) wipe the invoice line and drop the payment (OGC-3406).                                                                       
                                                                                            
Re-run the corrected migration under a new name so it executes again on already-upgraded databases, fixing three bugs that caused the zeroing:                    
                                                                                            
  - the allocation lookup matched the wrong `pricing_method` constants (`price_per_item`/`price_per_hour` instead of `per_item`/`per_hour`),                   
  - the `resource = mirror_of` guard was mis-parenthesised (AND bound tighter than the following ORs), and                                                    
  - the resource-content fallback read `content->'price_per_item'`, but the resource stores that value under `price_per_reservation`.                               
                                                                                            
Add an org upgrade task to recompute invoice lines and payments that a post-rollout refresh had already zeroed. Scoped to reservation invoices modified on or after the backfill rollout (2026-08-18) that are tied to a paying reservation — a per_item/per_hour allocation, or a per_item resource and refreshed only when safe (manual, still-open payment).

TYPE: Bugfix
LINK: ogc-3406

### DIFF
--- a/src/onegov/org/upgrade.py
+++ b/src/onegov/org/upgrade.py
@@ -1000,7 +1000,11 @@ def switch_to_parsed_event_filters(context: UpgradeContext) -> None:
     org.event_filter_parsed_definition = ParsedForm.from_formcode(definition)
 
 
-@upgrade_task('Refresh reservation invoices zeroed by the pricing backfill')
+@upgrade_task(
+    'Refresh reservation invoices zeroed by the pricing backfill',
+    requires='onegov.reservation:'
+             'Backfill reservation prices from invoice lines'
+)
 def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
     """ The `Backfill reservation prices from invoice lines` upgrade repairs
     the reservation *data*, but invoice lines that were already zeroed by a
--- a/src/onegov/org/upgrade.py
+++ b/src/onegov/org/upgrade.py
@@ -1011,8 +1011,9 @@ def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
     still show 0. Recompute them now that the reservation prices are restored.
 
     Scoped to reservation invoices touched since the backfill rollout that are
-    tied to a paying reservation: an allocation priced `per_item`/`per_hour`,
-    or a `per_item` resource (whose content fallback was the broken one). Only
+    tied to an allocation the bug could have disturbed (`per_item`/`per_hour`/
+    `free` — the original broken OR only ever matched `free`), or to a
+    `per_item` resource (whose content fallback was the broken one). Only
     refreshed when safe (manual, still-open payment).
     """
     from onegov.ticket import Ticket
@@ -1048,7 +1049,7 @@ def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
                     SELECT 1 FROM allocations a
                      WHERE a."group" = r.target
                        AND a.data->>'pricing_method'
-                           IN ('per_item', 'per_hour')
+                           IN ('per_item', 'per_hour', 'free')
                 )
            )
     """)).scalars().all()
--- a/src/onegov/org/upgrade.py
+++ b/src/onegov/org/upgrade.py
@@ -1006,14 +1006,12 @@ def switch_to_parsed_event_filters(context: UpgradeContext) -> None:
              'Backfill reservation prices from invoice lines'
 )
 def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
-    """ The `Backfill reservation prices from invoice lines` upgrade repairs
-    the reservation *data*, but invoice lines that were already zeroed by a
-    refresh (before the data was corrected) still show 0. Recompute those
-    invoices now that both the data and the price fallback are in place.
+    """ `Backfill reservation prices from invoice lines` repairs the
+    reservation data, but invoice lines already zeroed by an earlier refresh
+    still show 0. Recompute them now that data and fallback are in place.
 
-    Scoped to reservation tickets that actually need it (a reservation line at
-    0 while the reservation now has a non-zero price) and only refreshed when
-    it is safe to do so (manual, still-open payment).
+    Scoped to tickets that need it (reservation line at 0 but reservation price
+    now non-zero) and only refreshed when safe (manual, still-open payment).
     """
     from onegov.ticket import Ticket
 
@@ -1024,8 +1022,7 @@ def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
     if not context.has_table('reservations'):
         return
 
-    # only org-based apps have reservation tickets (and an org with a
-    # rounding base); other apps sharing these tables have nothing to do here
+    # only org-based apps have reservation tickets and a rounding base
     org = getattr(context.app, 'org', None)
     if org is None:
         return
--- a/src/onegov/org/upgrade.py
+++ b/src/onegov/org/upgrade.py
@@ -1010,8 +1010,10 @@ def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
     reservation data, but invoice lines already zeroed by an earlier refresh
     still show 0. Recompute them now that the reservation prices are restored.
 
-    Scoped to tickets that need it (reservation line at 0 but reservation price
-    now non-zero) and only refreshed when safe (manual, still-open payment).
+    Scoped to reservation invoices touched since the backfill rollout that are
+    tied to a paying reservation: an allocation priced `per_item`/`per_hour`,
+    or a `per_item` resource (whose content fallback was the broken one). Only
+    refreshed when safe (manual, still-open payment).
     """
     from onegov.ticket import Ticket
 
@@ -1034,13 +1036,20 @@ def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
           JOIN invoice_items ii
             ON ii.invoice_id = t.invoice_id
            AND ii.group = 'reservation'
-           AND ii.unit = 0
+           AND COALESCE(ii.modified, ii.created) >= '2026-08-18'
           JOIN reservations r
             ON r.token = t.handler_id::uuid
+          JOIN resources res
+            ON res.id = r.resource
          WHERE t.handler_code = 'RSV'
            AND (
-                COALESCE((r.data->>'price_per_item')::numeric, 0) <> 0
-             OR COALESCE((r.data->>'price_per_hour')::numeric, 0) <> 0
+                res.content->>'pricing_method' = 'per_item'
+             OR EXISTS (
+                    SELECT 1 FROM allocations a
+                     WHERE a."group" = r.target
+                       AND a.data->>'pricing_method'
+                           IN ('per_item', 'per_hour')
+                )
            )
     """)).scalars().all()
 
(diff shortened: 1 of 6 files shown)


src/onegov/reservation/upgrade.py
```diff
@@ -506,67 +506,86 @@ def add_source_id_to_reserved_slots(context: UpgradeContext) -> None:
     )
 
 
-@upgrade_task('Store pricing settings on reservations')
-def store_pricing_settings_on_reservations(context: UpgradeContext) -> None:
```
Daverball, line 510: Did this migration already run on all of the instances? If not you should probably keep it and fix the two incorrect constants in the CTE instead.

If the intent is that your migration will work, even if this original migration never ran, you will need to explicitly handle `free` and `per_item`/`per_hour` unit prices of `0` and store those settings on the reservation too, since otherwise it will be possible to override those by changing the settings on the linked resource or allocation.

It might also be fine to just fix the existing migration, give it a new name, so it gets run again. Instead of adding a new migration. Since that should also fix the price settings.

Backfilling the settings from the invoice items seems dangerous, since we don't store whether it is a `per_hour` or `per_item` unit on the item, so you could end up with mixed settings, where the pricing method is based on a new setting, but the unit price is based on old setting. You could try to infer from the quantity on the item, whether it was a per-item or per-hour item, but it's not very reliable. So I think we should use the resources and allocations as the only source of truth.

Re-running the previous migration with a fix carries a small risk of changing the stored settings on this two week window of reservations, if the settings on any linked resources/allocations have been changed within this window. But that risk seems negligible, since the same thing would've happened if they did the same thing at any point in time before those two weeks.

Daverball: If we rely on the existing migration, it would also be very easily hotfixable, since we could run the migration again `sudo ogc core upgrade` before the next rollout. So the affected time period would be a little smaller than the full two weeks.

src/onegov/reservation/models/custom_reservation.py
```diff
@@ -90,6 +90,30 @@ def invoice_item(
...
+                    price_per_item = recovered
+            elif pricing_method == 'per_hour' and not price_per_hour:
+                allocation = allocation or self.allocation_obj
+                recovered = (
+                    (allocation.data or {}).get('price_per_hour')
+                    if allocation is not None else None
+                ) or resource.price_per_hour
+                if recovered:
+                    price_per_hour = recovered
+            # end FIXME OGC-3406
```
Daverball: This seems a little paranoid, if we're confident we fixed the problem, we shouldn't need a workaround. If we leave the workaround it'll be harder to catch whether we actually fixed it.

src/onegov/reservation/upgrade.py
Tschuppi81: Yes, deployed and executed last Tuesday.

src/onegov/reservation/models/custom_reservation.py
Tschuppi81: Yeah, its true!

src/onegov/reservation/upgrade.py
Daverball: It still seems less error-prone to fix the migration and re-run it than try to get a secondary migration right. We still have the original source data, so there's no good reason to develop an entirely new secondary migration, that might also make some small mistakes.

It seems more valuable to add a test-case for all possible permutations and check whether the migration ends up with the correct reservation data in all cases. You'll need 3 resources with 4 allocations for each `pricing_method` setting, so 3 resources, 12 allocations and 12 reservations in total. You can probably cut it down to 1 resource 4 allocations and 4 reservations if you use `pytest.mark.parametrize` to change the `pricing_method` on the resource.

Tschuppi81: Ok, let me try to 'just' fix the migration.

src/onegov/reservation/upgrade.py
```diff
@@ -526,49 +549,78 @@ def store_pricing_settings_on_reservations(context: UpgradeContext) -> None:
...
+                                'pricing_method',
+                                res.content->'pricing_method',
+                                'price_per_hour',
+                                COALESCE(
+                                    res.content->'price_per_hour',
+                                    '0.0'::jsonb
+                                ),
+                                'price_per_item',
+                                COALESCE(
+                                    res.content->'price_per_reservation',
```
Tschuppi81: the fallback case price lives on the resource content under `price_per_reservation`

src/onegov/reservation/upgrade.py
```diff
@@ -526,49 +549,78 @@ def store_pricing_settings_on_reservations(context: UpgradeContext) -> None:
                    ) AS pricing
               FROM allocations
              WHERE resource = mirror_of
-               AND data->>'pricing_method' = 'price_per_item'
-                OR data->>'pricing_method' = 'price_per_hour'
-                OR data->>'pricing_method' = 'free'
+               AND (
+                    data->>'pricing_method' = 'per_item'
+                 OR data->>'pricing_method' = 'per_hour'
```
Tschuppi81: those were wrong

Tschuppi81:
I also found a few schemas that have reservations that had no pricing stored.

risch:                11417 - per_item Tageskarte, 0 -> 40
stadt_zug          5099 - halls/rooms, 0 -> real price
huenenberg     4369 - per_item recovery
wil/meggen   2 each - per_item recovery

there might be more... still testing

src/onegov/reservation/upgrade.py
```diff
@@ -526,49 +549,78 @@ def store_pricing_settings_on_reservations(context: UpgradeContext) -> None:
...
-           FROM resources
-          WHERE resources.id = resource
-
+           SET data = computed.new_data
+          FROM computed
+         WHERE reservations.id = computed.id
+           AND computed.new_data IS DISTINCT FROM computed.old_data
+           -- leave free reservations untouched: the price is never applied
+           -- (invoice_item returns None for 'free'), so don't churn them
+           AND computed.new_data->>'pricing_method' IS DISTINCT FROM 'free'
```
Daverball: This is incorrect, we still have to store the fact that the pricing method was `free`, otherwise it can suddenly carry a cost in the future when the resource/allocation is changed via the fallback, which is not what we want.

Tschuppi81:
Note: We decided to log all touched entries in order to keep track of the applied changes

src/onegov/reservation/upgrade.py
```diff
@@ -526,49 +549,78 @@ def store_pricing_settings_on_reservations(context: UpgradeContext) -> None:
...
+    for row in result:
+        count += 1
+        log.info(
+            'Stored pricing on reservation %s: method=%s, per_item=%s, '
+            'per_hour=%s', row.id, row.pricing_method,
+            row.price_per_item, row.price_per_hour
+        )
+    if count:
+        log.info('Stored pricing settings on %d reservation(s)', count)
+
```
Daverball: This is fine for testing/debugging, but for the real migration I would not print out anything, it makes the migration a lot slower and potentially very noisy.

I would also simplify the query again and just always update, regardless of whether the data has changed or not. It's not worth spending the extra cycles to detect which reservations would change.

src/onegov/org/upgrade.py
```diff
@@ -997,3 +998,77 @@ def switch_to_parsed_event_filters(context: UpgradeContext) -> None:
...
+
+    ticket_ids = context.session.execute(text("""
+        SELECT DISTINCT t.id
+          FROM tickets t
+          JOIN invoice_items ii
+            ON ii.invoice_id = t.invoice_id
+           AND ii.group = 'reservation'
+           AND ii.unit = 0
+          JOIN reservations r
+            ON replace(r.token::text, '-', '') = t.handler_id
```
Daverball:
```suggestion
            ON r.token = t.handler_id::uuid
```
This is easier, since you don't need any string manipulation.

src/onegov/org/upgrade.py
```diff
@@ -997,3 +998,77 @@ def switch_to_parsed_event_filters(context: UpgradeContext) -> None:
...
+         WHERE t.handler_code = 'RSV'
+           AND (
+                COALESCE((r.data->>'price_per_item')::numeric, 0) <> 0
+             OR COALESCE((r.data->>'price_per_hour')::numeric, 0) <> 0
+           )
+    """)).scalars().all()
```
Daverball, line 1055: There are technically some edge-cases that can be missed by this. We may want to instead refresh all the invoices that have been updated since the rollout date and are tied to reservations on an allocation that has a `pricing_method` set to something other than `free` or `inherit`.

Or since `resource->'price_per_item'` was incorrect, also all resources with `pricing_method` set to `per_item`.

src/onegov/reservation/upgrade.py
Daverball: It's still a little too complex with the `updated` intermediary table, now that you're not logging/filtering that intermediary table is no longer needed.

I expect the final diff in the SQL to be the three lines that needed to be changed. The rest I would expect to stay the same unless there were other mistakes in the query.

Tschuppi81:
@Daverball I am not sure how it should be: currently a `method=='free` gets an `per_item` price. Is this how it should be or not? This confuses me, I would expect a `per_item` price of 0.0. Technically it does not matter as long a its free...

```
2026-08-24 14:42:34 - INFO: Stored pricing on reservation 238: method=free, per_item=45.0, per_hour=0.0
2026-08-24 14:42:34 - INFO: Stored pricing on reservation 181: method=free, per_item=45.0, per_hour=0.0
```

Daverball:
> @Daverball I am not sure how it should be: currently a `method=='free` gets an `per_item` price. Is this how it should be or not? This confuses me, I would expect a `per_item` price of 0.0. Technically it does not matter as long a its free...
> 
> ```
> 2026-08-24 14:42:34 - INFO: Stored pricing on reservation 238: method=free, per_item=45.0, per_hour=0.0
> 2026-08-24 14:42:34 - INFO: Stored pricing on reservation 181: method=free, per_item=45.0, per_hour=0.0
> ```

This can happen because of how our field dependencies work. If you switch from `per_item` to `free` the value in `price_per_item` does not disappear, so when you switch back to `per_item` you see the original price you entered once upon a time.

It's harmless for that to be propagated to the reservation, since as soon as the method is `free`, we ignore all the other settings. The main reason it propagates, is to keep the code simple, there's not much to be gained to propagate different fields based on the value of `pricing_method`. All it does is make the code more complicated and more likely to contain bugs.

Tschuppi81:
I attached a log to the ticket showing that many reservations are being changed with this upgrade

Daverball:
Since `per_item` is the most common pricing method and it's more common for the pricing method to be inherited from the resource, it's unsurprising that a lot of reservations are affected by the misspelled `price_per_reservation`. (we probably should've migrated when we renamed that property, instead of just pointing the property to a different key in the dictionary)

Luckily it's not very harmful for the pricing information to be incorrect for a little while, since invoices are only refreshed on-demand.

src/onegov/org/upgrade.py
```diff
@@ -1034,13 +1036,20 @@ def refresh_zeroed_reservation_invoices(context: UpgradeContext) -> None:
...
          WHERE t.handler_code = 'RSV'
            AND (
-                COALESCE((r.data->>'price_per_item')::numeric, 0) <> 0
-             OR COALESCE((r.data->>'price_per_hour')::numeric, 0) <> 0
+                res.content->>'pricing_method' = 'per_item'
+             OR EXISTS (
+                    SELECT 1 FROM allocations a
+                     WHERE a."group" = r.target
+                       AND a.data->>'pricing_method'
+                           IN ('per_item', 'per_hour')
```
Daverball: Since the unparanthesized `OR` in the original CTE was also incorrect, we probably still need to include `free` here as well.

Tschuppi81: I agree

Daverball (approved):

PR #2652, opened 2026-08-24
Add search to settings

cyrillkuettel:

## Commit message

Add search to settings

<Optional Description>

TYPE: Feature
LINK: OGC-3385

## Checklist

### DIFF
--- a/src/onegov/org/views/settings.py
+++ b/src/onegov/org/views/settings.py
@@ -148,14 +148,14 @@ def view_settings(
                 )
 
                 if fieldset and source_fieldset not in seen_fieldsets:
-                    fieldset_id = source_fieldset.lower().replace(' ', '-')
+                    fieldset_id = f'fieldset-{as_internal_id(source_fieldset)}'
                     fieldsets.append(
                         {
                             'title': fieldset,
                             'search_text': search_text(
                                 fieldset, source_fieldset
                             ),
-                            'link': (f'{setting_link}#fieldset-{fieldset_id}'),
+                            'link': f'{setting_link}#{fieldset_id}',
                         }
                     )
                     seen_fieldsets.add(source_fieldset)
--- a/src/onegov/org/views/settings.py
+++ b/src/onegov/org/views/settings.py
@@ -7,12 +7,14 @@ from dectate import Query
 from markupsafe import Markup
 from webob.exc import HTTPForbidden
 from onegov.api.models import ApiKey
+from onegov.core.directives import fetch_form_class
 from onegov.core.elements import Link, Confirm, Intercooler
 from onegov.core.security import Secret
 from onegov.core.templates import render_macro
 from onegov.event.models.event import EventFilterValue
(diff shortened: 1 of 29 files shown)


Daverball:
The implemented search functionality is quite primitive currently in a few ways:
1. It only takes into account the title of the setting category
2. It only performs a simple case-insensitive containment search
    - it performs no locale specific word stemming
    - it performs no trigram word similarity search to deal with spelling mistakes
3. It does not rank the results in any way, it just applies a visibility filter

As a result I'm not sure how useful the search currently is, since it essentially gives you roughly the same result that `Ctrl + F` would give you, just with a slightly better user experience.

At a minimum you would probably want to be able to find individual settings and not just setting categories and to be able to see which settings from which categories matched the search term.

Maybe we should take a look at how the mkdocs search works, since our settings are static we could probably do things in a very similar way.

cyrillkuettel:
> At a minimum you would probably want to be able to find individual settings and not just setting categories and to be able to see which settings from which categories matched the search term.

Yes that would have been my next step

Daverball:
As an additional pointer, it might be worth adding an explicit setting view registry at this point. We have been piggybacking off the view registry for settings, which makes querying for all of the settings kind of expensive and unreliable, since we have no guardrails against missing metadata. In a dedicated registry it would be quite simple to record all the meta data required for searching.

The easiest way to implement this would be to add a new `CompositeAction` that emits a `HtmlHandleFormAction` for registering the actual view and a `SettingAction` for registering the metadata into a dedicated setting section. Although it might better to write a dedicated `SettingViewMetaAction` that provides its own registry.

cyrillkuettel:
https://github.com/user-attachments/assets/c04ec29a-4abe-4f2a-b780-6f36df271e07



Search in actual settings now. Tough I'm not happy yet how it looks visually.

Daverball:
Nice. A drop-down, instead of displaying the results inline would probably work a little bit better. You may also want to take into account fieldsets and field descriptions.

src/onegov/org/views/settings.py
```diff
@@ -72,35 +74,111 @@ def view_settings(
...
-                yield setting
+                description, source_description = translated_text(
+                    field.kwargs.get('description')
+                )
+                fieldset, source_fieldset = translated_text(
+                    field.kwargs.get('fieldset')
+                )
+
+                if fieldset and source_fieldset not in seen_fieldsets:
+                    fieldset_id = source_fieldset.lower().replace(' ', '-')
```
Daverball: Since we now depend on that derived spelling of the fieldset label as a id in more than one place, we may want to add an `id` `cached_property` to `Fieldset` and switch to `as_internal_id` to generate it (`f'fieldset-{utils.as_internal_id(self.label)}'`), since that handles special-characters more robustly.

cyrillkuettel:
https://github.com/user-attachments/assets/f7cbcbfd-c1ba-4ec4-826e-c0a1f362f915

With dropdown now.

PR #2659, opened 2026-08-27
Feature/ogc 3387 search depriorize ris documents

Tschuppi81:
Search: Lower ranking of RIS meeting/meeting item/political business documents in search results

Lowering of parties and politicians is not needed

TYPE: Feature
LINK: ogc-3387

### DIFF
--- a/src/onegov/org/models/search.py
+++ b/src/onegov/org/models/search.py
@@ -290,6 +290,8 @@ class Search(Pagination[Any]):
         else:
             ts_query = func.websearch_to_tsquery(self.language, self.query)
 
+        # FIXME: pages/topics should have a much lower decay as other documents
+        # as this content is considered valid for a long period.
         decay = 0.99
         scale = (90 * 24 * 3600)  # 90 days to reach target decay
         offset = (7 * 24 * 3600)  # 7 days without decay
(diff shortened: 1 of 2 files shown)


src/onegov/org/models/search.py
Tschuppi81: I will take care of this in a separate branch/PR

src/onegov/org/models/search.py
```diff
@@ -340,6 +340,9 @@ def generic_search(self) -> Query[SearchIndex]:
                     # NOTE: Tickets may be excluded entirely in the
                     #       future but for now we'll de-prioritize them
                     (SearchIndex.owner_tablename == 'tickets', 0.2),
+                    # de-priorize RIS meeting documents
+                    (SearchIndex.owner_tablename == 'par_meetings', 0.1),
+                    (SearchIndex.owner_tablename == 'par_meeting_items', 0.1),
```
Daverball, line 345: Eventually we probably want to address the `HACK` comment above and take a closer look at each searchable model again to decide how it should rank relative to other models. But for now this is fine I suppose.

Daverball (approved):
Looks good, although the `CASE` statement is starting to get unwieldy.

src/onegov/org/models/search.py
Tschuppi81: Started a branch for it: https://github.com/OneGov/onegov-cloud/pull/2660

PR #2664, opened 2026-08-31
Feature/ogc 2313 reservation confirmation email with magic link

Tschuppi81:
Reservation: Adds ics subscribe url to email confirmation, add limited summary view via durable magic link (no login)

This replaces the single mtan access in the confirmation email with a durable link to a limited summary view with option to (citizen) login.

TYPE: Feature
LINK: ogc-2313


src/onegov/org/views/resource.py
```diff
@@ -1566,11 +1582,16 @@ def view_my_reservations_json(
...
+    conditions = [
+        func.lower(stmt.c.email) == email.lower(),
         start <= stmt.c.start,
-        stmt.c.start <= end
-    )))
+        stmt.c.start <= end,
+    ]
+    if limited:
+        # magic-link visitors: confirmed reservations only
+        conditions.append(stmt.c.accepted.is_(True))
```
Daverball: Pending reservations don't seem like sensitive information to me. It makes sense as a default filter on the iCal subscription, but in this calendar they can more easily see which reservations are not yet accepted. I would either include the same reservations, or limit the time-window they can see. Although I'm not sure what a good time-range would be here. We will probably get some complaints if we make it too short. It seems safe to exclude reservations that are older than a month though.

Daverball (commented):
This seems reasonable overall.

The only thing I'm not quite sure about is the decision to limit the reservations to accepted ones, since almost certainly the first time someone clicks on that link they will see a completely empty calendar, which is not particularly helpful.

It makes sense as the default for the iCal subscription, since it's easier to mistake a pending reservation for a confirmed one there, but for the FullCalendar view I'd include them.

We can maybe think about restricting the time window they can see. We could e.g. safely hide reservations that are more than one month in the past. Since each reservation e-mail gets its own unique link, we could also consider encoding reservation session UUID into the magic link token, and only show the reservations that match the token. That way a leaked URL will never expose reservations made in the past/future.

src/onegov/org/views/resource.py
Tschuppi81: I am not sure if we should limit it at all:
- someone wants to lookup when/where a past recurrence was
- somebody offers yoga and whatever classes on separate series. nice to see all reservations

Daverball: But then they can just login, if they need to see more. Isn't the point of the simple link, that isn't time-limited, that they can see the things that are relevant to the e-mail they received, without having to log in?

The argument is more sound for the iCal link. But they can still just subscribe to every link they receive individually and get a more complete picture that way. Or log in and subscribe to the long-lived version.

We definitely would want to communicate well, that the links in the e-mail only apply to the reservations in the same series, so they know that they can't just reuse an old link for a new ticket, but that seems fine to me.

Tschuppi81: Makes sense.

src/onegov/org/views/reservation.py
```diff
@@ -964,30 +963,52 @@ def finalize_reservation(self: Resource, request: OrgRequest) -> Response:
     return morepath.redirect(url)
 
 
-def get_my_reservations_url(request: OrgRequest, email: str) -> str | None:
+def get_my_reservations_url(
+    request: OrgRequest,
+    email: str,
+    reservation_token: object = None
```
```suggestion
    reservation_token: str | None = None
```
Or `UUID | None` would be fine too, depends on whether you're retrieving it from the `Ticket` where it's stored as a string, or from the `Reservation` where it's stored as a UUID.

It's better to be explicit here, otherwise the type checker won't help us spot bugs.

src/onegov/org/views/reservation.py
```diff
@@ -964,30 +963,52 @@ def finalize_reservation(self: Resource, request: OrgRequest) -> Response:
...
-    )
-    return request.link(
-        auth,
-        name='confirm-citizen-login',
-        query_params={'token': tan_obj.tan}
+    salt = secrets.token_urlsafe(16)
+    payload: object = email if reservation_token is None else {
+        'email': email,
+        'token': str(reservation_token),
+    }
```
```suggestion
    payload = {
        'email': email,
        'token': reservation_token,
    }
```
Let's keep things simple and always use the same payload shape, so we don't need `isinstance` checks. We can just do `if payload.get('token')` instead.

src/onegov/org/views/reservation.py
```diff
@@ -964,30 +963,52 @@ def finalize_reservation(self: Resource, request: OrgRequest) -> Response:
...
+) -> str | None:
+    """ Durable magic link to the recipient's reservations calendar feed. """
+    if not request.app.org.citizen_login_enabled:
+        return None
+
+    salt = secrets.token_urlsafe(16)
+    url_obj = URL(request.class_link(
+        ResourceCollection, name='my-reservations-ical'
+    ))
+    token = request.new_url_safe_token(email, salt)
```
Daverball: We should use the same payload shape here, I probably would also limit this to the reservation token. As discussed they can subscribe to multiple links if they want to be subscribed to each of their reservation series.

src/onegov/org/views/resource.py
```diff
@@ -1536,6 +1537,26 @@ def view_occupancy(
...
+    restricted to that single ticket/token."""
+    if request.authenticated_email:
+        return request.authenticated_email, False, None
+    payload = request.load_url_safe_token(
+        request.GET.get('token'), request.GET.get('salt'), None
+    )
+    if isinstance(payload, str):
+        return payload, True, None
+    if isinstance(payload, dict) and isinstance(payload.get('email'), str):
+        return payload['email'], True, payload.get('token')
```
Daverball: Let's keep things simple and assume we always get a dictionary with the correct shape, we're in control of what kind of data we decide to encode.

If someone manages to smuggle an invalid payload in we have bigger problems, and crashing is actually a good thing.
```suggestion
    assert isinstance(payload, dict)
    return payload['email'], True, payload.get('token')
```

src/onegov/org/views/resource.py
```diff
@@ -1566,11 +1588,16 @@ def view_my_reservations_json(
...
+    conditions = [
+        func.lower(stmt.c.email) == email.lower(),
         start <= stmt.c.start,
-        stmt.c.start <= end
-    )))
+        stmt.c.start <= end,
+    ]
+
+    if reservation_token:
+        conditions.append(stmt.c.token == reservation_token)
```
Daverball: `stmt.c.token` is a UUID column, you may need to `type_coerce` `reservation_token` or manually convert to UUID in order to get the correct behavior here. Or store a UUID in the payload to begin with, so you don't need to convert here.

src/onegov/org/views/resource.py
```diff
@@ -1616,12 +1645,15 @@ def view_my_reservations_pdf(
...
     conditions = [
-        func.lower(stmt.c.email) == request.authenticated_email.lower(),
+        func.lower(stmt.c.email) == email.lower(),
         start <= stmt.c.start,
         stmt.c.start <= end,
     ]
 
-    if request.GET.get('accepted') == '1':
+    if reservation_token:
+        conditions.append(stmt.c.token == reservation_token)
```
Daverball: Same thing here

src/onegov/org/views/reservation.py
```diff
@@ -964,30 +963,60 @@ def finalize_reservation(self: Resource, request: OrgRequest) -> Response:
     return morepath.redirect(url)
 
 
-def get_my_reservations_url(request: OrgRequest, email: str) -> str | None:
+def get_my_reservations_url(
+    request: OrgRequest,
+    email: str,
+    reservation_token: str | None = None
```
Daverball: Looking at this again it's probably better to take a `UUID | None` here, so we serialize the token as a `UUID`, that way we don't have to convert from a string to a UUID every time we retrieve this payload.

src/onegov/org/views/reservation.py
```diff
@@ -964,30 +963,60 @@ def finalize_reservation(self: Resource, request: OrgRequest) -> Response:
...
+            'token': request.new_url_safe_token(payload, salt),
+            'salt': salt,
+        }
     )
 
 
+def get_reservations_subscribe_url(
+    request: OrgRequest,
+    email: str,
+    reservation_token: str | None = None
```
Daverball: Same thing here.

src/onegov/org/views/resource.py
```diff
@@ -1536,6 +1538,24 @@ def view_occupancy(
     }
 
 
+def resolve_reservations_access(
+    request: OrgRequest
+) -> tuple[str | None, bool, str | None]:
```
Daverball, line 1543: Which means we return `tuple[str | None, bool, UUID | None]` here instead.

src/onegov/org/views/resource.py
```diff
@@ -1566,11 +1587,16 @@ def view_my_reservations_json(
...
+    conditions = [
+        func.lower(stmt.c.email) == email.lower(),
         start <= stmt.c.start,
-        stmt.c.start <= end
-    )))
+        stmt.c.start <= end,
+    ]
+
+    if reservation_token:
+        conditions.append(stmt.c.token == UUID(reservation_token))
```
Daverball, line 1597: Which means we no longer need to coerce here, which might fail if someone passed an invalid token, which type checkers wouldn't help us catch.

src/onegov/org/views/resource.py
```diff
@@ -1616,12 +1644,15 @@ def view_my_reservations_pdf(
...
     conditions = [
-        func.lower(stmt.c.email) == request.authenticated_email.lower(),
+        func.lower(stmt.c.email) == email.lower(),
         start <= stmt.c.start,
         stmt.c.start <= end,
     ]
 
-    if request.GET.get('accepted') == '1':
+    if reservation_token:
+        conditions.append(stmt.c.token == UUID(reservation_token))
```
Daverball, line 1653: Same thing here

src/onegov/org/views/resource.py
```diff
@@ -1811,12 +1869,15 @@ def view_my_reservations_ical(
...
-    records = request.session.execute(select(*stmt.c).where(and_(
+    conditions = [
         func.lower(stmt.c.email) == email.lower(),
         s <= stmt.c.start, stmt.c.start <= e,
         # only include accepted reservations in ICS file
         stmt.c.accepted.is_(True)
-    )))
+    ]
+    if reservation_token:
+        conditions.append(stmt.c.token == UUID(reservation_token))
```
Daverball, line 1879: Same thing here

src/onegov/org/views/reservation.py
```diff
@@ -1137,7 +1166,10 @@ def accept_reservation(
                 'form': form,
                 'message': message,
                 'my_reservations_url': get_my_reservations_url(
-                    request, self.email
+                    request, self.email, ticket.handler_id
+                ),
+                'subscribe_url': get_reservations_subscribe_url(
+                    request, self.email, ticket.handler_id
```
Daverball: I would just switch from `self.handler_id` to `self.token` here, so you get the UUID instead of the string.

src/onegov/org/views/reservation.py
```diff
@@ -1899,7 +1931,10 @@ def send_reservation_summary(
                 'code': self.handler.data.get('key_code'),
                 'changes': self.handler.get_changes(request),
                 'my_reservations_url': get_my_reservations_url(
-                    request, recipient
+                    request, recipient, self.handler_id
+                ),
+                'subscribe_url': get_reservations_subscribe_url(
+                    request, recipient, self.handler_id
```
Daverball: And here you could use `self.handler.reservations[0].token`

Daverball (commented):
Looks good, but it seems more robust to store the reservation token as a UUID, so we don't accidentally generate invalid URLs.

src/onegov/org/views/reservation.py
Tschuppi81: Nice!

src/onegov/org/views/reservation.py
```diff
@@ -979,7 +980,7 @@ def get_my_reservations_url(
     salt = secrets.token_urlsafe(16)
     payload = {
         'email': email,
-        'token': reservation_token,
+        'token': reservation_token.hex if reservation_token else None,
```
Daverball: I didn't notice we're not using our own json serializer for `new_url_safe_token` and `load_url_safe_token`.

If you pass `onegov.core.custom.json` as the `serializer` argument to `URLSafeTimedSerializer` in those two methods, then you don't need to convert the UUID to a string.

Tschuppi81: Valid point, but when changing it I found `SoftUUID` from I think `libres/SQLAlchemy` a subclass of `UUID` that fails to serialize - subclasses seem not to be allowed/supported: 
```
>       return orjson.dumps(
            obj,
            default=default_serializers.encode,
            option=options
        )
E       TypeError: Type is not JSON serializable: SoftUUID
```

Options:

1. we need to add and register a UUID serializer and wire it (separate PR)
2. make it a `UUID` by `uuid.UUID(reservation_token.hex)`
3. or keep the `.hex` notation

I think Option 2 with a FIXME comment is the way to go for this PR

Daverball: I already had to do 2. for `_blocking_resource_id_mapping`, yes. I would also go with that, but use the integer value for constructing the new object, that bypasses parsing, so is more efficient: `UUID(int=reservation_token.int)`

Tschuppi81:
@Daverball is it fine with the custom serializer?

src/onegov/org/views/resource.py
Daverball:
```suggestion
) -> tuple[str | None, bool, UUID | None]:
```

Since we're storing a UUID now, we should be getting one here as well.

src/onegov/org/views/resource.py
Daverball:
```suggestion
        conditions.append(stmt.c.token == reservation_token)
```
Which means we don't need to coerce here

src/onegov/org/views/resource.py
Daverball:
```suggestion
        conditions.append(stmt.c.token == reservation_token)
```
or here

src/onegov/org/views/resource.py
Daverball:
```suggestion
        conditions.append(stmt.c.token == reservation_token)
```
or here

src/onegov/org/views/resource.py
```diff
@@ -1778,14 +1830,20 @@ def view_my_reservations_ical(
...
 ) -> Response:
 
-    token = request.GET.get('token')
     salt = request.GET.get('salt')
-    email = request.load_url_safe_token(token, salt, None)
+    token = request.GET.get('token')
+    payload = request.load_url_safe_token(token, salt, None)
     include_key_code = False
-    if email is None:
+    reservation_token = None
```
Daverball:
```suggestion
    reservation_token: UUID | None = None
```
Let's tell the type checker what we expect to be working with.

Daverball (commented):
Apart from a now incorrect type hint and some unecessary type coercions this is now fine, yes.

PR #2671, opened 2026-09-01
Add cli for finding feriennet instances in specific phases

BreathingFlesh:

## Commit message

Feriennet: Add CLI for searching ferienpass instances

Lists ferienpass instances that are in the specified state

TYPE: Feature

### DIFF
--- a/src/onegov/feriennet/cli.py
+++ b/src/onegov/feriennet/cli.py
@@ -436,3 +436,60 @@ def match_fairness(
                        f'no place {gunplaced}/{gn} ({pct(gunplaced, gn)})')
 
     return match_fairness
+
+
+@cli.command(name='which-ferienpass-is-currently',
+             context_settings={'default_selector': '*'})
+@click.argument('phase', required=True)
+def which_ferienpass(
+    phase: str,
+) -> Callable[[FeriennetRequest, FeriennetApp], None]:
+    """ Displays all ferienpass instances where the option applies.
+
+    Example::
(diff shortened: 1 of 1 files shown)


src/onegov/feriennet/cli.py
```diff
@@ -436,3 +436,60 @@ def pct(part: int, whole: int) -> str:
...
+
+    """
+    def search_ferienpass(
+        request: FeriennetRequest,
+        app: FeriennetApp
+    ) -> None:
+
+        if not hasattr(app, 'active_period'):
+            return
+
```
Tschuppi81, line 462: Guard the input string against invalid `phases`. May list the valid once to help the user..

src/onegov/feriennet/cli.py
```diff
@@ -436,3 +436,60 @@ def pct(part: int, whole: int) -> str:
...
+        if phase == 'inactive':
+            if (
+                not app.active_period or (
+                    app.active_period.phase == 'inactive'
+                )
+            ):
+                click.echo(f'{app.schema} - {app.org.title}')
+        if not app.active_period:
```
Tschuppi81, line 470: isn't this a duplicate of line 460?

BreathingFlesh: No, `hasattr(app, 'active_period')` returns `False` for all non-feriennet instances and `True` for Feriennet. `app.active_period` returns either the active period or `None` if there is no active period, but the attribute still exists.

PR #2678, opened 2026-09-04
Feature/ogc 2313 list unavailable dates for series

Tschuppi81:
Reservation: Show chronologically all available and unavailable slots for booking series

TYPE: Feature
LINK: ogc-2313


src/onegov/org/views/resource.py
```diff
@@ -816,16 +816,105 @@ def spot_infos_for_free_slots(
...
+                        continue
+                    # flag dates with no bookable slot in any room
+                    if not any(
+                        isclose(slot.availability, 100.0, abs_tol=.005)
+                        for slots in room_slots[date].values()
+                        for slot in slots
+                    ):
+                        series_overview.append({
+                            'date': date, 'time': None, 'room': None
+                        })
```
Daverball: This seems far too complicated and expensive and duplicates a lot of the logic for creating the reservations. `reserved_dates` tells you which dates and rooms were reserved, don't emit another query to find the reservations you just created, we don't want previous form submissions to leak state into the current submission either. At this point we should just know exactly what happened, this should be a simple comprehension like `missing_dates`. At most there should be a single branch to distinguish between reserved/missing dates/rooms.

Daverball (commented):
I'm not convinced it's worth doing anything here. The problem is not what this list looks like. The problem is that only the person creating the series gets to ever see it and that it's out of date as soon as they make manual changes to the pre-selected dates.

Tschuppi81:
> I'm not convinced it's worth doing anything here. The problem is not what this list looks like. The problem is that only the person creating the series gets to ever see it and that it's out of date as soon as they make manual changes to the pre-selected dates.

That is fully clear. In my opinion it just helps you to see lets say 1 or 2 occasions in your series can not happen as planned and may help to find an available slot in another room...

src/onegov/org/views/resource.py
Tschuppi81: I see, let me refine that

Daverball:
> That is fully clear. In my opinion it just helps you to see lets say 1 or 2 occasions in your series can not happen as planned and may help to find an available slot in another room...

I mean you can already see that, since it lists all of the missing dates/rooms. For large series this output seems too verbose and repeats information you have both in the table and in the list of reservations. Getting the same list of dates three times seems overwhelming to me. A more condensed output that tells you which dates to focus your attention on seems more helpful to me in this workflow.

Your view is more helpful at the end, after you've submitted/finalized the reservations, when you need to see a summary of all of the dates at a glance including the ones in the series that were skipped. It's a good idea, it's just not the right place for it imho and gets more tricky to get right once you're dealing with multiple rooms, since the information will be spread across multiple tickets.

Tschuppi81:
Weekly: its the wrong location in the workflow. Later on with the confirmation.

PR #2679, opened 2026-09-04
Bugfix/pro 1580 sanitize user tags

Tschuppi81:
User: Sanitize stringified repr tag values and fix HH:MM delivery-times

  1. PRO-1580 — Some user tags were stored as the Python repr of a list                  
     ("['Sport']", "[]") instead of the list contents.                        
                                                                                         
  2. OGC-3205 (follow-up) — The sanitize added above strips every                        
     non-alphanumeric server-side, and tags-input.js already stripped them               
     client-side, so ":" was removed from HH:MM delivery times ("07:05" ->               
     "07 05")     
                                                                                         
  TYPE: Bugfix                                                                           
  LINK: pro-1580                                                                         
  LINK: ogc-3205 

Starting on this ticket I found weird repr like wrapped user tags:
```
reto@ister:/opt/containers/onegov-cloud/tree/d79c928411d2/app/lib/python3.14/site-packages/onegov$ sudo ogc user --select /onegov_feriennet/* analyze-tags
  ['[]']
  ['[]']
  ['[]']
  ['[]']
4 user(s) with corrupted tags in onegov_feriennet/aarau
  ['[]']
  ['[]']
  ['[]']
  ['[]']
  ['[]']
  ['[]']
  ['[]']
7 user(s) with corrupted tags in onegov_feriennet/aarwangen
  ['["[\'[]\']"]']
1 user(s) with corrupted tags in onegov_feriennet/affoltern
0 user(s) with corrupted tags in onegov_feriennet/affoltern_albis
0 user(s) with corrupted tags in onegov_feriennet/albula
  ['[\'["[\\\'[]\\\']"]\']']
1 user(s) with corrupted tags in onegov_feriennet/am_alten_rhein
  ['[]']
  ["['[]']"]
  ["['[]']"]
  ['[]']
  ['[]']
5 user(s) with corrupted tags in onegov_feriennet/arbon
0 user(s) with corrupted tags in onegov_feriennet/attiswil
  ['[]']
1 user(s) with corrupted tags in onegov_feriennet/auw
0 user(s) with corrupted tags in onegov_feriennet/avenchesvully
  ['[]']
  ['[]']
2 user(s) with corrupted tags in onegov_feriennet/bachenbuelach
0 user(s) with corrupted tags in onegov_feriennet/baden
  ['[]']
1 user(s) with corrupted tags in onegov_feriennet/balsthal
  ['[]']
  ["['[]']"]
  ['["[\'[]\']"]']
  ['["[\'[]\']"]']
  ['[]']
  ['[]']
6 user(s) with corrupted tags in onegov_feriennet/bellmund_hermrigen_merzlingen
  ['[]']
  ['[]']
  ['[]']
3 user(s) with corrupted tags in onegov_feriennet/berg
  ['[]']
  ['[]']
2 user(s) with corrupted tags in onegov_feriennet/bergdietikon
0 user(s) with corrupted tags in onegov_feriennet/best_of_swissweb
0 user(s) with corrupted tags in onegov_feriennet/bex
  ['[]']

..
```

### DIFF
--- a/src/onegov/form/fields.py
+++ b/src/onegov/form/fields.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import inspect
 import phonenumbers
+import re
 import sedate
 
 from cssutils.css import CSSStyleSheet  # type:ignore[import-untyped]
@@ -746,16 +747,14 @@ class TagsField(StringField):
     """
 
     widget = TagsWidget()
-    # FIXME: Why does data have a different shape depending on if it's
-    #        passed in by the form or the object?! This seems like a bug
-    data: str | list[str]  # type:ignore[assignment]
+    # mirrors the client-side sanitation of tags-input.js: keep alphanumerics
+    # and latin-1 accented characters, everything else becomes whitespace
+    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]')
+    data: list[str]  # type:ignore[assignment]
 
     def _value(self) -> str:
-        # Without this override, we had the underlying data corrupted
-        # containing strings like ["['[]']"]
-        if isinstance(self.data, list):
-            return ','.join(self.data)
-        return self.data or ''
+        # data is always a list; join for the comma-separated widget value
+        return ','.join(self.data) if self.data else ''
 
     def process_formdata(self, valuelist: list[RawFormValue]) -> None:
         if not valuelist:
@@ -764,13 +763,16 @@ class TagsField(StringField):
 
         values_str = valuelist[0]
         if isinstance(values_str, str) and values_str != '':
-            values = (v.strip() for v in values_str.split(','))
+            values = (
+                ' '.join(self._sanitize.sub(' ', v).split())
+                for v in values_str.split(',')
+            )
             self.data = [v for v in values if v]
         else:
             self.data = []
 
     def process_data(self, value: list[str] | None) -> None:
-        self.data = ','.join(value) if value else ''
+        self.data = list(value) if value else []
 
 
 class IconField(StringField):
--- a/src/onegov/form/fields.py
+++ b/src/onegov/form/fields.py
@@ -776,12 +776,7 @@ class TagsField(StringField):
 
 
 class DeliveryTimesField(TagsField):
-    """ TagsField for HH:MM times; keeps digits and colons.
-
-    Keeps the tags-input chip UI but opts into ':' both client-side (via
-    DeliveryTimesWidget's data-allow-chars) and server-side (_sanitize),
-    so HH:MM times are not mangled to "HH MM".
-    """
+    """ TagsField for HH:MM times; keeps ':' client- and server-side. """
 
     widget = DeliveryTimesWidget()
     _sanitize = re.compile(r'[^0-9:]')
--- a/src/onegov/form/fields.py
+++ b/src/onegov/form/fields.py
@@ -750,7 +750,7 @@ class TagsField(StringField):
     widget = TagsWidget()
     # mirrors the client-side sanitation of tags-input.js: keep alphanumerics
     # and latin-1 accented characters, everything else becomes whitespace
-    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]')
+    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]+')
     data: list[str]  # type:ignore[assignment]
 
     def _value(self) -> str:
--- a/src/onegov/form/fields.py
+++ b/src/onegov/form/fields.py
@@ -34,7 +34,6 @@ from onegov.form.widgets import OrderedMultiCheckboxWidget
 from onegov.form.widgets import PanelWidget
 from onegov.form.widgets import PreviewWidget
 from onegov.form.widgets import TableFieldWidget
-from onegov.form.widgets import DeliveryTimesWidget
 from onegov.form.widgets import TagsWidget
 from onegov.form.widgets import TextAreaWithTextModules
 from onegov.form.widgets import TreeSelectWidget
@@ -750,7 +749,7 @@ class TagsField(StringField):
     widget = TagsWidget()
     # mirrors the client-side sanitation of tags-input.js: keep alphanumerics
     # and latin-1 accented characters, everything else becomes whitespace
-    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]')
+    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]+')
     data: list[str]  # type:ignore[assignment]
 
     def _value(self) -> str:
@@ -764,7 +763,7 @@ class TagsField(StringField):
         values_str = valuelist[0]
         if isinstance(values_str, str) and values_str != '':
             values = (
-                ' '.join(self._sanitize.sub(' ', v).split())
+                self._sanitize.sub(' ', v).strip()
                 for v in values_str.split(',')
             )
             self.data = [v for v in values if v]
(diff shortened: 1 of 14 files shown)


src/onegov/form/fields.py
```diff
@@ -746,16 +748,13 @@ class TagsField(StringField):
     """
 
     widget = TagsWidget()
-    # FIXME: Why does data have a different shape depending on if it's
-    #        passed in by the form or the object?! This seems like a bug
-    data: str | list[str]  # type:ignore[assignment]
+    # mirrors the client-side sanitation of tags-input.js: keep alphanumerics
+    # and latin-1 accented characters, everything else becomes whitespace
+    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]')
```
Daverball:
```suggestion
    _sanitize = re.compile(r'[^A-Za-z0-9À-ÿ]+')
```
So we don't need to normalize repeated spaces after calling `re.sub`.

src/onegov/form/fields.py
```diff
@@ -764,13 +763,23 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
 
         values_str = valuelist[0]
         if isinstance(values_str, str) and values_str != '':
-            values = (v.strip() for v in values_str.split(','))
+            values = (
+                ' '.join(self._sanitize.sub(' ', v).split())
```
Daverball:
```suggestion
                self._sanitize.sub(' ', v)
```
This can be a simple `sub` call, if we fix our pattern to match adjacent forbidden characters/spaces.

src/onegov/form/fields.py
```diff
@@ -764,13 +763,23 @@ def process_formdata(self, valuelist: list[RawFormValue]) -> None:
...
     def process_data(self, value: list[str] | None) -> None:
-        self.data = ','.join(value) if value else ''
+        self.data = list(value) if value else []
+
+
+class DeliveryTimesField(TagsField):
+    """ TagsField for HH:MM times; keeps ':' client- and server-side. """
+
+    widget = DeliveryTimesWidget()
+    _sanitize = re.compile(r'[^0-9:]')
```
Daverball: This doesn't really work, invalid times don't get rejected and invalid characters get translated to spaces, so the end-result will still be invalid. You may be better off refactoring this into a `parse_item` method that for `TagsField` does the `_sanitize.sub` and for this field calls `time.strptime` so a `ValueError` happens for an invalid time. Sanitization doesn't really make sense in this case. Invalid data should be a hard error.

But it's also somewhat questionable to derive this field from `TagsField`, since the client-side validation/sanitization will be wrong as well. It might be better to build a custom field based on `FieldList(TimeField(...), ...)` and write the necessary JavaScript to make that work. If the client-side input for adding a tag is of type `time` it will not be possible to add invalid values to begin with.

src/onegov/org/assets/js/tags-input.js
```diff
@@ -24,6 +24,12 @@
     var SEPERATOR = ',';
     var COPY_PROPS = 'placeholder pattern spellcheck autocomplete autocapitalize autofocus accessKey accept lang minLength maxLength required'.split(' ');
     function tagsInput(input) {
+        // data-allow-chars: chars to keep beyond alnum (e.g. ':'); escaped
+        // before use in the char class to avoid regex injection
+        var allowChars = (input.getAttribute('data-allow-chars') || '')
+            .replace(/[.*+?^${}()|[\]\\-]/g, '\\$&');
```
Daverball: Isn't the only thing you need to escape `]` here, since you're inserting it into a `[]` group?

But this is kind of a bad idea anyways, I don't think we should allow this kind of flexibility, just to make something else work as a `TagsInput` that shouldn't really be one to begin with.

src/onegov/org/layout.py
```diff
@@ -2414,6 +2414,7 @@ class ResourceRecipientsFormLayout(DefaultLayout):
     def __init__(self, model: Any, request: OrgRequest, title: str) -> None:
         super().__init__(model, request)
         self.title = title
+        self.request.include('tags-input')  # delivery-times chip UI
```
Daverball: Why doesn't the `TagsWidget.__call__` do this step? This seems very fragile, since it's easy to forget.

Some of our other widgets should already work this way and pull in their JS dependencies on-demand like this.

src/onegov/user/upgrade.py
```diff
@@ -355,3 +358,50 @@ def add_on_update_cascade_to_username_fk(context: UpgradeContext) -> None:
...
+        if isinstance(tag, str) and tag.startswith('[') and tag.endswith(']'):
+            try:
+                parsed = ast.literal_eval(tag)
+            except (ValueError, SyntaxError):
+                parsed = None
+            if isinstance(parsed, list):
+                result.extend(_unwrap_tags(parsed))
+                continue
+        if isinstance(tag, str) and tag.strip():
+            result.append(tag.strip())
```
Daverball:
```suggestion
        if not isinstance(tag, str):
            continue
        if tag.startswith('[') and tag.endswith(']'):
            try:
                parsed = ast.literal_eval(tag)
            except (ValueError, SyntaxError):
                parsed = None
            if isinstance(parsed, list):
                result.extend(_unwrap_tags(parsed))
                continue
        if stripped := tag.strip():
            result.append(stripped)
```

src/onegov/user/upgrade.py
```diff
@@ -355,3 +358,50 @@ def add_on_update_cascade_to_username_fk(context: UpgradeContext) -> None:
...
+    ):
+        return
+
+    if not context.has_table('users'):
+        return
+
+    for user in context.session.query(User):
+        tags = user.tags
+        if not tags:
+            continue
```
Daverball, line 403:
```suggestion
    for user in (
        context.session.query(User)
        .filter(func.jsonb_array_length(User.tags) > 1)
    ):
```

Let's not load users we don't need to modify.

src/onegov/user/upgrade.py
Daverball: I'm a little unsure about why we strip the tags in addition to to undoing the list repr. It doesn't really have anything to do with the problem we're trying to clean up. If it does in fact not occur, then get rid of this part.

The fact that this is recursive is a little awkward as well, do we have any cases of nested string literals? If not it would seem better to just extend the result without the call to `_unwrap_tags`.

src/onegov/user/upgrade.py
```diff
@@ -355,3 +358,50 @@ def add_on_update_cascade_to_username_fk(context: UpgradeContext) -> None:
...
+
+        fixed = _unwrap_tags(tags)
+        if fixed != tags:
+            user.tags = fixed
+            flag_modified(user, 'data')
```
Daverball: `flag_modified` really shouldn't be necessary here, that's only needed if we modify the existing list in place instead of assigning the whole list.

Daverball (commented):
I'm not super happy about relying on `TagsField` for `DeliveryTimesField`, the current implementation is still buggy and introduces additional points of failure. It seems better to create a proper dedicated field for this.

src/onegov/user/upgrade.py
Tschuppi81: yes, there are weird cases like `"['[\"[\\'[]\\']\"]']"` to be unwrapped in recursion. I added to the test

src/onegov/org/layout.py
Tschuppi81: Removed

src/onegov/org/assets/js/tags-input.js
Tschuppi81: is gone

src/onegov/form/fields.py
Tschuppi81: Switched to `StringField` base: `class DeliveryTimesField(StringField)`

PR #2680, opened 2026-09-04
Feature/pro 1472 inline user search

Tschuppi81:
User: Adds inline user search

TYPE: Feature
LINK: pro-1472


Daverball (commented):
Copying the `SearchWidget` approach from directories and occurrences seems like complete overkill here. I think something like what we provide for `PoliticalBusinessCollection` seems more appropriate here, where we just add a `query`/`term` parameter to the collection. A search function is useful for all schemas, we don't need to make this pluggable, since only managers can see this management view anyways.

Tschuppi81:
Agree, much simpler. I will do in analogy to `PoliticalBusinessCollection`

src/onegov/user/collections/user.py
```diff
@@ -100,6 +110,12 @@ def query(self) -> Query[User]:
         """
         query = self.session.query(User)
 
+        if self.term:
+            query = query.filter(or_(
+                User.username.ilike(f'%{self.term}%'),
+                User.realname.ilike(f'%{self.term}%')
+            ))
```
Daverball: While partial matches seem useful it could also be annoying, we might be better off relying on the full text search here, like with the political businesses. Even if you're not relying on the full text search you should probably use `unaccent` to normalize the name and escape the search term, so search terms containing `_` and `%` don't yield unexpected results. There's an existing helper function and other examples of `ilike` that use it.

PR #2681, opened 2026-09-07
Org: Switch to edit mode for resources recipients form layout

Tschuppi81:
Org: Switch to edit mode for resources recipients form layout

TYPE: Feature
LINK: None


Daverball:
Looking at this it might be easiest to just render this with a simple `TextInput` widget and perform validation and value coercion to and from `list[str]` in a custom `Field` subclass instead of relying on `TagsField` and JavaScript.

If we do make this something more dynamic using JavaScript, then presumably the new hint text here would be incorrect.

Tschuppi81:
Let me split the two things and fix the `TagField` in the other PR

PR #2683, opened 2026-09-07
Api: Undefer occurrence content column

Tschuppi81:
Api: Eager load occurrence/event data preventing N+1 queries

TYPE: Performance
LINK: ONEGOV-CLOUD-5YG
LINK: ONEGOV-CLOUD-5YE

### DIFF
--- a/src/onegov/event/collections/occurrences.py
+++ b/src/onegov/event/collections/occurrences.py
@@ -663,7 +663,6 @@ class OccurrenceCollection(Pagination[Occurrence]):
 
         """
 
-        # Event join: filters/ordering and the batch eager-load
         query = self.apply_common_filters(
             self.session.query(Occurrence).join(Event)
         )
--- a/src/onegov/event/collections/occurrences.py
+++ b/src/onegov/event/collections/occurrences.py
@@ -657,9 +657,14 @@ class OccurrenceCollection(Pagination[Occurrence]):
 
         """
 
+        event = contains_eager(Occurrence.event)
         query = self.apply_common_filters(
-            self.session.query(Occurrence).join(Event)
-            .options(contains_eager(Occurrence.event).joinedload(Event.image))
+            self.session.query(Occurrence)
+            .join(Event)
+            .options(
+                event.joinedload(Event.image),
+                event.undefer(Event.content)
+            )
             .options(undefer(Occurrence.content))
         )
 
--- a/src/onegov/event/collections/occurrences.py
+++ b/src/onegov/event/collections/occurrences.py
@@ -663,9 +663,9 @@ class OccurrenceCollection(Pagination[Occurrence]):
             .join(Event)
             .options(
                 event.joinedload(Event.image),
-                event.undefer(Event.content)
+                event.undefer(Event.content),
+                undefer(Occurrence.content),
             )
-            .options(undefer(Occurrence.content))
         )
 
         if self.term:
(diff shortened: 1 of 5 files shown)


src/onegov/event/collections/occurrences.py
```diff
@@ -657,9 +657,15 @@ def query(self) -> Query[Occurrence]:
...
         query = self.apply_common_filters(
-            self.session.query(Occurrence).join(Event)
-            .options(contains_eager(Occurrence.event).joinedload(Event.image))
+            self.session.query(Occurrence)
+            .join(Event)
+            .options(
+                event.joinedload(Event.image),
+                event.undefer(Event.content)
+            )
+            .options(undefer(Occurrence.content))
```
Daverball:
```suggestion
                event.undefer(Event.content),
                undefer(Occurrence.content),
            )
```

src/onegov/event/collections/occurrences.py
```diff
@@ -87,7 +88,7 @@ class _Sentinel(Enum):
 ))
 
 
-class OccurrenceCollection(Pagination[Occurrence]):
+class OccurrenceCollection(ApiBatchEagerLoad[Occurrence]):
```
Tschuppi81: this approach to be applied to the PR https://github.com/OneGov/onegov-cloud/pull/2684 as well

Daverball: Maybe it would be better if there was a way to inject options into `Pagination.batch` in general and not via subclassing but via individual objects. Maybe something like `Pagination.set_query_options`. That would be more flexible and ensure we only eager load things when we actually need them.

Technically there is already a hacky way via `cached_subset`, by re-assigning the modified query to that cached attribute, but that will also affect the `subset_count` query.

This would be closer to how we use other collections where we can do `collection.query().options(...)`, with `Pagination` the equivalent thing would then be `collection.set_query_options(...); collection.batch`.

src/onegov/event/collections/occurrences.py
```diff
@@ -657,9 +657,9 @@ def query(self) -> Query[Occurrence]:
 
         """
 
+        # Event join needed for filters/ordering and the batch eager-load
         query = self.apply_common_filters(
             self.session.query(Occurrence).join(Event)
-            .options(contains_eager(Occurrence.event).joinedload(Event.image))
```
Daverball, line 662: Let's make these the default options on this collection, so we don't regress.

I.e. set `query_options = (contains_eager(Occurrence.event).joinedload(Event.image),)`

src/onegov/event/collections/occurrences.py
```diff
@@ -736,6 +736,18 @@ def query(self) -> Query[Occurrence]:
...
+        Eager-loads the event data serialized per occurrence, batch only.
+        """
+        # relies on the Event join added in query()
+        event = contains_eager(Occurrence.event)
+        return self.set_query_options(
+            event.joinedload(Event.image),
+            event.undefer(Event.content),
+            undefer(Occurrence.content),
+        )
+
```
Daverball: I'm not convinced it's worth pulling this out into its own method, if it's only used by the API, just let the API set those options directly. If we end up reusing this somewhere else we can refactor it then, but even then it seems better to keep it separate, since it might diverge again in the future.

Keeping it separate will also enable us to be a lot more efficient in some places and only load the half a dozen or so columns we actually access instead of all of them, which can be a big win for list views.

Daverball (commented):
Looks good, although we should avoid regressing elsewhere by changing the default options. We can still make those options opt-in by putting them into `query_options` as the default, instead of applying them unconditionally.

Tschuppi81:
Like it! Now api collection can and does apply different query options than the regular collection

Daverball (approved):

PR #2687, opened 2026-09-08
 Org: Add PDF attachments to reservation acceptance messages.

cyrillkuettel:
## Commit message

Org: Add PDF attachments to reservation acceptance messages.

TYPE: Feature
LINK: OGC-3371

## Checklist

### DIFF
--- a/src/onegov/org/path.py
+++ b/src/onegov/org/path.py
@@ -386,7 +386,6 @@ def get_file_for_org(
 
     public_ticket_attachment = obj.type == 'messagefile' and any(
         isinstance(message, TicketChatMessage)
-        and message.meta.get('origin') == 'external'
         for message in obj.links
     )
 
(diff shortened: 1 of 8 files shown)


src/onegov/org/path.py
```diff
@@ -386,7 +386,6 @@ def get_file_for_org(
 
     public_ticket_attachment = obj.type == 'messagefile' and any(
         isinstance(message, TicketChatMessage)
-        and message.meta.get('origin') == 'external'
         for message in obj.links
```
Daverball:
```suggestion
        for message in obj.linked_messages  # type: ignore[attr-defined]
```

Going through `links` which queries every file association table is extremely expensive, we should only query the relationship that actually matters.

cyrillkuettel: thank you, ill check the other commit as well which has attachement upload.

PR #2690, opened 2026-09-09
Ogc 643 release notes admin

BreathingFlesh:

## Commit message

Town6: Show new release notes

TYPE: Feature
LINK: OGC-643

## Checklist


src/onegov/town6/layout.py
```diff
@@ -166,6 +169,87 @@ def drilldown_back(self) -> str:
...
+                year, number = value.split('.', 1)
+                return int(year), int(number)
+
+            current_release_path = repo_root / 'changes' / 'current'
+            if not current_release_path.exists():
+                return features
+
+            for yaml_file in sorted(current_release_path.glob('*.yaml')):
+                with yaml_file.open('r', encoding='utf-8') as fh:
+                    add_payload(yaml.safe_load(fh) or {})
```
Daverball, line 212: This current vs past folder thing was only intended for our release workflow, in production there should be no distinction between current and past releases. The extra folder is for unreleased changes that will be moved by the released script.

I would structure changes as `changes/{release_tag}/{internal_change_name_that_doubles_as_sort_key}.yaml`, pending changes live in `changes` without a subfolder and get moved by the release script into a new folder for the release that's being created.

This method should then walk the subfolders of `repo_root / 'changes'` in reverse sorted order, ignoring anything that isn't a folder and then for each subfolder walk all of the yaml files. You can factor this into a single nested loop, special-casing the first iteration, in that the limit of 5 features doesn't apply to the newest release.

We probably also want to cache the available releases and the parsed yaml files per release in redis, so we don't have to walk the files on every request for every logged in user that tries to access this property.

changes/current/example_german.yaml
```diff
@@ -0,0 +1,13 @@
+# title: Seitenschlagwörter prominenter dargestellt
+# description:
+#   Schlagwörter auf einer Seite werden nun deutlicher sichtbar angezeigt und erleichtern damit die Einordnung des Inhalts.
+# applications:
+#   - all
+# #   - onegov_feriennet
```
Daverball, line 6: These are namespaces and the same application can have multiple namespaces, e.g. `onegov_town6_without_yubikey`, so I don't think this is a good way to filter this.

What makes most sense to me is to structure this in a similar way it is structured in our docs and in the user interface and define a set of modules like:

- Reservations
- Forms
- RIS
- ...

Each application then defines which of those modules it supports and we use that to filter which changes are relevant.

We may want an additional setting on the Organisation to ignore changes for certain modules that aren't used by this Organisation and/or exclude some modules dynamically, like RIS changes, when the RIS is not enabled.

src/onegov/user/auth/core.py
```diff
@@ -369,6 +369,7 @@ def complete_login(
 
         user.last_login = utcnow()
         user.save_current_session(request)
+        del request.browser_session['dismissed_new_features']
```
Tschuppi81, line 372: Couldn't this break the login in case the key was never set? 
```
request.browser_session.pop('dismissed_new_features', None)
```

src/onegov/town6/layout.py
```diff
@@ -166,6 +169,87 @@ def drilldown_back(self) -> str:
...
+                            break
+                        with yaml_file.open('r', encoding='utf-8') as fh:
+                            add_payload(yaml.safe_load(fh) or {})
+
+            # Translations if there are locales
+            for feature in features:
+                for value in feature:
+                    if type(feature[value]) is dict:
+                        feature[value] = feature[value].get(
+                            str(self.request.locale).split('_')[0], None
```
Tschuppi81, line 248: Default locale should be `de` instead of None

Daverball: `self.request.locale` takes the `default_locale` specified in the settings into account, I would probably just put a
```python
if self.request.locale is None:
    return None
```
at the top, if the instance wasn't configured for localization, we don't need to bother with emitting a changelog. In our case that should never happen anyways.

I also would lift the operation of turning the locale into a language code out of the loop and put this at the top of the function as well:

```python
language = self.request.locale.split('_', 1)[0]
```

And then you can just reuse that local variable everywhere.

do/release
```diff
@@ -138,6 +138,30 @@ function new-version() {
...
+    shopt -s nullglob
+    local files=("$current_release_dir"/*)
+    shopt -u nullglob
+
+    for file in "${files[@]}"; do
+        if [[ ! -f "$file" ]]; then
+            continue
+        fi
+
+        if [[ "$file" == "$release_dir" ]]; then
```
Daverball: Isn't this check redundant? You're already excluding folders with the above check.

You are missing a check against the example file though, unless the idea is that it goes away once we have our first real change and the other changes serve as examples instead.

src/onegov/town6/layout.py
```diff
@@ -167,6 +170,74 @@ def drilldown_back(self) -> str:
...
+    def new_features(self) -> list[dict[str, Any]]:
+        features: list[dict[str, Any]] = []
+
+        if user := self.request.current_user:
+            user_release = user.release_features
+            if user_release == self.app.version:
+                return features
+
+            repo_root: Path | None = None
+            for parent in Path(__file__).resolve().parents:
```
Daverball: This is not very robust and only works if OneGov is installed from a cloned git repository in development mode, which isn't the case in our production containers.

You will need to include the release folder tree using the `MANIFEST` file and/or through setuptools package data configuration in `setup.cfg`, so those files also end up in `site-packages` or some other well-known place referenced by the package metadata. Then you should be able to access these files using [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) which uses a slightly more cumbersome API, since resources may not exist on the filesystem, since they could e.g. be compressed into an archive instead and need to be decompressed before they can be accessed, although `Traversable` implements a subset of the `pathlib.Path` API, so the code can probably mostly stay the same.

src/onegov/user/auth/core.py
```diff
@@ -355,6 +355,7 @@ def complete_login(
         to = to or self.to
 
         response = self.redirect(request, to)
+        request.browser_session.pop('dismissed_new_features', None)
```
Daverball: Maybe we should put this into the `on_login` callback for Town6 instead, since this feature no longer is used for all applications? This line is relatively benign, so I don't mind it being here too much, but it might encourage future extensions in this spot.

src/onegov/user/upgrade.py
```diff
@@ -53,6 +53,14 @@ def add_realname_column(context: UpgradeContext) -> None:
     context.session.flush()
 
 
+@upgrade_task('Add release features column')
+def add_release_features_column(context: UpgradeContext) -> None:
+    context.operations.add_column(
+        'users', Column('release_features', Text, nullable=True))
+
+    context.session.flush()
```
Daverball: Generally you shouldn't need to flush changes in upgrade tasks, unless you're doing something else in the same upgrade task right afterwards. In this case I don't think the flush would do anything either way though, since alembic operations happen immediately afaik.

tests/onegov/town6/test_views_usermanagement.py
```diff
@@ -23,3 +31,46 @@ def test_add_new_user_hides_editmode_links(client: Client) -> None:
...
+        'description: A useful new feature\n',
+        encoding='utf-8'
+    )
+
+    town6_layout = importlib.import_module('onegov.town6.layout')
+    monkeypatch.setattr(
+        town6_layout,
+        '__file__',
+        str(tmp_path / 'src' / 'onegov' / 'town6' / 'layout.py')
+    )
```
Daverball:
```suggestion
    monkeypatch.setattr(
        'onegov.town6.layout',
        '__file__',
        str(tmp_path / 'src' / 'onegov' / 'town6' / 'layout.py')
    )
```
I believe this should work without having to explicitly import the module, `monkeypatch` accepts the module name in addition to the module. It might even work with a single argument IIRC:

```suggestion
    monkeypatch.setattr(
        'onegov.town6.layout.__file__',
        str(tmp_path / 'src' / 'onegov' / 'town6' / 'layout.py')
    )
```

changes/example.yaml
```diff
@@ -0,0 +1,6 @@
+# title: Neue Features bei Anmeldung
+# description:
+#   Neu werden ewähnenswerte Features beschrieben und bei Login angezeigt. Die
```
Tschuppi81: Minor: typo

do/release
BreathingFlesh: You're right, I'll remove it.

Yeah, the idea is that the example file moves just like the other files and we can just look at the last file for guidance.

src/onegov/town6/layout.py
BreathingFlesh: Since we only have this in town6 now, can I just use the same logic as with the feriennet sponsors?
```
    @cached_property
    def sponsors(self) -> list[Sponsor]:
        return load_sponsors(utils.module_path('onegov.feriennet', 'sponsors'))
```
In that case I would just move the 'changes' folder into town6.

Daverball: That will work, yes. The `MANIFEST` currently includes the entire `src` directory, apart from `__pycache__` and `.mo`/`.pyo`.

It's not as robust as using `importlib.resources`, but it should be fine for now. Maybe you can also add a relative symlink from the current location in the repository root to the new location inside `town6`, so it's a bit easier to find.

src/onegov/town6/layout.py
```diff
@@ -197,8 +188,8 @@ def add_payload(
                     payload.keys()):
                     features.append(payload)
 
-            changes_path = repo_root / 'changes'
-            if not changes_path.exists():
+            changes_path = resource_files('onegov.town6').joinpath('changes')
```
Daverball, line 191: I think `Traversable` also supports `/` as shorthand for `joinpath`: https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable.__truediv__

src/onegov/town6/views/usermanagement.py
```diff
@@ -110,3 +110,33 @@ def town_handle_new_user(
     layout.edit_mode = True
     return handle_new_user(
         self, request, form, layout)
+
+
+@TownApp.view(
+        model=User,
+        permission=Secret,
```
Daverball: This should probably be `Personal`, or at most `Private`, otherwise only admins will be able to see the changelog. Same with the view below.

tests/onegov/town6/test_views_usermanagement.py
```diff
@@ -23,3 +30,44 @@ def test_add_new_user_hides_editmode_links(client: Client) -> None:
...
     assert 'save-link' not in added
     assert 'cancel-link' not in added
+
+
+def test_new_feature_close_and_read(
+    client: Client,
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: Path,
+) -> None:
+    client.login_admin()
```
Daverball: Let's make sure to test with the lowest privilege level that we want to be able to see changelogs. So `login_member` if we want to support all logged in users, or `login_editor` if we want to support editors and up, that way you will catch incorrect permissions on the two changelog views.

src/onegov/town6/layout.py
```diff
@@ -167,6 +170,66 @@ def drilldown_back(self) -> str:
...
+                return features
+
+            def add_payload(
+                    payload: dict[str, Any],
+            ) -> None:
+                required_keys = {'title', 'description'}
+
+                if isinstance(payload, dict) and required_keys.issubset(
+                    payload.keys()):
+                    features.append(payload)
```
Daverball: I think this helper is a little too simple to be useful now, it would also be good if we filtered before caching, so our cache is as small as possible, so I would fold this check into the comprehension for the `new_features_releases` creator.

src/onegov/town6/layout.py
```diff
@@ -167,6 +170,66 @@ def drilldown_back(self) -> str:
...
+                            yaml_file.read_text(encoding='utf-8')) or {}
+                        for yaml_file in sorted(
+                            (path for path in
+                             changes_path.joinpath(release_name).iterdir()
+                             if path.is_file() and path.name.endswith(
+                                 '.yaml')),
+                            key=lambda path: path.name
+                        )
+                    ]
+                )
```
Daverball: With the new simplified architecture, where there's no classification of features and everyone always sees all of the features, I think it makes sense to simplify this and replace it with a single shared cache that contains the maximum number of features anyone could ever see regardless of what their last viewed release is.

I would probably split this into a helper generator function with a nested loop iterating through the folder structure in sorted order that yields a tuple of release name and the parsed YAML file, making sure to already skip invalid entries in the generator function and stopping once you've yielded either yielded all features from the latest release or the maximum number of features. Then you can do `{release: [features for _, features in group] for release, group in groupby(releases_generator(), key=itemgetter(0)))}` to create the final dictionary of lists, that gets cached in redis.

Then all users can share the same cache and you just need to apply one final filter on the release name excluding the releases they've already seen the changelog for, the other filters are already pre-applied.

Daverball (commented):
With the new architecture where everyone sees all of the features we can simplify the caching.

It also looks like currently both TranslatorDirectory and Fsi would see the changelog, since they inherit from `TownApp` but unlike `Feriennet` and `Pas` they don't override `layout.pt`. I wonder if it's more robust if we add a setting like `enable_yubikey` that's on by default in `Town6` but off in all of the subclasses that determines whether the changelog will be shown. That way it will also be possible to explicitly turn off the changelog via `onegov.yml` on some instances if we want to.

Daverball (commented):
Looks good, there's one more small optimization I would do to avoid repeated accesses to the filesystem.

Other than that the only question I have is whether it's worth parsing and rendering the description as markdown using `render_untrusted_markdown`.

If we do want to support that, there are some potential pitfalls, since YAML string literals like to eat significant whitespace by default, so we would need to take care to use `|` block scalars so newlines and indentation are preserved so things like paragraphs and lists are rendered correctly.

src/onegov/town6/views/usermanagement.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from onegov.core.security import Secret
+from onegov.core.security import Secret, Private
```
Daverball:
```suggestion
from onegov.core.security import Secret, Personal
```

src/onegov/town6/views/usermanagement.py
```diff
@@ -110,3 +110,33 @@ def town_handle_new_user(
     layout.edit_mode = True
     return handle_new_user(
         self, request, form, layout)
+
+
+@TownApp.view(
+        model=User,
+        permission=Private,
```
Daverball:
```suggestion
        permission=Personal,
```

src/onegov/town6/views/usermanagement.py
```diff
@@ -110,3 +110,33 @@ def town_handle_new_user(
...
+) -> None:
+
+    request.assert_valid_csrf_token()
+
+    self.release_features = request.app.version
+
+
+@TownApp.view(
+        model=User,
+        permission=Private,
```
Daverball:
```suggestion
        permission=Personal,
```

.github/workflows/admin_digital_docs_release_notes.yaml
```diff
@@ -0,0 +1,74 @@
...
+          if [ "$COUNT" -gt 0 ]; then
+            echo "has_files=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_files=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: List yaml files
+        if: steps.check.outputs.has_files == 'true'
+        run: |
+          find ./latest-changes -type f -name "*.yaml"
```
Daverball: Instead of manually pulling the docker container, I think I would just run the job inside the container:

```
    container:
      image: ghcr.io/onegov/onegov-cloud:${{ inputs.image_tag }}
      credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.GITHUB_TOKEN }}
```

As for updating the changelog: It is more reliable if the job always creates the full changelog for as many releases we want to keep in the changelog, similar to how `do/changes` does it. That way if we add any features or change some formatting, everything will be updated consistently, instead of new changes looking different from old changes. That way we can also correct things like spelling mistakes in old changes after releasing them.

I would write a script like `do/changes` to generate the release notes markdown file. The main question is where it should live. It would make more sense if it lived in the docs repository, since you will be checking that repository out and adding a commit to it. If it lived in this repository we would need to also check out our own repository or put the script inside the container or include it with the workflow. The docs repository also knows best where the generated file should end up, the desired formatting etc. That way you can also easily test and iterate on the script offline, all it would need is a single argument pointing it toward the `changes` root directory.

So this job would:
1. Check if changes directory exists inside container (probably optional, since it should always exist)
2. Checkout the docs repository with our seantis-ci bot account credentials
3. Run the script from the repository to re-generate the release notes
4. Run the action to add and commit the changed file with our seantis-ci bot account credentials

You might need to guard against no changes between steps 3 and 4, but I believe `EndBug/add-and-commit` already cleanly handles cases where the added file is unchanged.

src/onegov/town6/views/usermanagement.py
```diff
@@ -1,4 +1,6 @@
-from onegov.core.security import Secret
+from __future__ import annotations
+
+from onegov.core.security import Secret, Private
```
Daverball:
```suggestion
from onegov.core.security import Secret, Private
```

tests/onegov/town6/test_views_usermanagement.py
```diff
@@ -1,5 +1,15 @@
+from __future__ import annotations
+
+
+
+from onegov.user import User
```
Daverball:
```suggestion
from onegov.user import User
```

Daverball (commented):
Looks like a couple of `from __future__ import annotations` snuck back in during the merge.

PR #2695, opened 2026-09-11
Org: Allow wildcard CSP domains for iFrame URLs

Tschuppi81:
Org: Allow wildcard CSP domains for iFrame URLs

TYPE: Bugfix
LINK: pro-1583

### DIFF
--- a/src/onegov/org/forms/page.py
+++ b/src/onegov/org/forms/page.py
@@ -144,14 +144,18 @@ class IframeForm(PageBaseForm):
             return
 
         domain = '/'.join(field.data.split('/', 3)[:3])
-        allowed_domains = {
-            d.rstrip('/')
-            for d in self.allowed_domains
-        }
-        if domain not in allowed_domains:
-            raise ValidationError(
-                _('The domain of the URL is not allowed for iFrames.')
-            )
+        for allowed in self.allowed_domains:
+            allowed = allowed.rstrip('/')
+            # CSP child_src entries may use a wildcard host (*.vimeo.com)
+            if '*' in allowed:
+                prefix, _sep, suffix = allowed.partition('*')
+                if domain.startswith(prefix) and domain.endswith(suffix):
+                    return
+            elif domain == allowed:
+                return
+        raise ValidationError(
+            _('The domain of the URL is not allowed for iFrames.')
+        )
 
 
 class PageUrlForm(ChangeAdjacencyListUrlForm):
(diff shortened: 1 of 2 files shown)


Daverball (approved):
This seems fine, although we may want to implement something more robust using `urlparse` or `purl.URL` which uses `urlparse`, so we can match scheme against scheme and `netloc` against `netloc`.

You may be able to get away with `fnmatch` for matching against the wildcards in `hostname`. Even if it's built for filename matching it should be no less accurate than what we have now.

Daverball (approved):

src/onegov/org/forms/page.py
```diff
@@ -143,15 +146,20 @@ def validate_url(self, field: URLField) -> None:
...
-            )
+        url = urlparse(field.data)
+        for allowed in self.allowed_domains:
+            allowed_url = urlparse(allowed.rstrip('/'))
+            # CSP child_src entries may use a wildcard host (*.vimeo.com)
+            if (
+                url.scheme == allowed_url.scheme
+                and url.hostname is not None
+                and allowed_url.hostname is not None
+                and fnmatch(url.hostname, allowed_url.hostname)
```
Daverball: This is mostly a nitpick and we currently don't use this setting this way, but CSP sources can technically contain a non-empty path. If the path ends with a `/` it will be a prefix match, otherwise it is an exact match. The only exception is the empty path so `https://example.com` and `https://example.com/` behave the same, since the former doesn't have a path so it matches any path and the latter is a prefix match starting from the root, so it still will match anything. But other path values would work differently.

src/onegov/org/forms/page.py
```diff
@@ -143,15 +146,20 @@ def validate_url(self, field: URLField) -> None:
...
-            d.rstrip('/')
-            for d in self.allowed_domains
-        }
-        if domain not in allowed_domains:
-            raise ValidationError(
-                _('The domain of the URL is not allowed for iFrames.')
-            )
+        url = urlparse(field.data)
+        for allowed in self.allowed_domains:
+            allowed_url = urlparse(allowed.rstrip('/'))
```
Daverball: Either way this `rstrip` is no longer necessary here.

src/onegov/org/forms/page.py
Daverball: Also there's technically some other special casing for schemes and ports that we would need to add if we wanted to fully mirror the behavior of the equivalent CSP source.

The source scheme `http` will allow `https` in addition to `http`, but not the other way around, an empty scheme in the source will allow any scheme.

An empy port will allow the default port for the specified scheme, an explicit port will allow exactly that port, only the port `*` will match any port.

We need to worry about these less, but we may at least want to document that we don't support all CSP source matching features here.

PR #2699, opened 2026-09-15
Feature/ogc 3423 ticket with due date

Tschuppi81:
Ticket: Tickets can have a due date now

TYPE: Feature
LINK: ogc-3423

TODO:
- yaml for release feature


src/onegov/org/custom.py
BreathingFlesh, line 1: Maybe you could also show the due ticket counter in the dropdown if there are no open tickets. Right now the counter is red and shows the open tickets if there are any and it's blue if there are none. But the due tickets should be a higher priority than the pending tickets.

src/onegov/town6/templates/macros.pt
```diff
@@ -1259,6 +1259,36 @@
...
+                <button type="submit" name="preset" value="in_one_week" title="${presets['in_one_week']}" i18n:translate>In one week</button>
+                <label class="due-date-custom">
+                    <span i18n:translate>Custom</span>
+                    <input type="date" name="date" data-auto-submit="true" value="${ticket.due_date and ticket.due_date.isoformat() or ''}">
+                </label>
+                <button tal:condition="current" type="submit" name="clear" value="1" class="due-date-clear" i18n:translate>Clear</button>
+            </form>
+        </details>
+    </tal:b>
+</metal:due_date>
```
BreathingFlesh, line 1290: I would do this post with intercooler and then just replace the ticket-panel. This way it shouldn't take this long to load and the page won't skip to the beginning.

Also it dosn't update correctly if you enter the date via keyboard and not the calendar.

src/onegov/org/custom.py
Tschuppi81: Done

src/onegov/town6/templates/macros.pt
Tschuppi81: Much faster!

PR #2704, opened 2026-09-17
Core: Add cache report tool for debugging

Tschuppi81:
Core: Add cache report tool for debugging

In analogy to `sql_query_report`

TYPE: Feature
LINK: None


Daverball (commented):
It might be worth hooking redis at a lower level, similar to how sentry does it: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/redis/_sync_common.py

Daverball:
Especially considering that we don't know for sure that going forward every single one of those calls will result in a redis query and that's the part we actually would want to track, if we cared about reducing redis overhead specifically. It also has the advantage that there's only `redis.client.Pipeline.execute` and `redis.StrictRedis.execute_command` that you will need to hook instead of half a dozen methods on our CacheRegion, where we may forget to hook additional methods in the future if the dogpile API grows in some way.

Daverball (approved):

PR #2705, opened 2026-09-17
Api: Prevents cache roundtrips for custom event tags

Tschuppi81:
Api: Prevents cache roundtrips for custom event tags, request cached now

TYPE: Performance
LINK: ONEGOV-CLOUD-5Z3


src/onegov/org/api.py
```diff
@@ -262,6 +262,11 @@ class EventApiEndpoint(ApiEndpoint['Occurrence', UUID]):
     endpoint = 'events'
     pk_type = UUID
 
+    @cached_property
+    def custom_event_tags(self) -> list[str] | None:
```
Tschuppi81: does it make sense to cache the redis cache?
The current sentry issue roundtripped 190 times, so I guess it makes sense

Daverball: If it does make sense, I would put the cache at the source and changing it from `property` to `request_cached`. Putting it here will only resurface the problem again later.

But it might be worth to change our cache API so we use a two or even three layered cache by default. Similar to how `orm_cached` operates.

Daverball (approved):

